PR #1, opened 2019-11-28
[MM-20303] Setup basic unit testing

streamer45:
#### Summary

PR sets up basic unit testing.

- Running `make test` should now work.
- Added basic tests for `userentity` package.
- Added basic tests for `memstore` package.
- Moved config to separate package.

#### Ticket

https://mattermost.atlassian.net/browse/MM-20303


cmd/loadtest/main.go
```diff
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strings"
 
+	"github.com/mattermost/mattermost-load-test/config"
```
reflog: should these still reference the old repo?

streamer45: Good catch, I was fooled cause I had the new branch on the old repository directory and everything worked. I'll update all references.

PR #2, opened 2019-11-29
[MM-20302] Implement UserController.SetRate()

streamer45:
#### Summary

PR implements UserController.SetRate() method.

#### Ticket

https://mattermost.atlassian.net/browse/MM-20302

### DIFF
--- a/loadtest/control/simplecontroller/controller.go
+++ b/loadtest/control/simplecontroller/controller.go
@@ -5,6 +5,7 @@ package simplecontroller
 
 import (
 	"errors"
+	"math"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/user"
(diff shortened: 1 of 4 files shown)


loadtest/control/simplecontroller/controller.go
```diff
@@ -48,17 +51,28 @@ func (c *SimpleController) Run(status chan<- user.UserStatus) {
...
+			case <-time.After(time.Millisecond * idleTime):
 			}
 		}
 
 		// status <- user.UserStatus{User: c.user, Info: "user loop done", Code: user.STATUS_DONE}
 	}
 }
 
+func (c *SimpleController) SetRate(rate float64) error {
+	if rate < 0 {
```
ethervoid: Are we going to allow 0 value? I'm asking this because if its 0 the [time.After clause](https://github.com/mattermost/mattermost-load-test-ng/pull/2/files#diff-49fc95e0da7ba3b3ed68753b4b3ee624R60) is going to be 0 as well

streamer45: Yes, we want to allow 0 as value for the case where we want to run actions as fast as possible (no wait times between API calls). We could add a check to avoid calling `time.After` but since efficiency is not our main concern at this time I think it's simpler to keep it this way.

PR #4, opened 2019-12-01
added CreateGroupChannel

reflog:


loadtest/user/userentity/actions.go
```diff
@@ -68,4 +68,17 @@ func (ue *UserEntity) CreatePost(post *model.Post) (string, error) {
...
 
 	return post.Id, err
+}
+
+func (ue *UserEntity) CreateGroupChannel(memberIds []string) (string, error) {
+	user, err := ue.store.User()
+	if user == nil || err != nil {
+		return "", errors.New("user was not initialized")
+	}
+	channel, resp := ue.client.CreateGroupChannel(append(memberIds, user.Id))
```
streamer45: I believe here it's enough to pass `memberIds` since the logged in user will always be part of the group channel.

PR #5, opened 2019-12-01
added ViewChannel

reflog:


loadtest/control/simplecontroller/actions.go
```diff
@@ -88,3 +88,20 @@ func (c *SimpleController) createGroupChannel() user.UserStatus {
 	
 	return user.UserStatus{User: c.user, Info: fmt.Sprintf("group channel created, id %v", channelId)}
 }
+
+func (c *SimpleController) viewChannel() user.UserStatus {
+	channel, err := c.user.Store().Channel("") // TODO: fetch channel randomly?
```
streamer45: Correct, here we are going to use [this](https://mattermost.atlassian.net/browse/MM-20704).

loadtest/store/store.go
```diff
@@ -9,14 +9,14 @@ import (
 
 type UserStore interface {
 	Id() string
+	Post(postId string) (*model.Post, error)
+	User() (*model.User, error)
+	Channel(channelId string) (*model.Channel, error)
```
streamer45: This is important. There was a reason why those were in the `MutableUserStore`. We are returning a pointer to the actual data stored and in the case of the memory store that's also a way to alter that data. Whatever method ends up in the `UserStore` interface needs to return a copy otherwise we break immutability. I would keep this unchanged for now to avoid dealing with [deep copying](https://en.wikipedia.org/wiki/Object_copying#Deep_copy) until we really need it.

reflog: @streamer45 
Understood. But:
How should `channel, err := c.user.Store().Channel("")` work now that there isn't a Channel() method in immutable store?

streamer45: The ticket was about implementing the `User` method. We lack an important piece still (comment above) before we can make the controller function work properly. I would't worry about this function too much. We'll get back to it in due time.

PR #6, opened 2019-12-01
added GetChannelUnread()

reflog:

PR #7, opened 2019-12-02
added GetChannelMembers

reflog:


reflog:
@streamer45 - Yea, I thought about adding it, but: I don't think we should store that info or maintain it. It's updated via websockets usually, so I don't think we should rely on our local 'cache' of membership. Thoughts?

streamer45:
> @streamer45 - Yea, I thought about adding it, but: I don't think we should store that info or maintain it. It's updated via websockets usually, so I don't think we should rely on our local 'cache' of membership. Thoughts?

The idea is that we are going to listen to WebSocket at some point in `user.Connect()` and update the store accordingly. Having said that, I feel it's important to update the store here because I would like to move away from returning that data entirely. `User` interface was not meant to be a simple wrapper to the API but it was created to completely separate behaviour and state handling logic.

reflog:
@streamer45 understood. Any reason why you didn't ask for the same change in https://github.com/mattermost/mattermost-load-test-ng/pull/6 ? Just missed it or you don't think we'll need that info?
If it's needed, I'll submit one PR for ViewChannel/GetChannelUnread/GetChannelMembers and GetChannelStats data storing

streamer45:
> @streamer45 understood. Any reason why you didn't ask for the same change in #6 ? Just missed it or you don't think we'll need that info?
> If it's needed, I'll submit one PR for ViewChannel/GetChannelUnread/GetChannelMembers and GetChannelStats data storing

I think it was a combination of the two really (probably overlooked a bit and data returned wasn't on my priority list). Now I am really focusing on getting the users to do something close to what we had in the [previous version](https://github.com/mattermost/mattermost-load-test) of the loadtest agent and getting channel members is definitely data we are going to need to do that. No need to update the old PR now, we'll get back to that when we need it.

reflog:
@streamer45 - to keep in the same vein as the other PR, should GetChannelMembers just return error and update the store?

streamer45:
> @streamer45 - to keep in the same vein as the other PR, should GetChannelMembers just return error and update the store?

Yeah, totally fine for now to just return the error.

PR #8, opened 2019-12-02
added GetChannelStats

reflog:


loadtest/user/user.go
```diff
@@ -36,6 +36,7 @@ type User interface {
 	CreateGroupChannel(memberIds []string) (string, error)
 	ViewChannel(view *model.ChannelView) (*model.ChannelViewResponse, error)
 	GetChannelUnread(channelId string) (*model.ChannelUnread, error)
+	GetChannelStats(channelId string) (*model.ChannelStats, error)
```
streamer45: Let's try and make *return something* the exception and not the rule for the methods wrapping the API calls. I say that unless there's some code depending on the return value I would make the function simply return the error.

reflog: @streamer45  emm... if we are not touching the store and not returning the stats here... then what is the method for? 🤔

streamer45: Stress the API. I believe it's a method called frequently when a user selects a channel.
We don't really care about the result cause we are not going to display or act on that info but we want to make sure we call it.

PR #11, opened 2019-12-03
Added User.CreateChannel

ethervoid:
Related issue https://mattermost.atlassian.net/browse/MM-20710


loadtest/user/userentity/actions.go
```diff
@@ -69,6 +69,20 @@ func (ue *UserEntity) CreatePost(post *model.Post) (string, error) {
 	return post.Id, err
 }
 
+func (ue *UserEntity) CreateChannel(channel *model.Channel) (*model.Channel, error) {
```
streamer45: Let's return the channel id only.

PR #12, opened 2019-12-03
Added User.CreateDirectChannel

ethervoid:
Related issue https://mattermost.atlassian.net/browse/MM-20711

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -82,17 +82,17 @@ func (ue *UserEntity) CreateGroupChannel(memberIds []string) (string, error) {
 	return channel.Id, err
 }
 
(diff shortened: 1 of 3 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -82,6 +82,19 @@ func (ue *UserEntity) CreateGroupChannel(memberIds []string) (string, error) {
 	return channel.Id, err
 }
 
+func (ue *UserEntity) CreateDirectChannel(userId1, userId2 string) (*model.Channel, error) {
```
streamer45: I think we can simplify. One of the two users is going to be the current user so we just need the id of the other and leverage the store to get the current user id.

Also let's not return a pointer to the channel but just the channel id (we don't want to break immutability).

PR #16, opened 2019-12-03
[MM-20301] Adds GetTeams method, team store methods and their tests

mgdelacroix:


loadtest/store/memstore/store.go
```diff
@@ -64,6 +64,33 @@ func (s *MemStore) SetChannel(channel *model.Channel) error {
...
+	}
+	return teams, nil
+}
+
+func (s *MemStore) SetTeams(teams []*model.Team) error {
+	for _, team := range teams {
+		s.teams[team.Id] = team
+	}
+	return nil
+}
```
streamer45: This can cause the state to get *polluted*. With the above function we always either add or update and never delete. I think the safest choice right now is to make all store setters always replace the data. In this case we could either replace the map entirely or empty it before setting the teams.

loadtest/user/userentity/actions.go
```diff
@@ -163,3 +163,25 @@ func (ue *UserEntity) GetChannelStats(channelId string) error {
...
+	}
+
+	if err := ue.store.SetTeams(teams); err != nil {
+		return nil, err
+	}
+
+	teamIds := make([]string, len(teams))
+	for _, team := range teams {
+		teamIds = append(teamIds, team.Id)
+	}
```
streamer45: No need to use `append` here. We allocated the slice, let's just assign by index.

loadtest/store/memstore/store.go
mgdelacroix: I'm ok on with the approach, but then we would have to either create update methods or allow updating on some to avoid code repetition. E.g.: after a `SearchUsers` action we cannot use `SetUsers` with the results as it would remove the existing ones that were saved previously on (for example) the `GetUsersForChannel` action.

We can solve this with having both `SetUsers` and `AddUsers`/`UpdateUsers`, for example, to allow for both paths, iterate through results and use `SetUser` (suboptimal) or have a flag that indicates if we want the method to replace preexisting users or not. Any preference here?

streamer45: Sure. I am fine with additional `Add`/`Remove` methods if it comes to that, but let's try to not write code we don't yet need. For example, the case of searching the users could be one of those cases where we only act on the API result and don't store anything.

ethervoid:
Before working on the `GetTeam` method, I think we should move this PR forward. What do you think @streamer45 @mgdelacroix ?

loadtest/user/userentity/actions.go
```diff
@@ -163,3 +163,25 @@ func (ue *UserEntity) GetChannelStats(channelId string) error {
 	}
 	return err
 }
+
+func (ue *UserEntity) GetTeams() ([]string, error) {
+	user, err := ue.store.User()
+	if user == nil || err != nil {
+		return nil, errors.New("user was not initialized")
+	}
```
streamer45: Here we should use `ue.getUserFromStore()` like in all other cases.

mgdelacroix: Oops! On it :)

PR #17, opened 2019-12-05
Added User.GetUsersStatusesByIds

ethervoid:
Added the method User.GetUsersStatusesByIds for the user entity

https://mattermost.atlassian.net/browse/MM-20844

### DIFF
--- a/example/sampleuser/user.go
+++ b/example/sampleuser/user.go
@@ -90,6 +90,10 @@ func (u *SampleUser) GetChannelStats(channelId string) error {
 	return nil
 }
 
+func (u *SampleUser) GetUsersStatusesByIds(userIds []string) ([]*model.Status, error) {
+	return nil, nil
+}
+
 func (u *SampleUser) SignUp(email, username, password string) error {
 	user := model.User{
 		Email:    email,
--- a/example/sampleuser/user.go
+++ b/example/sampleuser/user.go
@@ -90,8 +90,8 @@ func (u *SampleUser) GetChannelStats(channelId string) error {
 	return nil
 }
 
-func (u *SampleUser) GetUsersStatusesByIds(userIds []string) ([]*model.Status, error) {
-	return nil, nil
+func (u *SampleUser) GetUsersStatusesByIds(userIds []string) error {
+	return nil
 }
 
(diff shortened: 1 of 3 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -127,3 +127,8 @@ func (ue *UserEntity) GetChannelUnread(channelId string) (*model.ChannelUnread,
 	return channelUnreadResponse, resp.Error
 
 }
+
+func (ue *UserEntity) GetUsersStatusesByIds(userIds []string) ([]*model.Status, error) {
```
ethervoid: Is there a point to check here for user initialization like [here](https://github.com/mattermost/mattermost-load-test-ng/pull/17/files#diff-b68f89ec953f6d29ef590768f6d33f9eR124) for example? I understand that this should be done if we're going to use the user object but here I don't see the point

reflog:
@ethervoid - please resolve the conflicts
@streamer45 - does this method needs to update the store?

streamer45:
> @ethervoid - please resolve the conflicts
> @streamer45 - does this method needs to update the store?

We do store statuses on webapp and at some point we might as well here. But for now I don't see an immediate reason to store that data especially since we are yet not handling websocket updates.

loadtest/user/user.go
```diff
@@ -36,4 +36,7 @@ type User interface {
 	GetChannelStats(channelId string) error
 
 	// teams
+
+	// users
+	GetUsersStatusesByIds(userIds []string) ([]*model.Status, error)
```
streamer45: Let's just return an error for now.

ethervoid: Done

example/sampleuser/user.go
```diff
@@ -90,6 +90,10 @@ func (u *SampleUser) GetChannelStats(channelId string) error {
 	return nil
 }
 
+func (u *SampleUser) GetUsersStatusesByIds(userIds []string) ([]*model.Status, error) {
```
streamer45: We are missing this change and we are good. If you run `make test` you should see an error caused by this I believe.

ethervoid: Sorry haste makes waster :\

PR #21, opened 2019-12-05
[MM-20858] Implement userentity.AddTeamMember()

streamer45:
#### Summary

PR implements `userentity.AddTeamMember()` and related store methods.

#### Ticket

https://mattermost.atlassian.net/browse/MM-20858

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -14,6 +14,7 @@ type MemStore struct {
 	teams          map[string]*model.Team
 	channels       map[string]*model.Channel
 	channelMembers map[string]map[string]*model.ChannelMember
+	teamMembers    map[string]map[string]*model.TeamMember
 }
 
 func New() *MemStore {
@@ -22,6 +23,7 @@ func New() *MemStore {
 		teams:          map[string]*model.Team{},
 		channels:       map[string]*model.Channel{},
 		channelMembers: map[string]map[string]*model.ChannelMember{},
+		teamMembers:    map[string]map[string]*model.TeamMember{},
 	}
 }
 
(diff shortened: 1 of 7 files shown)


loadtest/store/memstore/store.go
```diff
@@ -108,3 +110,15 @@ func (s *MemStore) RemoveChannelMember(channelId string, userId string) error {
...
+func (s *MemStore) SetTeamMember(teamId string, teamMember *model.TeamMember) error {
+	if s.teamMembers[teamId] == nil {
+		s.teamMembers[teamId] = map[string]*model.TeamMember{}
+	}
+	s.teamMembers[teamId][teamMember.UserId] = teamMember
+	return nil
+}
+
+func (s *MemStore) TeamMember(teamId, userId string) (*model.TeamMember, error) {
+	return s.teamMembers[teamId][userId], nil
```
reflog: Let's add a check for existence here. Unless you like what TeamMember("TRY", "THIS") returns :)

streamer45: You know I would hate the result of that :grin: 
But if you approve this PR I promise my next one will be [this](https://mattermost.atlassian.net/browse/MM-20958).

streamer45: Actually it's not even that bad. 
https://play.golang.org/p/So5jYixXNNK

agnivade: This was interesting. I did not expect that. But after some ogling at it, it makes sense. If the key doesn't exist, a map returns the zero value of its value, which here is a nil map again. And again passing a key which does not exist, returns the zero value of `*model.TeamMember` which is a nil pointer. :)

streamer45: Yes, we were all quite baffled by this. I guess what made it confusing, at least for me, was that the zero value for map is `nil` so you would expect that trying to access that would cause a panic. Thankfully it wasn't hard to find an explanation in Golang's blog:  

> A nil map behaves like an empty map when reading, but attempts to write to a nil map will cause a runtime panic

Good to know for next time :muscle:

PR #23, opened 2019-12-05
Added user.AddChannelMember

ethervoid:
**Summary**

This PR implements the userentity.AddChannelMember method

**Ticket**

https://mattermost.atlassian.net/browse/MM-20846

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -164,6 +164,20 @@ func (ue *UserEntity) RemoveUserFromChannel(channelId, userId string) (bool, err
 	return ok, ue.store.RemoveChannelMember(channelId, userId)
 }
 
+func (ue *UserEntity) AddChannelMember(channelId, userId string) (*model.ChannelMember, error) {
+	member, resp := ue.client.AddChannelMember(channelId, userId)
+	if resp.Error != nil {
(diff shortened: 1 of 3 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -157,6 +157,20 @@ func (ue *UserEntity) CreateDirectChannel(otherUserId string) (string, error) {
 	return channel.Id, nil
 }
 
+func (ue *UserEntity) AddChannelMember(channelId, userId string) (*model.ChannelMember, error) {
```
ethervoid: Between return a custom struct or a map an return the `model.ChannelMember` structs, I decided for the latter

streamer45: The problem with this is that we don't want to return a pointer to the data we are storing cause that makes it potentially mutable by the user of this function. One solution would be to dereference and return a copy of `model.ChannelMember` but I'd avoid that for now cause it brings other issues and it wouldn't be consistent with other functions. I suggest we just return the error for now and revisit this later when we need it.

ethervoid: Ok! :)

ethervoid: Changed to return only error by now

PR #30, opened 2019-12-08
added GetTeamMembers

reflog:


loadtest/user/userentity/actions.go
```diff
@@ -291,6 +291,19 @@ func (ue *UserEntity) AddTeamMember(teamId, userId string) error {
...
+func (ue *UserEntity) GetTeamMembers(teamId string, page, perPage int) (error) {
+	members, resp := ue.client.GetTeamMembers(teamId, page, perPage, "")
+	if resp.Error != nil {
+		return resp.Error
+	}
+	for _, tm := range members {
+		if err := ue.store.SetTeamMember(teamId, tm); err != nil {
+			return err
+		}
+	}
```
streamer45: I think here we should implement and use a `MemStore.SetTeamMembers` (essentially identical to the existing `MemStore.SetChannelMembers` ) which will replace all elements with the ones provided. Otherwise we are just adding or updating.
For reference, same *issue* has been discussed earlier in https://github.com/mattermost/mattermost-load-test-ng/pull/16.

loadtest/store/memstore/store.go
```diff
@@ -142,6 +142,19 @@ func (s *MemStore) SetTeamMember(teamId string, teamMember *model.TeamMember) er
 	return nil
 }
 
+
+func (s *MemStore) SetTeamMembers(teamId string, teamMembers []*model.TeamMember) error {
+	if s.teamMembers[teamId] == nil {
+		s.teamMembers[teamId] = map[string]*model.TeamMember{}
+	}
```
streamer45: The idea here is that we always want to create a new map otherwise the behaviour is the same as before. We want to replace what we have with what we get. You can take a look at `SetChannelMembers`.

PR #34, opened 2019-12-08
added GetFileThumbnail

reflog:


loadtest/user/userentity/actions.go
```diff
@@ -325,4 +325,9 @@ func (ue *UserEntity) GetTeamsUnread(teamIdToExclude string) ([]*model.TeamUnrea
 	}
 
 	return unread, nil
+}
+
+func (ue *UserEntity) GetFileThumbnail(fileId string) ([]byte, error) {
+	data, response := ue.client.GetFileThumbnail(fileId)
+	return data, response.Error
```
streamer45: We don't want to directly return `response.Error` cause of the `nil != nil` problem we encountered [before](https://community.mattermost.com/core/pl/o6ngfmj9yjb57fz7jtzwcww73a).
Let's add a simple `if` block to check for it before returning. Also, minor detail, but let's name the variable `resp` as in all other cases.

PR #37, opened 2019-12-08
added SetProfileImage

reflog:


loadtest/user/userentity/actions.go
```diff
@@ -352,6 +354,21 @@ func (ue *UserEntity) AddTeamMemberFromInvite(token, inviteId string) error {
 	return ue.store.SetTeamMember(tm.TeamId, tm)
 }
 
+func (ue *UserEntity) SetProfileImage(data []byte) error {
+	user, err := ue.getUserFromStore()
+	if err != nil {
+		return err
+	}
+	ok, resp := ue.client.SetProfileImage(user.Id, data)
+	if resp != nil {
```
streamer45: We should check for `resp.Error` here.

reflog: oops

PR #39, opened 2019-12-09
[MM-20869] Adds check-style make target and fixes lint errors

mgdelacroix:


agnivade:
Maybe an ignorant question, but I don't see the circleCI pipeline being run here. Isn't this something which needs to happen ?

mgdelacroix:
I was planning on raising the same question as I saw that we already have a `.circleci` set up and some previous work existed. Now on a second read I think the file was copied over rather than specifically set up for the project, so I'm happy to add a basic lint&test workflow to this PR. @streamer45 can you please confirm this is the right approach?

agnivade:
Oh, then we just need to add this repo to circleCI. Otherwise it won't pick it up. pinging @cpanato.

streamer45:
Yes, that was in the backlog of things to do for sure. I don't know if the current file will work without modifications. If not, feel free to open a PR.

cpanato:
enabled:
- [x] mattermod
- [x] circleci

mgdelacroix:
Thanks @cpanato!!! :smile:

cpanato:
master is broken as well, fyi

streamer45:
> master is broken as well, fyi

Yes we are aware, we need to make a PR to fix the circleci config since it's coming for a forked repository and it needs some touches.

mgdelacroix:
I'll merge this one and will create another one with the CI changes

PR #43, opened 2019-12-09
[MM-20856] Adds GetUsersByIds implementation, users store methods and tests

mgdelacroix:


loadtest/store/memstore/store.go
```diff
@@ -188,3 +190,19 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
 	s.emojis = emoji
 	return nil
 }
+
+func (s *MemStore) Users() ([]*model.User, error) {
+	users := []*model.User{}
+	for _, user := range s.users {
+		users = append(users, user)
+	}
+	return users, nil
```
streamer45: This is a detail but I think here we can leverage the length of the map, preallocate the whole slice and then set by index to avoid the append.
```go
users := make([]*model.User, len(s.users))
```

mgdelacroix: Changed in the `Users()` function as well as in `Teams()` for consistency

PR #44, opened 2019-12-10
Replace old CI config with a lint&test workflow

mgdelacroix:
#### Summary

Simple CI workflow that runs the linter and tests for all branches. This is an initial approach and it doesn't contemplate packaging nor releases

#### Ticket link
https://mattermost.atlassian.net/browse/MM-21110

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,151 +1,28 @@
 version: 2.1
-
-orbs:
-  aws-s3: circleci/aws-s3@1.0.11
-
-only_branches: &only_branches
-  filters:
-    branches:
-      ignore: master
-
-only_master: &only_master
-  filters:
-    branches:
-      only: master
-
-only_v_tags: &only_v_tags
-  filters:
-    tags:
-      only: /^v.*/
-    branches:
-      ignore: /.*/
-
-mm_binary: &mm_binary
-  mm-binary: https://releases.mattermost.com/mattermost-platform/master/mattermost-enterprise-linux-amd64.tar.gz
+executors:
+  default:
+    docker:
+      - image: circleci/golang:1.13
 
 jobs:
-  build-ltops-and-test:
-    docker:
-      - image: circleci/golang:1.12
+  lint:
+    executor:
+      name: default
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


mgdelacroix:
This is a follow up to the conversation in #39

.circleci/config.yml
```diff
@@ -1,151 +1,28 @@
...
-      - run:
-          name: Test
-          command: |
-            set +x
-            ltops --help
-      - persist_to_workspace:
-          root: .
-          paths:
-            - dist
+      - run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.21.0
```
agnivade: @lieut-data did a great job in separating the install golangci-lint step into a separate job: https://github.com/mattermost/mattermost-marketplace/blob/b7089abfd665de995d8cee45ff5f7586b42f850c/.circleci/config.yml#L30. It also uses GOBIN rather than GOPATH/bin to install the binary which is preferred, since we are moving away from all GOPATH related references.

Let's do the same here.

.circleci/config.yml
```diff
@@ -1,151 +1,28 @@
...
-      - build-docker:
-          <<: *mm_binary
-          loadtest-binary: https://releases.mattermost.com/mattermost-load-test/${CIRCLE_TAG}/mattermost-load-test.tar.gz
-          docker-tag: ${CIRCLE_TAG}
-          context: matterbuild-docker
-          <<: *only_v_tags
-          requires:
-            - upload-s3-tags
+      - lint
+      - test
```
agnivade: Why not a build step also ?

mgdelacroix: You mean running `make build`? I was planning on tackling that when we start doing something with the generated artifact, but 1/5 really

agnivade: I think it's good to check if the compilation succeeds. `make test` may not catch all scenarios. But if you plan to do tackle that soon then I'm fine.

streamer45:
I believe we can merge this and revisit it after since we don't have a real use case for building yet.

PR #45, opened 2019-12-10
Add User.SaveReaction

ethervoid:
**Summary**

This PR adds User.SaveReaction method and the reaction storage methods as well

**Ticket**

https://mattermost.atlassian.net/browse/MM-20849

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -19,7 +19,7 @@ type MemStore struct {
 	channelMembers map[string]map[string]*model.ChannelMember
 	teamMembers    map[string]map[string]*model.TeamMember
 	users          map[string]*model.User
-	reactions      map[string]map[string]map[string]*model.Reaction
+	reactions      map[string]*model.Reaction
 }
 
 func New() *MemStore {
@@ -30,7 +30,7 @@ func New() *MemStore {
 		channelMembers: map[string]map[string]*model.ChannelMember{},
 		teamMembers:    map[string]map[string]*model.TeamMember{},
 		users:          map[string]*model.User{},
-		reactions:      map[string]map[string]map[string]*model.Reaction{},
+		reactions:      map[string]*model.Reaction{},
 	}
 }
 
@@ -196,18 +196,12 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
 }
 
 func (s *MemStore) SetReaction(reaction *model.Reaction) error {
-	if s.reactions[reaction.PostId] == nil {
-		s.reactions[reaction.PostId] = map[string]map[string]*model.Reaction{}
-		s.reactions[reaction.PostId][reaction.EmojiName] = map[string]*model.Reaction{}
-	} else if s.reactions[reaction.PostId][reaction.EmojiName] == nil {
-		s.reactions[reaction.PostId][reaction.EmojiName] = map[string]*model.Reaction{}
-	}
-	s.reactions[reaction.PostId][reaction.EmojiName][reaction.UserId] = reaction
+	s.reactions[reaction.PostId] = reaction
 	return nil
 }
 
 func (s *MemStore) Reaction(postId, emojiName, userId string) (*model.Reaction, error) {
-	return s.reactions[postId][emojiName][userId], nil
+	return s.reactions[postId], nil
 }
 
 func (s *MemStore) Users() ([]*model.User, error) {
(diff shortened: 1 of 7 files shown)


loadtest/store/memstore/store.go
```diff
@@ -193,6 +195,21 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
 	return nil
 }
 
+func (s *MemStore) SetReaction(reaction *model.Reaction) error {
+	if s.reactions[reaction.PostId] == nil {
```
ethervoid: Don't know if there is a better way to do this in Go

loadtest/store/memstore/store.go
```diff
@@ -193,6 +195,21 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
...
+		s.reactions[reaction.PostId] = map[string]map[string]*model.Reaction{}
+		s.reactions[reaction.PostId][reaction.EmojiName] = map[string]*model.Reaction{}
+	} else if s.reactions[reaction.PostId][reaction.EmojiName] == nil {
+		s.reactions[reaction.PostId][reaction.EmojiName] = map[string]*model.Reaction{}
+	}
+	s.reactions[reaction.PostId][reaction.EmojiName][reaction.UserId] = reaction
```
ethervoid: I've puto `emojiName` first because I think the logic path to check for reactions is first by post and then my emoji

loadtest/store/memstore/store.go
```diff
@@ -193,6 +195,21 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
...
+	return nil
+}
```
streamer45: This looks correct but I am not sure we need to store as much information. I think that for now we should be okay with just mapping posts to reactions, so given a `PostId` we would get a slice of all reactions for that post. This would simplify a lot the code above.

loadtest/user/userentity/actions.go
```diff
@@ -426,6 +426,15 @@ func (ue *UserEntity) GetEmojiList(page, perPage int) error {
 	return ue.store.SetEmojis(emojis)
 }
 
+func (ue *UserEntity) SaveReaction(reaction *model.Reaction) error {
+	r, resp := ue.client.SaveReaction(reaction)
+	if resp.Error != nil {
+		return resp.Error
+	}
+
+	return ue.store.SetReaction(r)
```
streamer45: No need to touch the store here. The store is going to get initialized on `GetReactions` and updated via websocket.

loadtest/store/memstore/store.go
```diff
@@ -19,7 +19,7 @@ type MemStore struct {
 	channelMembers map[string]map[string]*model.ChannelMember
 	teamMembers    map[string]map[string]*model.TeamMember
 	users          map[string]*model.User
-	reactions      map[string]map[string]map[string]*model.Reaction
+	reactions      map[string]*model.Reaction
```
streamer45: This should be a map to `[]*model.Reaction`

loadtest/store/memstore/store.go
```diff
@@ -196,18 +196,12 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
...
-	}
-	s.reactions[reaction.PostId][reaction.EmojiName][reaction.UserId] = reaction
+	s.reactions[reaction.PostId] = reaction
 	return nil
 }
 
 func (s *MemStore) Reaction(postId, emojiName, userId string) (*model.Reaction, error) {
-	return s.reactions[postId][emojiName][userId], nil
+	return s.reactions[postId], nil
 }
```
streamer45: This function should probably be named `Reactions` and return the slice we are storing for that `postId` which would be the only parameter.

loadtest/store/memstore/store.go
```diff
@@ -196,18 +196,12 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
...
 
 func (s *MemStore) SetReaction(reaction *model.Reaction) error {
-	if s.reactions[reaction.PostId] == nil {
-		s.reactions[reaction.PostId] = map[string]map[string]*model.Reaction{}
-		s.reactions[reaction.PostId][reaction.EmojiName] = map[string]*model.Reaction{}
-	} else if s.reactions[reaction.PostId][reaction.EmojiName] == nil {
-		s.reactions[reaction.PostId][reaction.EmojiName] = map[string]*model.Reaction{}
-	}
-	s.reactions[reaction.PostId][reaction.EmojiName][reaction.UserId] = reaction
+	s.reactions[reaction.PostId] = reaction
```
streamer45: This should be named `SetReactions` and take a `postId` and a `[]*model.Reaction`.

ethervoid:
@streamer45 thank you for the patience and sorry for the mess with the previous fix :hugs:

PR #46, opened 2019-12-10
Implements User.GetPostsForChannel

ethervoid:
**Summary**

This PR implements the GetPostsForChannel method and also adds the storage methods to dump a slice of posts in the storage

**Ticket**
https://mattermost.atlassian.net/browse/MM-20850

### DIFF
--- a/loadtest/store/memstore/store_test.go
+++ b/loadtest/store/memstore/store_test.go
@@ -85,17 +85,20 @@ func TestUser(t *testing.T) {
 	})
 
 	t.Run("ChannelPosts", func(t *testing.T) {
-		err := s.SetPosts([]*model.Post{})
-		require.Error(t, err)
 		channelId := model.NewId()
 		postId := model.NewId()
+		err := s.SetPosts([]*model.Post{})
+		require.Error(t, err)
+		channelPosts, err := s.ChannelPosts(channelId)
+		require.NoError(t, err)
+		require.Nil(t, channelPosts)
 		p := []*model.Post{
 			&model.Post{Id: postId, ChannelId: channelId},
 			&model.Post{Id: model.NewId(), ChannelId: model.NewId()},
 		}
 		err = s.SetPosts(p)
 		require.NoError(t, err)
-		channelPosts, err := s.ChannelPosts(channelId)
+		channelPosts, err = s.ChannelPosts(channelId)
 		require.NoError(t, err)
 		require.Equal(t, len(channelPosts), 1)
 		require.Equal(t, postId, channelPosts[0].Id)
--- a/loadtest/store/memstore/store_test.go
+++ b/loadtest/store/memstore/store_test.go
@@ -71,6 +71,19 @@ func TestUser(t *testing.T) {
 		require.Equal(t, p, uu)
 	})
 
+	t.Run("SetPosts", func(t *testing.T) {
+		err := s.SetPosts(nil)
+		require.Error(t, err)
+		err = s.SetPosts([]*model.Post{})
+		require.Error(t, err)
+		p := []*model.Post{&model.Post{Id: model.NewId()}}
+		err = s.SetPosts(p)
+		require.NoError(t, err)
+		uu, err := s.Post(p[0].Id)
+		require.NoError(t, err)
+		require.Equal(t, p[0], uu)
+	})
+
 	t.Run("SetReactions", func(t *testing.T) {
 		postId := model.NewId()
 		userId := model.NewId()
--- a/loadtest/store/memstore/store_test.go
+++ b/loadtest/store/memstore/store_test.go
@@ -84,6 +84,23 @@ func TestUser(t *testing.T) {
 		require.Equal(t, p[0], uu)
 	})
 
+	t.Run("ChannelPosts", func(t *testing.T) {
+		err := s.SetPosts([]*model.Post{})
+		require.Error(t, err)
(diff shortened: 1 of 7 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -142,6 +142,24 @@ func (ue *UserEntity) SearchPosts(teamId, terms string, isOrSearch bool) (*model
 	return postList, nil
 }
 
+func (ue *UserEntity) GetPostsForChannel(channelId string) ([]string, error) {
```
ethervoid: This is the most controversial change and I want to explain why I did it this way:

All the Posts search returns PostList but in our case we only want to have store the Posts (by now). We're storing them in a slice indexed by PostId so there is no need to use the PostList. 

Also we don't want to tie our storage to the API model so there is no need to pass the PostList or even the Map.

As we made in other method, I built the ids for the posts and returned them in a plain slice

ethervoid: I was thinking to add the `ChannelPosts` method but I didn't know if that would be too much logic for now or not. Chaging it...

loadtest/user/userentity/actions.go
```diff
@@ -142,6 +142,24 @@ func (ue *UserEntity) SearchPosts(teamId, terms string, isOrSearch bool) (*model
...
+	postlist, resp := ue.client.GetPostsForChannel(channelId, 0, 9999, "")
```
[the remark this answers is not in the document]
ethervoid: Sure

ethervoid:
I understand `ChannelPosts` is going to be used in other methods that will be implemented in the future

loadtest/store/memstore/store_test.go
```diff
@@ -71,6 +71,39 @@ func TestUser(t *testing.T) {
...
+		p := []*model.Post{&model.Post{Id: model.NewId()}}
+		err = s.SetPosts(p)
+		require.NoError(t, err)
+		uu, err := s.Post(p[0].Id)
+		require.NoError(t, err)
+		require.Equal(t, p[0], uu)
+	})
+
+	t.Run("ChannelPosts", func(t *testing.T) {
+		channelId := model.NewId()
```
ethervoid: Question: Is there a way to have a `AllTest` method or handler to create a new instance of the storage in every test `s := New()` instead at the beginning of the method?

loadtest/user/userentity/actions.go
```diff
@@ -142,6 +142,18 @@ func (ue *UserEntity) SearchPosts(teamId, terms string, isOrSearch bool) (*model
...
 
+func (ue *UserEntity) GetPostsForChannel(channelId string, page, perPage int) error {
+	postlist, resp := ue.client.GetPostsForChannel(channelId, page, perPage, "")
+	if resp.Error != nil {
+		return resp.Error
+	}
+	posts := make([]*model.Post, 0, len(postlist.Posts))
+	for _, v := range postlist.Posts {
+		posts = append(posts, v)
+	}
```
streamer45: Since we've already allocated we can avoid the append and simply assign by index.

loadtest/store/memstore/store_test.go
streamer45: You mean before every subtests? If so I am not sure you can. 
We could probably leverage [this](https://golang.org/pkg/testing/#hdr-Main) to init the store just once before every test but since the cost of creating a new *in memory* store is not that great I don't believe the extra complexity is justified.

loadtest/user/userentity/actions.go
ethervoid: Ok, good to know :)

loadtest/store/memstore/store_test.go
ethervoid: I tried in a few days ago to have that running but I wasn't able. I'll try again.

The problem is not the performance cost but to have a clean state in every test. In some storage methods we're appending and not replacing the data and we could find that data which is not related with our current test in memory

loadtest/user/userentity/actions.go
ethervoid: But there is not numeric index because the list is a map. You mean to include a counter and increment it instead of doing append?

streamer45: Yes, sorry for the confusion. We did something similar in https://github.com/mattermost/mattermost-load-test-ng/pull/43.

loadtest/store/memstore/store_test.go
streamer45: > I tried in a few days ago to have that running but I wasn't able. I'll try again.
> 
> The problem is not the performance cost but to have a clean state in every test. In some storage methods we're appending and not replacing the data and we could find that data which is not related with our current test in memory

Yes, that depends on a test to test basis but normally we want to have an empty state. Were you having issues creating a new store inside the subtest?

ethervoid: I did it [here](https://github.com/mattermost/mattermost-load-test-ng/blob/master/loadtest/store/memstore/store_test.go#L186) and was the way to do it. Creating a new store in every test

loadtest/user/userentity/actions.go
ethervoid: Done

loadtest/store/memstore/store_test.go
streamer45: Looks good to me :+1:

PR #50, opened 2019-12-12
added RemoveTeamMember

reflog:
#### Summary
added RemoveTeamMember

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-20860

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -190,6 +190,14 @@ func (s *MemStore) RemoveChannelMember(channelId string, userId string) error {
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


loadtest/store/memstore/store.go
```diff
@@ -190,6 +190,14 @@ func (s *MemStore) RemoveChannelMember(channelId string, userId string) error {
 	return nil
 }
 
+func (s *MemStore) RemoveTeamMember(teamId string, userId string) error {
+	if s.teamMembers[teamId] == nil {
+		return nil
+	}
```
streamer45: I think we are safe without this block. You can call delete on an empty/nil map.

PR #51, opened 2019-12-13
Use empty struct in the stop chan signal

agnivade:
#### Summary
Using an empty struct is more idiomatic and also better
because it does not take up any space.


agnivade:
It just makes it more idiomatic and consistent with the rest of Go codebases, and makes the API bulletproof. Currently, you can do `controller.stop <- true` or `controller.stop <- false` and it does not make a difference, which is slightly weird. Using an empty struct removes the chances of making such mistakes.

PR #52, opened 2019-12-13
Implements User.GetPostsBefore

ethervoid:
#### Summary

This PRs adds the method `GetPostsBefore` to the user entity

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21026

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -147,21 +147,25 @@ func (ue *UserEntity) GetPostsForChannel(channelId string, page, perPage int) er
 	if resp.Error != nil {
 		return resp.Error
 	}
-	posts := make([]*model.Post, 0, len(postlist.Posts))
-	i := 0
-	for _, v := range postlist.Posts {
-		posts[i] = v
-		i++
-	}
-	return ue.store.SetPosts(posts)
+	return ue.store.SetPosts(ue.postsMapToSlice(postlist.Posts))
 }
 
 func (ue *UserEntity) GetPostsBefore(channelId, postId string, page, perPage int) error {
-	_, resp := ue.client.GetPostsBefore(channelId, postId, page, perPage, "")
+	postlist, resp := ue.client.GetPostsBefore(channelId, postId, page, perPage, "")
 	if resp.Error != nil {
 		return resp.Error
 	}
-	return nil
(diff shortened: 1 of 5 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -156,6 +156,14 @@ func (ue *UserEntity) GetPostsForChannel(channelId string, page, perPage int) er
 	return ue.store.SetPosts(posts)
 }
 
+func (ue *UserEntity) GetPostsBefore(channelId, postId string, page, perPage int) error {
+	_, resp := ue.client.GetPostsBefore(channelId, postId, page, perPage, "")
+	if resp.Error != nil {
+		return resp.Error
+	}
+	return nil
```
ethervoid: Don't add it to the storage until we have an append method in the storage because its a partial result

streamer45: Thinking it twice, isn't the `MemStore.SetPosts` already what we need here? We are simply merging posts (adding or updating).

ethervoid: Oh right, we're adding by postId index instead of replacing the whole post slice. Adding that part then :)

loadtest/user/userentity/actions.go
```diff
@@ -147,13 +147,25 @@ func (ue *UserEntity) GetPostsForChannel(channelId string, page, perPage int) er
...
+
+func (ue *UserEntity) GetPostsBefore(channelId, postId string, page, perPage int) error {
+	postlist, resp := ue.client.GetPostsBefore(channelId, postId, page, perPage, "")
+	if resp.Error != nil {
+		return resp.Error
+	}
+	return ue.store.SetPosts(ue.postsMapToSlice(postlist.Posts))
+}
+
+func (ue *UserEntity) postsMapToSlice(postsMap map[string]*model.Post) []*model.Post {
```
ethervoid: Given that we have three methods doing the same thing I think is better to extract it to a new method. I've done in a non-exported method inside userentity but maybe we can have a helper or something. What do you think?

streamer45: Yes. It shouldn't be a method of `UserEntity` but a simple function. We can make a `utils.go` inside `userentity` package.

ethervoid: Done

PR #53, opened 2019-12-13
Implements User.GetChannelsForTeam

ethervoid:
#### Summary

This PRs adds:
- User.GetChannelsForTeam method to the user entity
- SetChannels method for the storage

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21131

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -242,8 +242,12 @@ func (ue *UserEntity) GetChannel(channelId string) error {
 	return ue.store.SetChannel(channel)
 }
 
-func (ue *UserEntity) GetChannelsForTeam(teamId, userId string) error {
-	channels, resp := ue.client.GetChannelsForTeamForUser(teamId, userId, "")
+func (ue *UserEntity) GetChannelsForTeam(teamId string) error {
+	user, err := ue.store.User()
+	if err != nil {
+		return err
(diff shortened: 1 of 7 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -242,6 +242,15 @@ func (ue *UserEntity) GetChannel(channelId string) error {
 	return ue.store.SetChannel(channel)
 }
 
+func (ue *UserEntity) GetChannelsForTeam(teamId, userId string) error {
```
streamer45: This method should only need a `teamId` since the `userId` should be the one of the current user and should be taken from the internal state.

agnivade: Wait, this is the same as my PR https://github.com/mattermost/mattermost-load-test-ng/pull/58. We should probably make the method name the same as the underlying client API name to avoid confusion.

streamer45: We applied the same convention to previous implementations and I would keep it that way. The idea is that we add the suffix `ForUser` when we need to specify the `userId`. There are cases where we want to have them both and one would wrap the other to simplify usage. An example is `GetProfileImage` and `GetProfileImageForUser`.

ethervoid: Changed :)

PR #55, opened 2019-12-13
MM-21143: implement SimpleController.reload

agnivade:
#### Summary
Added all the methods that we currently have that happens
when a user performs a reload.

Marked the rest of them as TODO.

While at it, we also move the getter methods from MutableUserStore
to UserStore to be able to access from the controller.

Tested with this piece of code. (Needs the server running)

```go
func TestReload(t *testing.T) {
	c := SimpleController{}
	c.Init(userentity.New(memstore.New(), 0, userentity.Config{
	}))

	status := c.signUp()
	t.Logf("%+v", status)

	status = c.login()
	t.Logf("%+v", status)

	status = c.reload()
	t.Logf("%+v", status)
}
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21143

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -110,44 +110,30 @@ func (c *SimpleController) viewChannel() control.UserStatus {
 }
 
 func (c *SimpleController) reload() control.UserStatus {
-	teams, err := c.user.Store().Teams()
-	if err != nil {
-		return c.newErrorStatus(err)
-	}
-
 	// Getting preferences.
-	err = c.user.GetPreferences()
+	err := c.user.GetPreferences()
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
 
+	prefs, _ := c.user.Store().Preferences()
+	userIds := make([]string, len(prefs))
 	chanId := ""
-	if len(teams) > 0 {
-		// Assuming this is a browser and the user has selected the
-		// first team as the default.
-		t := teams[0]
-		// TODO: This will be populated by GetChannelsForTeamForUser API
-		// which is to be implemented. For now, we assume that the user
-		// already belongs to some channels, and they are stored in the
-		// store's memory.
-		chans, err := c.user.Store().Channels(t.Id)
+	for i, p := range prefs {
+		if p.Name == model.PREFERENCE_NAME_LAST_CHANNEL {
+			chanId = p.Value
+		}
+		userIds[i] = p.UserId
+	}
+
+	if chanId != "" {
+		// Marking the channel as viewed
+		_, err := c.user.ViewChannel(&model.ChannelView{
+			ChannelId:     chanId,
+			PrevChannelId: "",
+		})
 		if err != nil {
 			return c.newErrorStatus(err)
-		}
-		for _, ch := range chans {
-			// Assuming the first channel to be the currently
-			// selected one.
-			if chanId == "" {
-				chanId = ch.Id
-			}
-			// Marking the channels as viewed
-			_, err := c.user.ViewChannel(&model.ChannelView{
-				ChannelId:     ch.Id,
-				PrevChannelId: "",
-			})
-			if err != nil {
-				return c.newErrorStatus(err)
-			}
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


loadtest/control/simplecontroller/actions.go
```diff
@@ -108,3 +109,100 @@ func (c *SimpleController) viewChannel() control.UserStatus {
...
+	// TODO: GetChannelMembersForUser
+
+	// Getting unread teams.
+	_, err = c.user.GetTeamsUnread("")
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+
+	// Get users by Ids.
+	ids, err := c.user.GetUsersByIds([]string{strconv.Itoa(c.user.Id())})
```
streamer45: As discussed `user.Id` is not related to the API. Also `GetUsersByIds()` is usually called with a list of ids taken from the output of `GetPreferences()` which lists things like open direct messaging channels.

loadtest/control/simplecontroller/actions.go
```diff
@@ -108,3 +109,100 @@ func (c *SimpleController) viewChannel() control.UserStatus {
...
+			// Marking the channels as viewed
+			_, err := c.user.ViewChannel(&model.ChannelView{
+				ChannelId:     ch.Id,
+				PrevChannelId: "",
+			})
+			if err != nil {
+				return c.newErrorStatus(err)
+			}
+		}
+	}
```
streamer45: I believe this code should go after we get the preferences because in this action we mainly act based on that info and with that result we decide what to fetch from the API.  
Also I would avoid viewing all channels since that is a behaviour which is not strictly related to this function which should be just about reloading. Viewing the last viewed channel (taken from preferences) or a default should be enough.

loadtest/control/simplecontroller/actions.go
agnivade: Ok, I will need to move the `Preferences` getter to `UserStore` too.

loadtest/control/simplecontroller/actions.go
agnivade: Yeah, but I did see ViewChannel requests for all channels on a reload. Do you think it was the webapp doing things apart from the reload functionality ?

streamer45: > Yeah, but I did see ViewChannel requests for all channels on a reload. Do you think it was the webapp doing things apart from the reload functionality ?

Usually I believe you need to click on an channel's name in the LHS in order to view it. It seems strange it's being called preemptively. That said we should definitely take some extra [mana](https://docs.mattermost.com/process/training.html#mana) to investigate this so we properly mimic the behaviour.

loadtest/store/memstore/store.go
```diff
@@ -53,8 +53,10 @@ func (s *MemStore) SetUser(user *model.User) error {
 	return nil
 }
 
-func (s *MemStore) Preferences() (*model.Preferences, error) {
-	return s.preferences, nil
+func (s *MemStore) Preferences() (model.Preferences, error) {
+	newPref := make(model.Preferences, len(*s.preferences))
```
streamer45: I'd like to add a check for `s.preferences != nil` so we don't run into troubles.

loadtest/control/simplecontroller/actions.go
```diff
@@ -108,3 +108,87 @@ func (c *SimpleController) viewChannel() control.UserStatus {
...
+
+	prefs, _ := c.user.Store().Preferences()
+	userIds := make([]string, len(prefs))
+	chanId := ""
+	for i, p := range prefs {
+		if p.Name == model.PREFERENCE_NAME_LAST_CHANNEL {
+			chanId = p.Value
+		}
+		userIds[i] = p.UserId
+	}
```
streamer45: There's a slight problem with this. The `p.UserId` we are using is always going to be the current user's id.  
What we are looking for here could be, for example, the users with whom we have direct channels open.  
To find that info we need to look for `p.Category == PREFERENCE_CATEGORY_DIRECT_CHANNEL_SHOW` and then for `p.Name` which should hold that user's id.  
If this doesn't make 100% sense I agree but it looks that's how the API is currently working.

agnivade: Yeah that's right actually. I am not fully versed with the APIs, so I will defer to your judgement. :+1:

PR #56, opened 2019-12-13
Adds GetProfileImage to user entity

ethervoid:
#### Summary

This PR adds the `GetProfileImage` method to the user entity

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21132


streamer45:
Okay this is one of those cases where we actually want to get data related to a different user. I propose we also add a `GetProfileImageForUser` that gets a `userId` as input.

ethervoid:
@streamer45 added that change

PR #59, opened 2019-12-16
MM-21233: Implement User.GetTeamsForUser

agnivade:
#### Summary

Implement User.GetTeamsForUser

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21233

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -332,6 +332,21 @@ func (ue *UserEntity) GetTeam(teamId string) error {
 	return ue.store.SetTeam(team)
 }
 
+func (ue *UserEntity) GetTeamsForUser(userId string) ([]*model.Team, error) {
+	teams, resp := ue.client.GetTeamsForUser(userId, "")
+	if resp.Error != nil {
+		return nil, resp.Error
+	}
+
+	for _, t := range teams {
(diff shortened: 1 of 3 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -332,6 +332,21 @@ func (ue *UserEntity) GetTeam(teamId string) error {
 	return ue.store.SetTeam(team)
 }
 
+func (ue *UserEntity) GetTeamsForUser(userId string) ([]*model.Team, error) {
+	teams, resp := ue.client.GetTeamsForUser(userId, "")
```
streamer45: As previously discussed in #53 we should rename this to `GetTeams` and leverage the current user's id taken from the store.

agnivade: @streamer45 - Then that functionality is already implemented here https://github.com/mattermost/mattermost-load-test-ng/blob/03663efb16c1b1230f5a291117abbd8eb54f407f/loadtest/user/userentity/actions.go#L574. Should we close this issue then ?

streamer45: What I was suggesting was to make that function you highlighted just a wrapper of your `GetTeamsForUser` which while it won't see a lot of usage it could still come handy in the future as an *admin only* method.

PR #64, opened 2019-12-17
MM-21282: Remove vendoring from repo

agnivade:
#### Summary
And while at it, work on other small improvements in the Makefile:
- Remove -i flag from go install. It wasn't necessary, and has
been already removed in the main repo.
- Add the -trimpath flag newly introduced in 1.13
- Remove PACKAGES variable. It was only needed to pass the list of
packages which is anyways available through "./...". So it avoids
the need to run another "go list" to run the tests. This is an
artifact copied from the main repo, where "./..." does not work
because the enterprise repo is a symlink so we need to explicitly
pass the package list for the Go toolchain to pick it up.
- Remove -run from tests. It is not necessary.
- Remove GO_LINKER_FLAGS. It wasn't used anywhere.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21282


agnivade:
Requesting Mario to take a look since Eli is out.

PR #68, opened 2019-12-18
Implements GetConfig in the user entity

ethervoid:
#### Summary

This PR adds the method `GetConfig`to the user entity. It doesn't do anything with configuration because we only want to test the load for that endpoint so we don't need to store the result in the cache

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21273


ethervoid:
Are we going to modify the state of the server from the load test? If not we could store it as inmutable in the initialization of the controller otherwise I'll add it to the storage

streamer45:
> Are we going to modify the state of the server from the load test? If not we could store it as inmutable in the initialization of the controller otherwise I'll add it to the storage

I think we should always go through the `UserEntity`. I'd like to avoid calling the API directly in the controller. So I would say to keep a `*model.Config` in the `MemStore` and return a copy in the getter function.

ethervoid:
> I would say to keep a *model.Config in the MemStore

I have a doubt/question about this. Shouldn't we have some kind of ServerStore? I mean, the server is a unique instance so we maybe want to have it synchronized for all the user entities that are going to interact with it. What do you think?

streamer45:
> > I would say to keep a *model.Config in the MemStore
> 
> I have a doubt/question about this. Shouldn't we have some kind of ServerStore? I mean, the server is a unique instance so we maybe want to have it synchronized for all the user entities that are going to interact with it. What do you think?

The API is the only source of truth and the users are made 100% independent from each other's by design. We don't want to enforce synchronization between them. Each controller is responsible for keeping its own user in sync with the server. If it needs something it should fetch it using the API.

loadtest/store/store.go
```diff
@@ -11,6 +11,8 @@ type UserStore interface {
 	Id() string
 	// TODO: Move all getters to this interface
 
+	// Config return the server configuration settings
+	Config() *model.Config
```
streamer45: Ideally this should return a copy cause as it is it's mutable but given it's a config object which is only going to be used from the outside I think it's fine to return like this, at least for now.

ethervoid: Changed. Let me know it it's ok :D

streamer45: Sorry, what I meant is that even like this it's still mutable cause `model.Config` is a deeply nested struct full of pointers. We would have to make a deep copy to make it really immutable but it's okay like this for now. We can merge this :+1:

PR #70, opened 2019-12-18
Add more module checks to CI

agnivade:
#### Summary
Now that we have moved to normal module fetch workflow, some additional
commands are needed to verify everything is working okay.

- We add a "go mod verify" command which matches the checksums in the DB
with the values in go.sum
- We also use the -mod=readonly flag during testing and building to ensure
that no additional modifications to go.mod happen during those commands.
This is similar to the "npm ci" command.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,6 +55,7 @@ jobs:
       - *restore_cache
       - run: make check-style
       - run: make test
+      - run: make verify-gomod
       - *save_cache
 
 workflows:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,9 +53,9 @@ jobs:
       - checkout
       - install-golangci-lint
       - *restore_cache
+      - run: make verify-gomod
       - run: make check-style
       - run: make test
-      - run: make verify-gomod
       - *save_cache
 
 workflows:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -55,6 +55,7 @@ jobs:
       - *restore_cache
       - run: make check-style
       - run: make test
+      - run: make verify-gomod
```
streamer45: Just wondering, why verify after running the tests?

agnivade: Right, this looks odd. It is needed after the tests because we need the modules to be downloaded before they can be verified. "go test" does that. An alternative is to run "go mod download" first to prefill the cache. But that looked like an extra step, and this is simple enough.

streamer45: I see. I guess it wouldn't really be an extra step but the same done explicitly before everything else and not hidden inside the test step. My worry is about the logic of this more than anything. If we verify after `make test` it means waiting for all that process to complete only to discover that it's not valid because some modules mismatched. I'd rather know that from the start, as soon as we downloaded what we need. Does that make sense?

agnivade: That makes sense. Will update :+1:

PR #72, opened 2019-12-18
MM-21232: Implement User.GetChannelMembersForUser

agnivade:
#### Summary

Implement User.GetChannelMembersForUser

Tested with:
```go

func TestNewAPI(t *testing.T) {
	c := SimpleController{}
	c.Init(userentity.New(memstore.New(), 0, userentity.Config{
	}))
	status := c.signUp()
	require.Nil(t, status.Err)
	status = c.login()
	require.Nil(t, status.Err)
	userId := c.user.Store().Id()
	teamId, err := c.user.CreateTeam(&model.Team{
		Name:        "myteam",
		DisplayName: "myteam",
		Type:        model.TEAM_OPEN,
	})
	require.Nil(t, err)
	err = c.user.AddTeamMember(teamId, userId)
	require.Nil(t, err)
	channelId, err := c.user.CreateChannel(&model.Channel{
		Name:        "mychannel",
		DisplayName: "mychannel",
		TeamId:      teamId,
		Type:        model.CHANNEL_OPEN,
	})
	require.Nil(t, err)
	err = c.user.AddChannelMember(channelId, userId)
	require.Nil(t, err)

	err = c.user.GetChannelMembersForUser(userId, teamId)
	require.Nil(t, err)

	members, err := c.user.Store().ChannelMembers(channelId)
	require.Nil(t, err)
	t.Log(members)
}
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21232

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -176,16 +176,24 @@ func (s *MemStore) SetTeams(teams []*model.Team) error {
 	return nil
 }
 
-func (s *MemStore) SetChannelMembers(channelId string, channelMembers *model.ChannelMembers) error {
+// SetChannelMembers stores the given channel members in the store.
+func (s *MemStore) SetChannelMembers(channelMembers *model.ChannelMembers) error {
 	if channelMembers == nil {
 		return errors.New("channelMembers should not be nil")
 	}
-	membersMap := make(map[string]*model.ChannelMember)
-	members := *channelMembers
-	for _, m := range members {
-		membersMap[m.UserId] = &m
+
+	for _, member := range *channelMembers {
+		// Initialize the maps as necessary.
+		if s.channelMembers == nil {
+			s.channelMembers = make(map[string]map[string]*model.ChannelMember)
+		}
+		if s.channelMembers[member.ChannelId] == nil {
+			s.channelMembers[member.ChannelId] = make(map[string]*model.ChannelMember)
+		}
+		// Set value.
+		s.channelMembers[member.ChannelId][member.UserId] = &member
 	}
-	s.channelMembers[channelId] = membersMap
+
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -339,6 +339,36 @@ func (ue *UserEntity) GetChannelMembers(channelId string, page, perPage int) err
...
+			chanIdMap[member.ChannelId] = &chanMembers
+		}
+	}
+
+	for chanId, members := range chanIdMap {
+		err := ue.store.SetChannelMembers(chanId, members)
+		if err != nil {
+			return err
+		}
+	}
```
streamer45: I kind of feel like this logic belongs to the `Store` layer.  
I propose we get rid of the redundant `channelId` parameter in `Store.SetChannelMembers()` and integrate this code there so that it would work for both API calls by using the data (`ChannelId`, `TeamId`) found in the `model.ChannelMember` structs.

agnivade: Agree.

loadtest/store/memstore/store.go
```diff
@@ -176,16 +176,24 @@ func (s *MemStore) SetTeams(teams []*model.Team) error {
...
-		membersMap[m.UserId] = &m
+
+	for _, member := range *channelMembers {
+		// Initialize the maps as necessary.
+		if s.channelMembers == nil {
+			s.channelMembers = make(map[string]map[string]*model.ChannelMember)
+		}
+		if s.channelMembers[member.ChannelId] == nil {
+			s.channelMembers[member.ChannelId] = make(map[string]*model.ChannelMember)
+		}
```
streamer45: Just pointing out that we are *slightly* changing the behaviour with these changes.

Before we would replace this map whether or not it was nil. That was done to keep the state consistent since we currently don't keep it updated through WebSocket and at each `SimpleController.reload()` we expect to get the store loaded properly.

That said, this is a recurring problem, especially when we need to store partial results and we cannot just replace what we have so I am thinking a good option might be to expose some functionality to clear the store (or part of it) and use that when needed such as in the `reload()` case.

agnivade: Right. I can always do a `s.channelMembers = make(map[string]map[string]*model.ChannelMember)` unconditionally to reset the map. But I agree that we need partial updates also, as that is more close to how the client will behave.

Keeping this as-is for now.

PR #73, opened 2019-12-19
MM-21283: Implement websocket listener

agnivade:
#### Summary

- The listener listens on the Event and Response channels.
And keeps reconnecting unless the Disconnect method was called explicitly.

- We also expose a SendWebsocketMessage method if in future we want to send
a message to the server.

Tested properly by checking for race conditions:

`$go test -race -v -run=UserConnect -count=1 .`

```go
func TestUserConnect(t *testing.T) {
	user := New(memstore.New(), 0, Config{
	})

	email := fmt.Sprintf("testuser%d@example.com", user.Id())
	username := fmt.Sprintf("testuser%d", user.Id())
	password := "testPass123$"

	err := user.SignUp(email, username, password)
	require.Nil(t, err)

	err = user.Login()
	require.Nil(t, err)

	err = user.Connect()
	require.Nil(t, err)
	t.Log("user connected")

	user.SendWebsocketMessage("user_typing", map[string]interface{}{"key": "value"})

	t.Log("sleeping")
	time.Sleep(10 * time.Second)

	err = user.Disconnect()
	t.Log("user disconnected")
	require.Nil(t, err)
}
```

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21283

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -41,10 +41,12 @@ func (c *SimpleController) login() control.UserStatus {
 		return c.newErrorStatus(err)
 	}
 
-	err = c.user.Connect()
-	if err != nil {
-		return c.newErrorStatus(err)
-	}
+	errChan := c.user.Connect()
+	go func() {
+		for err := range errChan {
+			c.status <- c.newErrorStatus(err)
+		}
+	}()
 
 	return c.newInfoStatus("logged in")
 }
--- a/loadtest/control/simplecontroller/controller.go
+++ b/loadtest/control/simplecontroller/controller.go
@@ -13,9 +13,10 @@ import (
 )
 
 type SimpleController struct {
-	user user.User
-	stop chan struct{}
-	rate float64
+	user   user.User
+	stop   chan struct{}
+	status chan<- control.UserStatus
+	rate   float64
 }
 
 func (c *SimpleController) Init(user user.User) {
@@ -29,6 +30,9 @@ func (c *SimpleController) Run(status chan<- control.UserStatus) {
 		c.sendFailStatus(status, "controller was not initialized")
 		return
 	}
+	// TODO: This needs to be revamped. Status needs to be passed during
+	// initialization.
+	c.status = status
 
 	actions := []UserAction{
 		{
--- a/loadtest/user/userentity/user.go
+++ b/loadtest/user/userentity/user.go
@@ -22,6 +22,7 @@ type UserEntity struct {
 	wsClient    *model.WebSocketClient
 	closing     chan struct{}
 	closed      chan struct{}
+	wsErrorChan chan error
 	config      Config
 }
 
@@ -60,21 +61,24 @@ func New(store store.MutableUserStore, id int, config Config) *UserEntity {
 	ue.store = store
 	ue.closing = make(chan struct{})
 	ue.closed = make(chan struct{})
+	ue.wsErrorChan = make(chan error, 1)
 	return &ue
 }
 
 // Connect creates a websocket connection to the server and starts listening for messages.
-func (ue *UserEntity) Connect() error {
+func (ue *UserEntity) Connect() chan error {
 	if ue.client.AuthToken == "" {
-		return errors.New("user is not authenticated")
+		ue.wsErrorChan <- errors.New("user is not authenticated")
+		return ue.wsErrorChan
 	}
 	cli := ue.getWsClient()
 	if cli != nil {
-		return errors.New("user is already connected")
+		ue.wsErrorChan <- errors.New("user is already connected")
+		return ue.wsErrorChan
 	}
 
-	go ue.listen()
-	return nil
+	go ue.listen(ue.wsErrorChan)
+	return ue.wsErrorChan
 }
 
 // Disconnect closes the websocket connection.
@@ -83,21 +87,22 @@ func (ue *UserEntity) Disconnect() error {
 	if cli == nil {
 		return errors.New("user is not connected")
 	}
+	// We exit the listener loop first, and then close the connection.
+	// Otherwise, it tries to reconnect first, and then
+	// exits, which causes unnecessary delay.
+	close(ue.closing)
+
+	// We wait to get the response with a timeout, because
+	// the loop may be sleeping on a reconnect cycle.
+	select {
+	case <-ue.closed:
+	case <-time.After(minWebsocketReconnectDuration):
+	}
+
+	close(ue.wsErrorChan)
+
 	cli.Close()
 	cli = nil
-	// Wait for the listener to shut down.
-	close(ue.closing)
-	<-ue.closed
-	return nil
-}
-
-// SendWebsocketMessage sends a given action type with data to the websocket.
-func (ue *UserEntity) SendWebsocketMessage(action string, data map[string]interface{}) error {
-	cli := ue.getWsClient()
-	if cli == nil {
-		return errors.New("user is not connected")
-	}
-	cli.SendMessage(action, data)
 	return nil
 }
 
--- a/loadtest/user/userentity/user.go
+++ b/loadtest/user/userentity/user.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"net"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store"
@@ -14,11 +15,14 @@ import (
 )
 
 type UserEntity struct {
-	id       int
-	store    store.MutableUserStore
-	client   *model.Client4
-	wsClient *model.WebSocketClient
-	config   Config
+	id          int
+	store       store.MutableUserStore
+	client      *model.Client4
+	wsClientMut sync.RWMutex
+	wsClient    *model.WebSocketClient
+	closing     chan struct{}
+	closed      chan struct{}
+	config      Config
 }
 
 type Config struct {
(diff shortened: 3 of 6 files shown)


loadtest/user/userentity/websocket.go
```diff
@@ -0,0 +1,85 @@
...
+func (ue *UserEntity) listen() {
+	connectionFailCount := 0
+start:
+	client, err := model.NewWebSocketClient4(ue.config.WebSocketURL, ue.client.AuthToken)
+	if err != nil {
+		mlog.Error("WebsocketClient creation error:", mlog.Err(err))
+		connectionFailCount++
+		time.Sleep(getWaitTime(connectionFailCount))
+		// Reconnect again.
+		goto start
```
ethervoid: Question: Why use a goto instead a for {} and continue with label?

loadtest/user/userentity/websocket.go
```diff
@@ -0,0 +1,85 @@
...
+			_ = ev // TODO: handle event
+		case resp, ok := <-ue.wsClient.ResponseChannel:
+			if !ok {
+				chanClosed = true
+				break
+			}
+			_ = resp // TODO: handle response
+		case <-ue.closing:
+			// Explicit disconnect. Return.
+			close(ue.closed)
```
ethervoid: Maybe add a `mLog.Info`? 1/5

loadtest/user/userentity/user.go
```diff
@@ -54,35 +58,56 @@ func New(store store.MutableUserStore, id int, config Config) *UserEntity {
...
+	// Wait for the listener to shut down.
+	close(ue.closing)
+	<-ue.closed
+	return nil
+}
+
+// SendWebsocketMessage sends a given action type with data to the websocket.
+func (ue *UserEntity) SendWebsocketMessage(action string, data map[string]interface{}) error {
+	cli := ue.getWsClient()
+	if cli == nil {
```
ethervoid: Maybe we can do this comparison in the `getWsClient`? 2/5

loadtest/user/userentity/websocket.go
agnivade: Sigh .. what was I thinking. Yes obviously that is much better. Will update.

loadtest/user/userentity/websocket.go
```diff
@@ -0,0 +1,85 @@
...
+	maxWebsocketFails             = 7
+)
+
+// listen starts to listen for messages on various channels.
+// It will keep reconnecting if the connection closes.
+// Only on calling Disconnect explicitly, it will return.
+func (ue *UserEntity) listen() {
+	connectionFailCount := 0
+start:
+	client, err := model.NewWebSocketClient4(ue.config.WebSocketURL, ue.client.AuthToken)
```
streamer45: Do we need to create a new client every time we disconnect? Calling the `Listen()` method should be enough.

loadtest/user/userentity/websocket.go
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See License.txt for license information.
+
+package userentity
+
+import (
+	"time"
+
+	"github.com/mattermost/mattermost-server/v5/mlog"
```
streamer45: I'd like to avoid logging in this package. All information should go through the upper layer and eventually become a `control.UserStatus`. Since we are mainly interested in errors we could probably return a `<-chan error` on `Connect()` and use that in the controller to handle related errors.

agnivade: That's much better. But I'm wondering how will we handle that in the controller though. As of now, each action function returns a status, which is channeled up to the controller running it. But now if user.Connect also returns a channel, then I'm not sure how are we going to take the error from the channel and return it ? Because each action function is synchronous.

Somehow we would need to pass the status chan passed to the `(*Controller).Run` to this action method. And then keep sending the statuses to that channel. But that needs slight tinkering with the `run` function signature, which is `run       func() control.UserStatus`.

What do you think ?

loadtest/user/userentity/user.go
```diff
@@ -54,35 +58,56 @@ func New(store store.MutableUserStore, id int, config Config) *UserEntity {
...
+	cli.Close()
+	cli = nil
+	// Wait for the listener to shut down.
+	close(ue.closing)
+	<-ue.closed
+	return nil
+}
+
+// SendWebsocketMessage sends a given action type with data to the websocket.
+func (ue *UserEntity) SendWebsocketMessage(action string, data map[string]interface{}) error {
```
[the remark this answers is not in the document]
agnivade: That sounds fine with me. We can do that later. Will remove this. This was more to help me with testing.

loadtest/user/userentity/websocket.go
agnivade: Yes, I believe so. When the eventchannel and responsechannel are closed, the underlying connection is also closed. And `Listen` just starts to do `if _, rawMsg, err = wsc.Conn.ReadMessage(); err != nil {`. Which won't work if the connection is closed. We need to create a new one.

I have tested the code by shutting down the server, trying to reconnect and verified that it succesfully reconnects when the server comes up again.

streamer45: Mmm, interesting. Can't we just call `WebSocketEvent.Connect()` if we disconnect?

agnivade: Ah yes, that works too :+1: Sorry, hadn't seen that method. :stuck_out_tongue:

loadtest/user/userentity/websocket.go
streamer45: One option could be to create a goroutine that pipes those errors into the `UserStatus` channel. Something like:

```go
ec := c.user.Connect()
go func() {
  for err := range ec {
  }
}()
```

Making sure we `close(ec)` on disconnect.

agnivade: Sure, that's fine. Then we need to have the status channel as a field of the controller.

streamer45: Correct. It's definitely not the most clever solution but it's simple and should be good enough for our current needs.

loadtest/user/userentity/websocket.go
agnivade: As discussed, we will move to an error channel.

loadtest/user/userentity/user.go
agnivade: Actually, on the Connect method, we check for `cli != nil`, and now we are returning the error in a channel. So the code is different in different places.

loadtest/user/userentity/user.go
```diff
@@ -60,21 +61,24 @@ func New(store store.MutableUserStore, id int, config Config) *UserEntity {
 	ue.store = store
 	ue.closing = make(chan struct{})
 	ue.closed = make(chan struct{})
+	ue.wsErrorChan = make(chan error, 1)
 	return &ue
 }
 
 // Connect creates a websocket connection to the server and starts listening for messages.
-func (ue *UserEntity) Connect() error {
+func (ue *UserEntity) Connect() chan error {
```
streamer45: I'd like to be explicit with the direction of the channel here to make sure it's *readonly* for the controller.

loadtest/user/userentity/user.go
```diff
@@ -22,6 +22,7 @@ type UserEntity struct {
 	wsClient    *model.WebSocketClient
 	closing     chan struct{}
 	closed      chan struct{}
```
streamer45: Any opinion on adding the `ws` prefix also to these two? Might seem a bit vague otherwise.

loadtest/control/simplecontroller/actions.go
```diff
@@ -41,10 +41,12 @@ func (c *SimpleController) login() control.UserStatus {
...
 	}
 
-	err = c.user.Connect()
-	if err != nil {
-		return c.newErrorStatus(err)
-	}
+	errChan := c.user.Connect()
+	go func() {
+		for err := range errChan {
+			c.status <- c.newErrorStatus(err)
```
ethervoid: So we have to revamp the actions to send the error and info status to the channel, right? Instead of having two different ways to cope with status

loadtest/user/userentity/websocket.go
```diff
@@ -0,0 +1,84 @@
...
+	maxWebsocketFails             = 7
+)
+
+// listen starts to listen for messages on various channels.
+// It will keep reconnecting if the connection closes.
+// Only on calling Disconnect explicitly, it will return.
+func (ue *UserEntity) listen(errChan chan error) {
+	connectionFailCount := 0
+	for {
+		client, err := model.NewWebSocketClient4(ue.config.WebSocketURL, ue.client.AuthToken)
```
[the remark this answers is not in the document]
agnivade: Yes, sorry about that. We discussed offline on this. Using the `Connect` method to reconnect works, but it creates a nasty race condition when the user is executing client.SendMessage. This creates a race on the connection because at the same time, you are trying to replace the connection object and write to it.
Creating a new connection from scratch avoids this because it is an entirely new object. And the user will try to write to the old connection, which will fail, but there will be no races.

loadtest/control/simplecontroller/controller.go
```diff
@@ -13,9 +13,10 @@ import (
 )
 
 type SimpleController struct {
-	user user.User
-	stop chan struct{}
-	rate float64
+	user   user.User
+	stop   chan struct{}
+	status chan<- control.UserStatus
```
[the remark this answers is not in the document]
streamer45: The `SampleController` is just an example of implementation to showcase how a `UserController` could work. So as long as it compiles and runs without errors I think it's fine to leave it as it is.
There's isn't a real usecase for storing the status channel in the `SampleController` so I am not sure adding it makes any difference.

loadtest/user/userentity/user.go
```diff
@@ -54,35 +59,60 @@ func New(store store.MutableUserStore, id int, config Config) *UserEntity {
...
+	go ue.listen(ue.wsErrorChan)
+	return ue.wsErrorChan
 }
 
+// Disconnect closes the websocket connection.
 func (ue *UserEntity) Disconnect() error {
-	if ue.wsClient == nil {
+	cli := ue.getWsClient()
+	if cli == nil {
 		return errors.New("user is not connected")
```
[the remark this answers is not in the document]
agnivade: The `cli = nil` happens only on Disconnect. And at that time, those channels are already closed. And when `cli` is not nil, we are already closing them in the next lines. Perhaps I have misunderstood something ?

loadtest/control/simplecontroller/controller.go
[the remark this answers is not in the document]
agnivade: Oh yeah, good point. :+1:

loadtest/user/userentity/user.go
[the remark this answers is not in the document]
ethervoid: No, just asking in case I've missed something. I want to be sure we're closing all the channels properly.

loadtest/control/simplecontroller/actions.go
agnivade: Not sure whether this was something for me or just an observation. :sweat_smile: We are just changing the `Connect` method to return the channel. The actions like signUp, login, logout remains the same because now the controller has the status channel as a field.

I am not sure if you meant for me to change something.

ethervoid: Well is a question and the action to take is based on the answer :smile: . 

I mean, why do we need to return synchronous status in the controller actions if we are able to pass that status through the status channel and process it in the `handleStatus` function.

If we want to change the status logic inside the controller actions we have two options:
- Change all in this PR
- Create a new ticket to revamp it in other PR

streamer45: I'd opt for the second option. I think this would be out of scope for the current PR and also I would tend to wait on a similar change until we start using those statuses so we can figure out exactly what we need. But that's excellent feedback @ethervoid, thank you!

ethervoid:
LGTM! Great job @agnivade :)

loadtest/control/simplecontroller/actions.go
agnivade: Yes, it would be ideal if the status channel is passed all the way from top to the individual user methods. It is not good to spin up goroutines without having a way to gracefully shut them down. If we have 100 controllers logging in, there would be 100 goroutines, which is not really needed.

We need to get this fixed some time.

PR #74, opened 2019-12-20
MM-21367: Fix panic when Disconnect gets called consecutively

agnivade:
#### Summary

The issue was that when the controller loop runs for the 2nd time,
the user does not login again, and the API throws "user already signed up error".
This means user.Connect does not get called. Now when it goes forward,
it calls Disconnect again during logout, essentially calling disconnect 2 times
consecutively.

This creates a panic because we call `close(ue.wsClosing)` twice consecutively.

We fix 2 bugs here:

1. We reset the channels in Connect.
2. We add a state check to indicate whether the user is disconnected/connected.
And depending on that, bail out early from the disconnect function.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21367

### DIFF
--- a/loadtest/user/userentity/user.go
+++ b/loadtest/user/userentity/user.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"net"
 	"net/http"
-	"sync"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store"
@@ -18,7 +17,6 @@ type UserEntity struct {
 	id          int
 	store       store.MutableUserStore
 	client      *model.Client4
-	wsClientMut sync.RWMutex
 	wsClient    *model.WebSocketClient
 	wsClosing   chan struct{}
 	wsClosed    chan struct{}
@@ -72,8 +70,7 @@ func (ue *UserEntity) Connect() <-chan error {
 		ue.wsErrorChan <- errors.New("user is not authenticated")
 		return ue.wsErrorChan
 	}
-	cli := ue.getWsClient()
-	if cli != nil {
+	if ue.connected {
 		ue.wsErrorChan <- errors.New("user is already connected")
 		return ue.wsErrorChan
 	}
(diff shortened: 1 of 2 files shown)


streamer45:
Funny how this was actually caused by another bug cause the users should login at every loop.

The problem is that we store the user's password in the `model.User` object that we save into the store and when we call `user.GetMe()` in the `reload()` action we reset that field since the API will return it blank.  
At the next iteration of the loop we attempt to login with an empty password hence we never connect.   
So, 3 bugs at the price of one :tada:

loadtest/user/userentity/user.go
```diff
@@ -78,13 +79,14 @@ func (ue *UserEntity) Connect() <-chan error {
...
 	go ue.listen(ue.wsErrorChan)
+	ue.connected = true
 	return ue.wsErrorChan
 }
 
 // Disconnect closes the websocket connection.
 func (ue *UserEntity) Disconnect() error {
 	cli := ue.getWsClient()
-	if cli == nil {
+	if cli == nil || !ue.connected {
```
streamer45: So I am wondering, now that we have this can we remove the mutex and the shared client object? As far as I can see `wsClient` can stay internal to the listen goroutine and synchronization from/to the outside done only through available channels.

agnivade: Yep !

loadtest/user/userentity/user.go
```diff
@@ -18,7 +17,6 @@ type UserEntity struct {
 	id          int
 	store       store.MutableUserStore
 	client      *model.Client4
-	wsClientMut sync.RWMutex
 	wsClient    *model.WebSocketClient
```
streamer45: Can we also remove this?

agnivade: Yep, that too. I also realized there is one edge case that can be due to this localized state. Previously, we were closing the connection from the Disconnect method, but now we are doing it from the goroutine. But this means that when we close the `closing` channel, it might be that the loop is sleeping during reconnect. And since we are not waiting for the response to come from the `closed` channel, the connection can still remain indefinitely open if it keeps on getting failed to be connected.

It's not the end of the world. But just to be extra-cautious I doubled the timeout on the `closed` channel.

streamer45: I see. What if instead of sleeping we do a `time.After` inside a select that also checks for the closing of the channel?

agnivade: Heh .. that's sneaky :smile: . I think this is getting out of scope for this PR. Mind if I do it in a separate PR ?

streamer45: Sure, let's merge this :+1:

PR #75, opened 2019-12-21
[MM-21375] Fix blanking user's password in MemStore.SetUser

streamer45:
#### Summary

PR fixes an issue with user's password being set to empty which caused login errors.

#### Ticket

https://mattermost.atlassian.net/browse/MM-21375

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -60,11 +60,7 @@ func (s *MemStore) SetUser(user *model.User) error {
 	if user == nil {
 		return errors.New("user should not be nil")
 	}
-
-	if s.user != nil && user.Password == "" {
-		user.Password = s.user.Password
-	}
-
+	restorePrivateData(s.user, user)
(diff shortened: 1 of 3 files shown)


loadtest/store/memstore/store.go
```diff
@@ -60,6 +60,11 @@ func (s *MemStore) SetUser(user *model.User) error {
 	if user == nil {
 		return errors.New("user should not be nil")
 	}
+
+	if s.user != nil && user.Password == "" {
```
agnivade: How about ignoring other stuff too, like AuthData and MfaSecret ? https://github.com/mattermost/mattermost-server/blob/fdd1ae974fc1c27a85378f9ed535119ab87331ec/model/user.go#L495

streamer45: I don't believe we'll be making use of that data any time soon but I guess it wouldn't hurt to ignore it so we don't get into this problem again. I'll update.

PR #76, opened 2019-12-21
[MM-21287] Refactor UserController/User

streamer45:
#### Summary

- Removed `UserController.Init()`. The implementations should now provide their factory functions.
- Removed status parameter from `UserController.Run()`. This will be now passed during controller creation.
- Removed `Id` from `UserEntity` and added it to the controller since it is used to identify the latter.
- Removed unnecessary slice of users in `LoadTester`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-21287

### DIFF
--- a/example/samplecontroller/controller.go
+++ b/example/samplecontroller/controller.go
@@ -26,10 +26,10 @@ type userAction struct {
 
 func New(id int, user user.User, status chan<- control.UserStatus) *SampleController {
 	return &SampleController{
-		id,
-		user,
-		make(chan struct{}),
-		status,
+		id:     id,
+		user:   user,
+		stop:   make(chan struct{}),
+		status: status,
 	}
 }
 
--- a/example/samplecontroller/controller.go
+++ b/example/samplecontroller/controller.go
@@ -13,6 +13,7 @@ import (
 )
 
 type SampleController struct {
+	id   int
 	user user.User
 	stop chan struct{}
 }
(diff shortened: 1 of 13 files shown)


example/samplecontroller/controller.go
```diff
@@ -13,23 +13,29 @@ import (
...
 	run       func() control.UserStatus
 	waitAfter time.Duration
 }
 
-func (c *SampleController) Init(user user.User) {
-	c.user = user
-	c.stop = make(chan struct{})
+func New(id int, user user.User, status chan<- control.UserStatus) *SampleController {
+	return &SampleController{
+		id,
```
agnivade: Let's mention the field names too for clarity.

loadtest/control/simplecontroller/controller.go
```diff
@@ -13,26 +13,28 @@ import (
...
 	stop   chan struct{}
 	status chan<- control.UserStatus
 	rate   float64
 }
 
-func (c *SimpleController) Init(user user.User) {
-	c.user = user
-	c.stop = make(chan struct{})
-	c.rate = 1.0
+func New(id int, user user.User, status chan<- control.UserStatus) *SimpleController {
```
agnivade: Let's add some godocs whenever we touch an exported function, so that it doesn't become a burden later :)

streamer45: Also created a dedicated [ticket](https://mattermost.atlassian.net/browse/MM-21389) so we try to improve on that front.

PR #77, opened 2019-12-24
Fix indefinite wait in websocket reconnect

agnivade:
#### Summary

There may be a case where the server is down and the websocket
keeps on trying to reconnect. In that case, if the user tries
to disconnect, the closing signal will never reach the channel.

So we used to just wait with a timeout and return. This would
leave the goroutine dangling.

A better way is to select on 2 channels, and return whenever
we get the closing signal.


agnivade:
I did not feel it's worth to refactor those 2 lines to a separate function. Because we would be replacing 2 lines with 1 line. And adding 4 lines more.

PR #78, opened 2019-12-28
[MM-21292] Expose API to add/remove users at runtime

streamer45:
#### Summary

- Exposed a simple API (`AddUser`/`RemoveUser`) to increment and decrement the number of concurrently active users at runtime.
- Improved/refactored the existing API by providing `Run`/`Stop` methods to `LoadTester`.
- Added tests and some initial documentation for `LoadTester`.
- Decoupled the `loadtest` package from the lower level implementations (`UserEntity`, `SimpleController`). Now a `LoadTester` is able to run independently of the type of `UserController` used.

#### Ticket

https://mattermost.atlassian.net/browse/MM-21292


loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+// control.UserController given an id and a channel of control.UserStatus
+type NewController func(int, chan<- control.UserStatus) control.UserController
+
+func (lt *LoadTester) handleStatus() {
+	for us := range lt.status {
+		if us.Code == control.USER_STATUS_STOPPED || us.Code == control.USER_STATUS_FAILED {
+			lt.wg.Done()
+		}
+		if us.Code == control.USER_STATUS_ERROR {
+			mlog.Info(us.Err.Error(), mlog.Int("controller_id", us.ControllerId))
```
agnivade: Should this be `mlog.Error` ?

loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+	started       bool
+	newController NewController
+}
+
+// NewController is a factory function that returns a new
+// control.UserController given an id and a channel of control.UserStatus
+type NewController func(int, chan<- control.UserStatus) control.UserController
+
+func (lt *LoadTester) handleStatus() {
+	for us := range lt.status {
```
agnivade: Perhaps renaming it to `st` instead of `us` makes it more readable ?

loadtest/loadtest_test.go
```diff
@@ -0,0 +1,120 @@
...
+	require.EqualError(t, err, "LoadTester is not running")
+
+	err = lt.Run()
+	require.NoError(t, err)
+	require.True(t, lt.started)
+
+	err = lt.Stop()
+	require.NoError(t, err)
+	require.False(t, lt.started)
+	require.Empty(t, lt.controllers)
```
agnivade: Let's also add some users and then stop it, and then verify this.

loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+		} else if us.Code == control.USER_STATUS_FAILED {
+			mlog.Error(us.Err.Error())
+			continue
+		}
+		mlog.Info(us.Info, mlog.Int("controller_id", us.ControllerId))
+	}
+}
+
+// AddUser increments by one the number of concurrently active users
+func (lt *LoadTester) AddUser() error {
```
agnivade: Should we also add wrapper functions which add/remove n users ?

loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+	if !lt.started {
+		return errors.New("LoadTester is not running")
```
agnivade: Let's use named errors for these. They can be reused in RemoveUser and Stop methods and also tested better in tests by avoiding a string match. Also, idiomatically speaking we should not capitalize error strings: https://github.com/golang/go/wiki/CodeReviewComments#error-strings. But I am fine with either ways, as long as we are consistent, which we are here.

loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+	activeUsers := len(lt.controllers)
+	if activeUsers == lt.config.UsersConfiguration.MaxActiveUsers {
+		return errors.New("Max active users limit reached")
+	}
+	controller := lt.newController(activeUsers+1, lt.status)
+	lt.wg.Add(1)
+	go func() {
+		controller.Run()
+	}()
+	lt.controllers = append(lt.controllers, controller)
```
agnivade: There is a possible race condition here during AddUser/RemoveUser if they get called from different goroutines. It may be worthwhile to protect the access to `controllers` slice with a mutex.

loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+// RemoveUser decrements by one the number of concurrently active users
+func (lt *LoadTester) RemoveUser() error {
+	if !lt.started {
+		return errors.New("LoadTester is not running")
+	}
+	activeUsers := len(lt.controllers)
+	if activeUsers == 0 {
+		return errors.New("No active users left")
+	}
+	controller := lt.controllers[activeUsers-1]
```
agnivade: I think to better mimic realistic conditions, it may be better to randomly pop an user from anywhere in the slice. The current method just pops them from the end, which simulates users logging in, and only new users being removed, whereas old users are still there. In a real-world situation, any user might log off.

loadtest/loadtest.go
```diff
@@ -0,0 +1,120 @@
...
+)
+
+// LoadTester is a structure holding all the state needed to run a load-test
+type LoadTester struct {
+	controllers   []control.UserController
+	config        *config.LoadTestConfig
+	wg            sync.WaitGroup
+	status        chan control.UserStatus
+	started       bool
+	newController NewController
```
agnivade: Is there a reason we are using this factory function here ? Is it just to de-duplicate the initialization code ?

cmd/loadtest/main.go
```diff
@@ -6,35 +6,71 @@ package main
...
+	config, err := config.GetConfig()
+	if err != nil {
+		return err
+	}
+
+	newSimpleController := func(id int, status chan<- control.UserStatus) control.UserController {
+		ueConfig := userentity.Config{
+			ServerURL:    config.ConnectionConfiguration.ServerURL,
+			WebSocketURL: config.ConnectionConfiguration.WebSocketURL,
+		}
```
reflog: I wonder, since loadtest is ran against one instance at a time - what's the point of storing ServerURL and WebsocketURL in each userEntity? Why not keep in some global/context?

cmd/loadtest/main.go
```diff
@@ -6,35 +6,71 @@ package main
...
+	err = lt.Stop()
+	mlog.Info("loadtest done", mlog.String("elapsed", time.Since(start).String()))
+
+	return err
 }
 
 func RunExampleCmdF(cmd *cobra.Command, args []string) error {
-	return example.Run()
+	lt := example.New("http://localhost:8065")
+	return lt.Run(4)
```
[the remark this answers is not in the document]
streamer45: I guess it was expected for those two to be diverging at some point. It's just an example after all.  
That said, the main thing to show there would be the `UserController` since at this time it's the only piece that realistically allows multiple implementations.  
Maybe we can cleanup some of the example and just keep the controller part but that should be a separate PR.

loadtest/loadtest.go
streamer45: I think originally the idea was to treat errors related to users as normal information since they are expected to happen but it's fine either way as long as we can track everything from the logs.

loadtest/loadtest.go
streamer45: I've intentionally kept it as simple as possible in order to let the user of the API be in control of the timings related to adding/removing users.  I believe it should be the higher level to decide how to add/remove users (e.g. all at once vs n per second or some other random way).

loadtest/loadtest.go
streamer45: By *named errors* you mean defining constants like `ErrSomething = errors.New("some error")`?. If so we can definitely do that.  
Also, I normally use lowercase strings for errors but in this case it's the name of the type that's capitalized. Maybe we can rephrase if we really want to follow the guidelines.

loadtest/loadtest.go
streamer45: Yes, this is another case where I've intentionally left some decisions for later.  
Access to `LoadTester.started` should also be atomic/guarded if we ever wanted to use it from different goroutines.  
I'd like to review this when we have a better picture of how we are going to use this API to avoid overengineering it now and wait until we have a concrete reason to do so.

loadtest/loadtest.go
streamer45: You are right that we want to mimic what happens with real users but at this time my main concern is to not add any extra complexity and especially try to avoid nondeterminism as much as possible. This is the same approach taken with `SimpleController`. It's not a very realistic representation of how a user behaves but it's important to have it nonetheless in order to keep the number of variables down to a minimum and allow us to have more stable results.
I prefer to go back to this kind of details later when we have the whole load-test system up and running and we begin the *tuning* process.
It might be interesting to make the way we remove a user something decidable by the user of this API, for example by passing a typed constant (e.g. random, first, last).

loadtest/loadtest.go
streamer45: The idea around the factory function is that we don't want the `LoadTester` to know anything about how the users/controllers are implemented so that the same code will work for any implementation of those two interfaces.  
At the same time we want to be able to create the concrete values from the inside without requiring the caller to provide them to us, so as to remove some complexity from the user's side. I find passing a factory function to be an elegant solution in this case.  
The reason the function is stored inside `LoadTester` is because we pass it once during initialization and then we use it at later time in `AddUser`.

cmd/loadtest/main.go
streamer45: In a sense we already have this global object which is the config.  
The idea of passing `ServerURL`  and  `WebSocketURL` there was to keep the lower level decoupled from the config and every user's state completely independent from each others. If it ever becomes a problem of memory we have several possible solutions but for now I am not too worried about this kind of redundancy.

loadtest/loadtest.go
agnivade: > By named errors you mean defining constants like ErrSomething = errors.New("some error")?. If so we can definitely do that.

Yep, I meant that.

> Also, I normally use lowercase strings for errors but in this case it's the name of the type that's capitalized

Ah ok, no worries :+1:

loadtest/loadtest.go
agnivade: That's fine. I would suggest to just leave a comment noting down this thought, so that it is apparent for someone reading the code.

loadtest/loadtest.go
agnivade: Sounds good. Let's document it somewhere (ticket/quick inline comment) so that we don't forget about it later.

loadtest/loadtest.go
agnivade: Ah okay, so you are saying there will be different user/controller implementations, and it will be cumbersome to change an internal `LoadTester` method every time we want to switch to a different implementation ?

I think this is a great point and should be noted in the godoc for `NewContoller`.

streamer45: Yes, there will definitely be at least two `UserController` implementations, the *simple* one we are building now and a more *simulative* one yet to be coded.  
I'll improve the documentation including some of these thoughts so to make clear the reasoning behind these design choices.

PR #79, opened 2019-12-31
MM-21416: complete SimpleController.reload implementation

agnivade:
#### Summary

- Invoke the commented methods which are now added
- Reorder it to happen after login
- Add Disconnect/Connect methods to it to better simulate a browser

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21416

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -41,12 +41,7 @@ func (c *SimpleController) login() control.UserStatus {
 		return c.newErrorStatus(err)
 	}
 
-	errChan := c.user.Connect()
-	go func() {
-		for err := range errChan {
-			c.status <- c.newErrorStatus(err)
-		}
-	}()
+	c.connect()
 
 	return c.newInfoStatus("logged in")
 }
@@ -111,21 +106,21 @@ func (c *SimpleController) viewChannel() control.UserStatus {
 	*/
 }
 
-func (c *SimpleController) reload() control.UserStatus {
-	err := c.user.Disconnect()
-	if err != nil {
-		return c.newErrorStatus(err)
-	}
-
-	errChan := c.user.Connect()
-	go func() {
-		for err := range errChan {
+// reload performs all actions done when a user reloads the browser.
+// If full parameter is enabled, it also disconnects and reconnects
+// the WebSocket connection.
+func (c *SimpleController) reload(full bool) control.UserStatus {
+	if full {
+		err := c.user.Disconnect()
+		if err != nil {
 			c.status <- c.newErrorStatus(err)
 		}
-	}()
+
+		c.connect()
+	}
 
 	// Getting preferences.
-	err = c.user.GetPreferences()
+	err := c.user.GetPreferences()
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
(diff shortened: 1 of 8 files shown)


loadtest/control/simplecontroller/actions.go
```diff
@@ -141,22 +156,56 @@ func (c *SimpleController) reload() control.UserStatus {
...
+	userId, err := c.user.GetMe()
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
 
-	// TODO: GetTeamsForUser
-	// TODO: GetTeamMembersForUser
-	// TODO: GetRolesByNames
-	// TODO: GetWebappPlugins
-	// TODO: GetAllTeams
```
streamer45: We are missing this call which is separate from `User.GetTeams()`.

loadtest/control/simplecontroller/controller.go
```diff
@@ -49,19 +49,19 @@ func (c *SimpleController) Run() {
 			waitAfter: 1000,
 		},
 		{
-			run:       c.createPost,
+			run:       c.reload,
```
streamer45: Since we already handle connecting and disconnecting in login/logout I think we can add a parameter to `reload()` specifying whether we want a complete reload or a first load with the only difference that in the former case we would also disconnect and connect.  
Then we can probably wrap it into an anonymous function here so we keep the action *parameterless*.

loadtest/control/simplecontroller/actions.go
```diff
@@ -141,22 +156,56 @@ func (c *SimpleController) reload() control.UserStatus {
 		}
 	}
 
-	// TODO: GetConfig
+	err = c.user.GetConfig()
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
 	// TODO: GetLicense
```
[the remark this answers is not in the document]
agnivade: This `GetOldClientLicense` is there. Perhaps we just need to add the route for new license ?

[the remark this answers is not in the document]
streamer45: Nice find. Looking at the code it seems it's just the returned format to be *old*. Everything else is the same. I guess we can add this to `UserEntity` and call it here.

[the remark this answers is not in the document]
agnivade: Done.

loadtest/control/simplecontroller/controller.go
```diff
@@ -49,19 +50,26 @@ func (c *SimpleController) Run() {
...
 		},
 		{
-			run:       c.createPost,
+			run: func() control.UserStatus {
+				// Do not perform a full reload
+				// on a first run.
+				if i == 0 {
+					return c.reload(false)
+				}
+				return c.reload(true)
```
streamer45: I don't think we need this check. After a login, `reload` should always be called with `false` since we don't need to disconnect/reconnect (we just did it). Each iteration of the actions loop should be the same in `SimpleController` and it should terminate with the user logging out. The only exception would be signing up which is done once. The full reload should be done later on in the loop (somewhere in the middle for this simple controller case but randomly in the future) to simulate a page refresh.

agnivade: Ah sorry, I misunderstood. Yes that makes sense.

PR #80, opened 2019-12-31
MM-20704: implement basic random utilities in UserStore

agnivade:
#### Summary

- RandomChannel
- RandomTeam
- RandomUser

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-20704

### DIFF
--- a/loadtest/store/memstore/random.go
+++ b/loadtest/store/memstore/random.go
@@ -4,10 +4,16 @@
 package memstore
 
 import (
+	"errors"
 	"math/rand"
 	"reflect"
 
 	"github.com/mattermost/mattermost-server/v5/model"
+)
+
+var (
+	ErrEmptyMap   = errors.New("cannot select from an empty map")
+	ErrEmptySlice = errors.New("cannot select from an empty slice")
 )
 
 // RandomChannel returns a random channel for a user.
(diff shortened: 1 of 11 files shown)


loadtest/store/memstore/store.go
```diff
@@ -315,3 +316,42 @@ func (s *MemStore) Roles() ([]model.Role, error) {
 	}
 	return roles, nil
 }
+
+// For all the random utilities we flatten the map to a slice
```
streamer45: I would put everything related to random picking data in a separate file since it's very likely this will grow further.
Also I would implement functions to randomly pick:

- A post
- An emoji
- A channel member
- A team member

This way we cover pretty much everything we currently store.

loadtest/store/memstore/store.go
```diff
@@ -315,3 +316,42 @@ func (s *MemStore) Roles() ([]model.Role, error) {
...
+// and then take a random index from the slice.
+// We do this, because even though the map iteration order
+// is "undefined" in the spec, practically it returns the
+// same set of values always.
+
+// RandomChannel returns a random channel for a user.
+func (s *MemStore) RandomChannel() (model.Channel, error) {
```
streamer45: I suggest we either add a `teamId` parameter or we add a function to pick a channel randomly but with the specified `teamId`.

agnivade: Yes, I was thinking about that too. Will update.

loadtest/store/memstore/store.go
agnivade: Ah okay, then it might be better to use a util function with `interface{}`.

streamer45: Yes, that's probably a good option and as long as we keep it internal to the package we shouldn't have problems of being too generic.

loadtest/store/memstore/random.go
```diff
@@ -59,45 +48,32 @@ func (s *MemStore) RandomEmoji() (model.Emoji, error) {
...
-	items := make([]interface{}, len(m))
-	i := 0
-	for _, val := range m {
-		items[i] = val
-		i++
-	}
-	return items[rand.Intn(len(items))]
+func pickRandomKeyFromMap(m interface{}) interface{} {
+	keys := reflect.ValueOf(m).MapKeys()
+	idx := rand.Intn(len(keys))
```
streamer45: `rand.Intn` will panic if given 0 as input. So I propose we either return an error or we make 100% sure we always call this with a non-empty map. Also a test case for this would be nice.

agnivade: Good catch.

loadtest/store/memstore/random.go
```diff
@@ -68,12 +92,22 @@ func (s *MemStore) RandomTeamMember(teamId string) (model.TeamMember, error) {
...
+	}
+	return *teamMemberMap[key.(string)], nil
 }
 
-func pickRandomKeyFromMap(m interface{}) interface{} {
-	keys := reflect.ValueOf(m).MapKeys()
+func pickRandomKeyFromMap(m interface{}) (interface{}, error) {
+	val := reflect.ValueOf(m)
+	if val.Kind() != reflect.Map {
+		return nil, errors.New("not a map")
```
streamer45: Is there a reason for not using a named error here?

agnivade: It's an internal error. There is no way to pass a parameter which will cause `pickRandomKeyFromMap` to get something else other than a map. The error can indeed bubble up, but it's unlikely. So there isn't much use to naming it.

PR #81, opened 2020-01-02
MM-21415: make LoadTester API goroutine-safe

agnivade:
#### Summary

Add appropriate mutexes necessary to guard internal state
for racy access.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21415

### DIFF
--- a/loadtest/loadtest.go
+++ b/loadtest/loadtest.go
@@ -14,12 +14,16 @@ import (
 
 // LoadTester is a structure holding all the state needed to run a load-test
 type LoadTester struct {
-	controllers   []control.UserController
-	config        *config.LoadTestConfig
-	wg            sync.WaitGroup
-	status        chan control.UserStatus
-	started       bool
-	newController NewController
+	controllersMut  sync.RWMutex
+	controllers     []control.UserController
+	config          *config.LoadTestConfig
+	wg              sync.WaitGroup
+	status          chan control.UserStatus
+	startedMut      sync.RWMutex
+	started         bool
+	statusClosedMut sync.RWMutex
+	statusClosed    bool
+	newController   NewController
 }
 
 // NewController is a factory function that returns a new
@@ -30,6 +34,9 @@ type LoadTester struct {
 type NewController func(int, chan<- control.UserStatus) control.UserController
 
 func (lt *LoadTester) handleStatus() {
+	lt.withStatusLock(func() {
+		lt.statusClosed = false
+	})
 	for st := range lt.status {
 		if st.Code == control.USER_STATUS_STOPPED || st.Code == control.USER_STATUS_FAILED {
 			lt.wg.Done()
@@ -47,9 +54,11 @@ func (lt *LoadTester) handleStatus() {
 
 // AddUser increments by one the number of concurrently active users
 func (lt *LoadTester) AddUser() error {
-	if !lt.started {
+	if !lt.hasStarted() {
 		return ErrNotRunning
 	}
+	lt.controllersMut.Lock()
+	defer lt.controllersMut.Unlock()
 	activeUsers := len(lt.controllers)
 	if activeUsers == lt.config.UsersConfiguration.MaxActiveUsers {
 		return ErrMaxUsersReached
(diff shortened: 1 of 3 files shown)


loadtest/loadtest.go
```diff
@@ -47,9 +54,11 @@ func (lt *LoadTester) handleStatus() {
 
 // AddUser increments by one the number of concurrently active users
 func (lt *LoadTester) AddUser() error {
-	if !lt.started {
+	if !lt.hasStarted() {
```
streamer45: I would have probably opted for an atomic variable in this case since it's a simple true/false state value. We would have to use an `int32` instead of a `bool` but at least we wouldn't need the additional `mutex`. What do you think?

loadtest/loadtest.go
```diff
@@ -99,17 +114,30 @@ func (lt *LoadTester) Run() error {
...
+	for range controllers {
 		if err := lt.RemoveUser(); err != nil {
 			mlog.Error(err.Error())
 		}
 	}
 	lt.wg.Wait()
-	close(lt.status)
+	lt.startedMut.Lock()
 	lt.started = false
+	lt.startedMut.Unlock()
```
streamer45: This way we wouldn't guard against two goroutines attempting to remove users at the same time. While it wouldn't cause a race condition cause of the internal locks in `RemoveUser` I don't think it should be allowed. 
Maybe we should add a `stopping` variable/channel to make sure everything in `Stop` is run just by one goroutine and also avoid any other API call during this teardown phase. Thoughts?

loadtest/loadtest.go
agnivade: That's better in fact.

loadtest/loadtest.go
agnivade: Do you mean since we are reading a copy of `controllers` and then removing users for the length of that slice, there may be a possibility that 2 goroutines get the same length and one of them return a `ErrNoUsersLeft` ?

You want to combine the controllersMut and startedMut in a single larger mutex ? But wouldn't that still cause racy access for lt.controllers now that it won't be controlled by the same mutex ?

streamer45: > Do you mean since we are reading a copy of `controllers` and then removing users for the length of that slice, there may be a possibility that 2 goroutines get the same length and one of them return a `ErrNoUsersLeft` ?

Yes, only one goroutine should be able to stop a running load-test. This is more about keeping integrity and make the API behave as expected.

> You want to combine the controllersMut and startedMut in a single larger mutex ? But wouldn't that still cause racy access for lt.controllers now that it won't be controlled by the same mutex ?

Not exactly. What I think we need here is an additional state variable (atomic) or channel (whatever comes the simplest to implement).
We call it `stopping` and use it to make sure that only one goroutine is allowed to run `Stop()` and also while that call is running no other goroutine is allowed to make any API call that touches the internal state (Start/Stop/AddUser/RemoveUser will have to check for both `started ` and this new variable to go through).
This would also remove the need for the `statusClosedMut` since the `close` would be guaranteed to happen only once.

agnivade:
Still some chinks remain in the armour, but it does most of the job.

agnivade:
That's okay, but then 2 concurrent calls to Stop won't immediately cause an error, but one would wait for the other to finish, and then cause an error. Is that okay ?

streamer45:
> That's okay, but then 2 concurrent calls to Stop won't immediately cause an error, but one would wait for the other to finish, and then cause an error. Is that okay ?

Maybe we can still leverage an atomic compare and swap operation just for the stop case right before we lock. Otherwise I am okay with waiting to fail. We can always revisit this at a later time.

agnivade:
> Maybe we can still leverage an atomic compare and swap operation just for the stop case right before we lock.

I think that will make things slightly inconsistent, because now Stop operations will immediately throw error, whereas `Run`s will wait and then throw an error. I think it's better to let this one wait for some time until the API becomes a bit more mature. Then we can revisit this.

agnivade:
I think it looks much better now !

loadtest/loadtest.go
```diff
@@ -47,7 +59,20 @@ func (lt *LoadTester) handleStatus() {
...
+	defer lt.mut.Unlock()
+
+	return lt.addUser(false)
+}
+
+// addUser is an internal API called from Run and AddUser both.
+// DO NOT call this by itself, because this method is not protected by a mutex.
+// isStarting is true when called from Start, and false when called from
+// AddUser.
+func (lt *LoadTester) addUser(isStarting bool) error {
```
streamer45: Why do we need to pass this `bool` value? Can't we just check the state?

agnivade: Yes, but this is more precise. Just checking for state would broaden the allowed states to 2. But with this, it can only be in one state.

PR #83, opened 2020-01-03
[MM-20702] Implement basic init command, add joinTeam action

streamer45:
#### Summary

PR does add a `init` method used to insert minimal data in a MM instance needed for a load-test to run.  
Right now it simply creates a config specified number of open teams which should be enough to get us started.

PR also adds a `joinTeam` action to the `SimpleController` and a utility method `IsSysAdmin` to `UserEntity` which is needed to allow for different behaviours based on what role the user has.

#### Ticket

https://mattermost.atlassian.net/browse/MM-20702

### DIFF
--- a/loadtest/store/store.go
+++ b/loadtest/store/store.go
@@ -20,6 +20,8 @@ type UserStore interface {
 	Channels(teamId string) ([]model.Channel, error)
 	// Teams return the teams a user belong to.
 	Teams() ([]model.Team, error)
+	// TeamMember returns the TeamMember for the given teamId and userId
+	TeamMember(teamdId, userId string) (model.TeamMember, error)
 	// Preferences return the preferences of the user.
 	Preferences() (model.Preferences, error)
 	// Roles return the roles of the user.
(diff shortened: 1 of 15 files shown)


loadtest/store/store.go
```diff
@@ -20,6 +20,8 @@ type UserStore interface {
 	Channels(teamId string) ([]model.Channel, error)
 	// Teams return the teams a user belong to.
 	Teams() ([]model.Team, error)
+	// TeamMember returns the TeamMember for the given teamId and userId
```
agnivade: Period at the end to be consistent with others. `Config` should also have it.

loadtest/user/user.go
```diff
@@ -104,4 +104,8 @@ type User interface {
 	// license
 	// GetClientLicense returns the client license in the old format.
 	GetClientLicense() error
+
+	// utils
+	// IsSysAdmin will return true if the user is a SystemAdmin, false otherwise
```
agnivade: Period at the end.

PR #85, opened 2020-01-08
added http agent

reflog:


cmd/loadtest/server.go
```diff
@@ -0,0 +1,162 @@
...
+		return nil, err
+	}
+	return lt, nil
+}
+func (a *Agent) runLoadTestHandler(w http.ResponseWriter, r *http.Request) {
+	lt, err := a.getLoadTestById(w, r)
+	if err != nil {
+		return
+	}
+	go lt.Run()
```
streamer45: Run is not a blocking call. It will return rather quickly. No need to create an additional goroutine.

cmd/loadtest/server.go
```diff
@@ -0,0 +1,162 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(map[string]string{"message": "load test started"})
+}
+
+func (a *Agent) stopLoadTestHandler(w http.ResponseWriter, r *http.Request) {
+	lt, err := a.getLoadTestById(w, r)
+	if err != nil {
+		return
+	}
+	_ = lt.Stop()
```
streamer45: Need to check the returned error.

cmd/loadtest/server.go
```diff
@@ -0,0 +1,162 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(map[string]loadtest.State{"status": lt.GetState()})
+}
+
+func (a *Agent) addUserHandler(w http.ResponseWriter, r *http.Request) {
+	lt, err := a.getLoadTestById(w, r)
+	if err != nil {
+		return
+	}
+	lt.AddUser()
```
streamer45: We need to support adding `n` users here through a parameter.

cmd/loadtest/server.go
```diff
@@ -0,0 +1,162 @@
...
+	_ = json.NewEncoder(w).Encode(map[string]string{"message": "user added"})
+
+}
+
+func (a *Agent) removeUserHandler(w http.ResponseWriter, r *http.Request) {
+	lt, err := a.getLoadTestById(w, r)
+	if err != nil {
+		return
+	}
+	lt.RemoveUser()
```
streamer45: Same as above. Need to support removing `n` users where `n` is a value passed in the request.

cmd/loadtest/server.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/simplecontroller"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/user/userentity"
+	"github.com/mattermost/mattermost-server/mlog"
+	"github.com/spf13/cobra"
+)
+
+type Agent struct {
+	loadTests map[string]*loadtest.LoadTester
+}
```
streamer45: To me the `Agent` is really the `LoadTester` which might actually need renaming.
I would call this type simply `API` and keep the router here (similar to what happens in the server repository). 
Also I think all this code (except the cmd part) should live in a dedicated `api` or `server` package in the project's root (alongside loadtest and cmd dirs).

loadtest/loadtest.go
```diff
@@ -130,6 +132,10 @@ func (lt *LoadTester) Run() error {
 	return nil
 }
 
+func (lt *LoadTester) GetState() State {
+	return lt.state
+}
```
streamer45: This is not needed. We'll have a `Status` function ready soon.

loadtest/loadtest.go
```diff
@@ -158,9 +164,11 @@ func New(config *config.LoadTestConfig, nc NewController) *LoadTester {
 	if config == nil || nc == nil {
 		return nil
 	}
+	id, _ := uuid.NewV1()
```
streamer45: I would probably keep the Id external. Only the API layer needs this.

loadtest/loadtest.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/mattermost/mattermost-load-test-ng/config"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
+	uuid "github.com/satori/go.uuid"
```
agnivade: Let's make this part of the mlog section. This is not a package in the project.

loadtest/loadtest.go
agnivade: satori/uuid is unmaintained. We should use https://github.com/gofrs/uuid now. Also, let's use V4 instead of V1.

cmd/loadtest/server.go
```diff
@@ -0,0 +1,162 @@
...
+	return lt, nil
+}
+func (a *Agent) runLoadTestHandler(w http.ResponseWriter, r *http.Request) {
+	lt, err := a.getLoadTestById(w, r)
+	if err != nil {
+		return
+	}
+	go lt.Run()
+	w.Header().Set("Content-Type", "application/json")
+	_ = json.NewEncoder(w).Encode(map[string]string{"message": "load test started"})
```
agnivade: Do we expect to pass more data in the response that we are using json ? As a first cut, can we just do - `fmt.Fprintf(w, "load test started")` ? Or is this more about being more consistent with the status handlers ? In which case, we can just do the same fmt.Fprintf() with the json string, instead of taking the json.NewEncoder route.

streamer45: I think it's fine to use JSON but I agree that for the time being it is more than enough to return a prepared JSON string without the need to encode anything.

reflog: I'm going to disagree here. What's the benefit of writing this with printf instead of properly encoding a structure? Almost no code-size difference, this is not a hot path and you loose ability to properly compose output.

streamer45: I guess for me it was more about readability and avoid adding unnecessary complexity but it's not a big deal. This is not the kind of detail we want to spend our time on right now so I am also super okay with keeping it as it is :+1:

loadtest/loadtest.go
```diff
@@ -130,6 +130,10 @@ func (lt *LoadTester) Run() error {
 	return nil
 }
 
+func (lt *LoadTester) GetState() State {
+	return lt.state
+}
```
streamer45: I think you forgot to remove this.

api/server.go
```diff
@@ -0,0 +1,173 @@
...
+func (a *API) addUserHandler(w http.ResponseWriter, r *http.Request) {
+	amount := getAmount(r)
+	lt, err := a.getLoadTestById(w, r)
+	if err != nil {
+		return
+	}
+
+	for i := 0; i < amount; i++ {
+		if err = lt.AddUser(); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
```
streamer45: I am bit concerned about the case where we successfully add some users and then we get an error (reaching the max users limit for example). We return the error but we actually added some users. I am wondering if it makes more sense for most of this API to always return the `LoadTester` status and deal with the error at the JSON level by including an `err` object if something went wrong. Thoughts?

api/server.go
```diff
@@ -0,0 +1,173 @@
...
+	writeJsonResponse(w, map[string]string{"message": "user removed"})
+}
+
+func SetupAPIRouter() *mux.Router {
+	router := mux.NewRouter()
+	r := router.PathPrefix("/loadtest").Subrouter()
+
+	agent := API{loadTests: make(map[string]*loadtest.LoadTester)}
+	r.HandleFunc("/create", agent.createLoadTestHandler).Methods("POST")
+	r.HandleFunc("/run/{id}", agent.runLoadTestHandler).Methods("POST")
```
streamer45: Wondering if it should be the other way around, e.g. `/{id}/actionName`

api/server.go
```diff
@@ -0,0 +1,173 @@
...
+	r.HandleFunc("/stop/{id}", agent.stopLoadTestHandler).Methods("POST")
+	r.HandleFunc("/destroy/{id}", agent.destroyLoadTestHandler).Methods("POST")
+	r.HandleFunc("/status/{id}", agent.getLoadTestStatusHandler).Methods("GET")
+	r.HandleFunc("/user/{id}", agent.addUserHandler).Methods("PUT").Queries("amount", "{[0-9]*?}")
+	r.HandleFunc("/user/{id}", agent.removeUserHandler).Methods("DELETE").Queries("amount", "{[0-9]*?}")
```
streamer45: I think `destroy` is the only call where it makes sense to use `DELETE` since we are actually deleting an object.
For adding/removing users I believe `POST` is more precise. Thoughts?

api/server.go
```diff
@@ -0,0 +1,185 @@
...
+			break // stop on first error, result is reported as part of status
+		}
+	}
+
+	writeJsonResponse(w, map[string]interface{}{"message": fmt.Sprintf("%d users removed", i+1), "status": MockStatus{}})
+}
+
+func SetupAPIRouter() *mux.Router {
+	router := mux.NewRouter()
+	r := router.PathPrefix("/loadtest").Subrouter()
```
streamer45: Should probably be `/loadagent` for the aforementioned reason.
Also I'd try to fix all the occurrences where we mixed up (mainly messages and API handlers).

api/server.go
```diff
@@ -0,0 +1,189 @@
...
+			WebSocketURL: config.ConnectionConfiguration.WebSocketURL,
+		}
+		ue := userentity.New(memstore.New(), ueConfig)
+		return simplecontroller.New(id, ue, status)
+	}
+
+	lt := loadtest.New(&config, newSimpleController)
+	a.agents[u.String()] = lt
+
+	writeJsonResponse(w, map[string]string{"loadTestId": u.String()})
```
streamer45: I would try to include the `LoadTester` status in the response every time we change it with the exception of the destroy call since that's final. This way the user of this API doesn't need to make an extra call.

go.mod
```diff
@@ -3,15 +3,31 @@ module github.com/mattermost/mattermost-load-test-ng
...
+	github.com/fatih/structs v1.1.0 // indirect
+	github.com/gavv/httpexpect v2.0.0+incompatible
+	github.com/go-ldap/ldap v3.0.3+incompatible // indirect
+	github.com/go-redis/redis v6.15.6+incompatible // indirect
+	github.com/gofrs/uuid v3.2.0+incompatible
+	github.com/gorilla/mux v1.7.3
+	github.com/imkira/go-interpol v1.1.0 // indirect
+	github.com/k0kubun/colorstring v0.0.0-20150214042306-9440f1994b88 // indirect
+	github.com/mattermost/mattermost-load-test v0.0.0-20191210214319-593c7cf2f402
+	github.com/mattermost/mattermost-server v5.11.1+incompatible
```
streamer45: Wondering what depends on these.

reflog: good catch!

cmd/loadtest/server.go
agnivade: The benefit would be performance. But at this stage, shaving off a few microseconds doesn't matter :smile: So no complaints from my side.

streamer45:
@reflog since we merged #87 implementing `LoadTester.Status()` it would be great if you could update this PR to use the real thing. Thanks :+1:

reflog:
@streamer45 done

PR #86, opened 2020-01-08
[MM-21557] Implement actions to add/remove reaction to/from a post

streamer45:
#### Summary

PR implements actions to add and remove reactions.
It also implemented some missing/needed API and store methods.

#### Ticket

https://mattermost.atlassian.net/browse/MM-21557

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -100,6 +100,55 @@ func (c *SimpleController) createPost() control.UserStatus {
 	return c.newInfoStatus(fmt.Sprintf("post created, id %v", postId))
 }
 
+func (c *SimpleController) addReaction() control.UserStatus {
+	// get posts from UserStore that have been created in the last minute
+	posts, err := c.user.Store().PostsSince(time.Now().Unix()*1000 + 60000)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	if len(posts) == 0 {
+		return c.newInfoStatus("no posts to add reaction to")
+	}
+
+	err = c.user.SaveReaction(&model.Reaction{
+		UserId:    c.user.Store().Id(),
+		PostId:    posts[0].Id,
+		EmojiName: "grinning",
+	})
+
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+
+	return c.newInfoStatus("added reaction")
+}
+
+func (c *SimpleController) removeReaction() control.UserStatus {
+	// get posts from UserStore that have been created in the last minute
+	posts, err := c.user.Store().PostsSince(time.Now().Unix()*1000 + 60000)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	if len(posts) == 0 {
+		return c.newInfoStatus("no posts to remove reaction from")
+	}
+
+	reactions, err := c.user.Store().Reactions(posts[0].Id)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	if len(reactions) == 0 {
+		return c.newInfoStatus("no reactions to remove")
+	}
+
+	err = c.user.DeleteReaction(&reactions[0])
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+
+	return c.newInfoStatus("removed reaction")
+}
+
 func (c *SimpleController) createGroupChannel() control.UserStatus {
 	channelId, err := c.user.CreateGroupChannel([]string{}) // TODO: populate memberIds parameter with other users
 	if err != nil {
--- a/loadtest/control/simplecontroller/controller.go
+++ b/loadtest/control/simplecontroller/controller.go
@@ -64,7 +64,11 @@ func (c *SimpleController) Run() {
 			waitAfter: 1000,
 		},
 		{
-			run:       c.createPost,
+			run:       c.addReaction,
+			waitAfter: 1000,
+		},
+		{
+			run:       c.removeReaction,
 			waitAfter: 1000,
 		},
 		{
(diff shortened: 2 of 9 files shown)


loadtest/control/simplecontroller/actions.go
```diff
@@ -100,6 +100,55 @@ func (c *SimpleController) createPost() control.UserStatus {
...
+	// get posts from UserStore that have been created in the last minute
+	posts, err := c.user.Store().PostsSince(time.Now().Unix()*1000 + 60000)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	if len(posts) == 0 {
+		return c.newInfoStatus("no posts to remove reaction from")
+	}
+
+	reactions, err := c.user.Store().Reactions(posts[0].Id)
```
ethervoid: Maybe we can iterate through it to check for reactions because we can have 10 posts in that minute and the 5th one is the one with reactions. What do you think?

loadtest/control/simplecontroller/controller.go
```diff
@@ -64,7 +64,11 @@ func (c *SimpleController) Run() {
 			waitAfter: 1000,
 		},
 		{
-			run:       c.createPost,
+			run:       c.addReaction,
```
ethervoid: Shouldn't this action be performed after a post is created? even though other user entities are creating posts

loadtest/store/memstore/store.go
```diff
@@ -101,6 +101,16 @@ func (s *MemStore) ChannelPosts(channelId string) ([]*model.Post, error) {
 	return channelPosts, nil
 }
 
+func (s *MemStore) PostsSince(ts int64) ([]model.Post, error) {
+	var posts []model.Post
+	for _, post := range s.posts {
+		if post.CreateAt > ts {
```
ethervoid: Greater or greater and equal?

loadtest/user/userentity/actions.go
```diff
@@ -647,6 +655,15 @@ func (ue *UserEntity) SaveReaction(reaction *model.Reaction) error {
 	return nil
 }
 
+func (ue *UserEntity) DeleteReaction(reaction *model.Reaction) error {
+	_, resp := ue.client.DeleteReaction(reaction)
+	if resp.Error != nil {
+		return resp.Error
+	}
+
+	return nil
```
ethervoid: No need to update the current storage?

loadtest/user/userentity/actions.go
```diff
@@ -647,6 +655,15 @@ func (ue *UserEntity) SaveReaction(reaction *model.Reaction) error {
 	return nil
 }
 
+func (ue *UserEntity) DeleteReaction(reaction *model.Reaction) error {
+	_, resp := ue.client.DeleteReaction(reaction)
```
ethervoid: `DeleteReaction` returns a boolean that maybe we should also check an return error if its false

loadtest/control/simplecontroller/actions.go
streamer45: We could but I also think it's fine to return if nothing is found. I am trying not to put too much logic into what the `SimpleController` does at the moment. We can always go back to this later to make it more realistic.

loadtest/control/simplecontroller/controller.go
streamer45: Not really. We are usually not trying to react to our own post. Also the order of these actions might change soon. So unless one action is strictly required to run before another (like with login) I wouldn't worry too much.

loadtest/store/memstore/store.go
streamer45: I think the MM API uses greater but it's not that important. This is more of a utility function that I used to avoid picking posts completely at random.

loadtest/user/userentity/actions.go
streamer45: Yeah, we probably should. Will do.

loadtest/user/userentity/actions.go
streamer45: If there's an error with the response that will always be false. Otherwise it should be true. I don't think we need to explicitly check for that.

loadtest/control/simplecontroller/actions.go
```diff
@@ -100,6 +100,55 @@ func (c *SimpleController) createPost() control.UserStatus {
...
+		return c.newErrorStatus(err)
+	}
+	if len(posts) == 0 {
+		return c.newInfoStatus("no posts to add reaction to")
+	}
+
+	err = c.user.SaveReaction(&model.Reaction{
+		UserId:    c.user.Store().Id(),
+		PostId:    posts[0].Id,
+		EmojiName: "grinning",
```
reflog: we should probably randomize this?

loadtest/control/simplecontroller/actions.go
```diff
@@ -100,6 +100,55 @@ func (c *SimpleController) createPost() control.UserStatus {
...
+
+	reactions, err := c.user.Store().Reactions(posts[0].Id)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	if len(reactions) == 0 {
+		return c.newInfoStatus("no reactions to remove")
+	}
+
+	err = c.user.DeleteReaction(&reactions[0])
```
reflog: should probably remove a random reaction, not always the first one

loadtest/control/simplecontroller/actions.go
streamer45: This is meant to be this way. If you notice we don't use any random function in any of the SimpleController's actions. This is because I want to keep it as deterministic as possible.

We'll get to a more realistic controller once we know more about how the system behaves as a whole. There's a tuning process needed to get to that point so for now my goal is to keep the number of variables down to a minimum, hence the lack of randomization.

PR #88, opened 2020-01-09
MM-21555: implement action for patchProfile

agnivade:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21555


loadtest/control/simplecontroller/actions.go
```diff
@@ -129,6 +129,20 @@ func (c *SimpleController) viewChannel() control.UserStatus {
 	*/
 }
 
+func (c *SimpleController) patchUser(username, nickname, firstName, lastName *string) control.UserStatus {
```
streamer45: We are trying to keep actions parameterless when possible. The reason is we want to keep the decision/complexity of what to update inside and not just wrap once again the underlying api call. Also I think this should be named `updateProfile`. The action's name should reflect what the user does at a higher level.

agnivade:
I mistakenly rebased instead of merging. My bad. Old habits die hard :)

loadtest/control/simplecontroller/actions.go
```diff
@@ -149,6 +150,38 @@ func (c *SimpleController) searchUsers() control.UserStatus {
...
+	})
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	return c.newInfoStatus("user patched")
+}
+
+func (c *SimpleController) updateProfileImage() control.UserStatus {
+	// TODO: take this from the config later.
+	imagePath := "./testdata/test.png"
```
streamer45: I would maybe change the filename to `test_profile.png` since we might have other images of different sizes/formats to test other things like file uploads.

loadtest/control/simplecontroller/actions.go
```diff
@@ -149,6 +150,38 @@ func (c *SimpleController) searchUsers() control.UserStatus {
...
+		LastName:  &lastName,
+	})
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	return c.newInfoStatus("user patched")
+}
+
+func (c *SimpleController) updateProfileImage() control.UserStatus {
+	// TODO: take this from the config later.
```
streamer45: Yes, also it might make sense to keep a shared _in memory_ cache for this kind of assets in the future cause we don't want to end up reading the same files 10000 times per minute if we can avoid it :)

agnivade: yes of course. I meant the file path needs to come from the config.

PR #92, opened 2020-01-12
MM-21503  - Loadtest: implement action to join/leave a channel for SimpleController

reflog:
#### Summary
added leave/join
#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-21503

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -66,6 +66,64 @@ func (c *SimpleController) logout() control.UserStatus {
 	return c.newInfoStatus("logged out")
 }
 
+func (c *SimpleController) joinChannel() control.UserStatus {
+	userStore := c.user.Store()
+	userId := userStore.Id()
+	teams, err := userStore.Teams()
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	for _, team := range teams {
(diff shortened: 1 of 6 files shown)


loadtest/store/store.go
```diff
@@ -18,6 +18,8 @@ type UserStore interface {
 	Config() model.Config
 	// Channels return the channels for a team.
 	Channels(teamId string) ([]model.Channel, error)
+	// ChannelMember returns the ChannelMember for the given channelId and userId.
+	ChannelMember(channelId, userId string) (*model.ChannelMember, error)
```
streamer45: By moving this to the `UserStore` we should make it return a value instead of a pointer.

loadtest/control/simplecontroller/actions.go
```diff
@@ -66,6 +66,64 @@ func (c *SimpleController) logout() control.UserStatus {
...
+		channels, err := userStore.Channels(team.Id)
+		for _, channel := range channels {
+			if err != nil {
+				return c.newErrorStatus(err)
+			}
+			cm, err := userStore.ChannelMember(team.Id, userId)
+			if err != nil {
+				return c.newErrorStatus(err)
+			}
+			if cm.UserId != "" {
```
streamer45: Can this ever be empty in a valid `model.ChannelMember` object?

PR #93, opened 2020-01-12
Add a tool to watch metrics

esdrasbeleza:
Following the discussion we had in [our other pull request](https://github.com/mattermost/mattermost-docker/pull/435) and [MM-21112](https://mattermost.atlassian.net/browse/MM-21112), I added my changes to a fork of this repository.

I made metricswatcher become a new CLI tool in this repo. Since there was some configuration code shared between loadtest and metricswatcher, I moved it into `config` package. I also ported my code to use `mlog` instead of the logging library I was using before.

Instead of hardcoded, the queries now come from [a JSON file](https://github.com/esdrasbeleza/mattermost-load-test-ng/blob/da20c7162ea85db5fe98455e765b95b0840866e5/config/prometheusqueries.sample.json). There's a threshold there for each metric. If it's reached and `Alert` is true, a warning is going to be printed. You can see the logic in [metrics.go](https://github.com/esdrasbeleza/mattermost-load-test-ng/blob/2fb4d2318e4aa73876b22e075fbac0b1ac34f185/cmd/metricswatcher/metrics.go#L82-L92). I left a `TODO` annotation there in case we need to trigger some more complex event.

Note that at the moment I don't know if the queries in the JSON file are meaningful and correct. Some knowledge about Prometheus and MM server is needed to figure out which metrics are relevant and which queries we need to use. As the filename suggests, they are there to be used as samples.

A [README.md](https://github.com/esdrasbeleza/mattermost-load-test-ng/blob/da20c7162ea85db5fe98455e765b95b0840866e5/cmd/metricswatcher/README.md) file was also added with instructions to run and execute the program.


streamer45:
Thank you @esdrasbeleza :+1:  :tada: 
Will review this soon.

cmd/metricswatcher/prometheushelper/prometheushelper.go
```diff
@@ -0,0 +1,75 @@
...
+		vec, _ := value.(model.Vector)
+
+		if len(vec) == 0 {
+			return 0, errors.New("Vector has length = 0")
+		}
+
+		textValue = vec[0].Value.String()
+	} else {
+		return 0, fmt.Errorf("Expected a vector, got a %s", value.Type().String())
+	}
```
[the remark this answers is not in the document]
esdrasbeleza: Ouch! You're completely right. How did I write this?! 😂

cmd/metricswatcher/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	Query       string
+	Threshold   float64
+	Alert       bool
+}
+
+func checkMetrics(queryFile string) {
+	var (
+		prometheusQueries = readPrometheusQueriesFile(queryFile)
+		prometheusHelper  = createPrometheusHelper()
+	)
```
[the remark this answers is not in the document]
esdrasbeleza: I'm a big fan of `var (` to keep things aligned, you probably noticed!
But okay, I'll change it :)

cmd/metricswatcher/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+
+	for {
+		checkQueries(prometheusHelper, prometheusQueries)
+
+		time.Sleep(time.Duration(configuration.PrometheusConfiguration.UpdateIntervalInMS) * time.Millisecond)
```
reflog: how should this loop terminate?

cmd/metricswatcher/metrics.go
```diff
@@ -0,0 +1,92 @@
...
+	}
+}
+
+func createPrometheusHelper() *prometheushelper.PrometheusHelper {
+	var (
+		prometheusURL         = configuration.PrometheusConfiguration.PrometheusURL
+		prometheusHelper, err = prometheushelper.NewPrometheusHelper(prometheusURL)
+	)
+
```
reflog: same comment about `:=`

cmd/metricswatcher/prometheushealthcheck/healthprovider.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	url.Path = path.Join(url.Path, "/-/healthy")
+
+	return &HealthProvider{url}, nil
+}
+
+func (h *HealthProvider) Check() HealthCheckResult {
+	var (
+		response, err = http.Get(h.healthEndpoint.String())
```
reflog: same

cmd/metricswatcher/metrics.go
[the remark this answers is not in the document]
streamer45: I've noticed and it makes sense. My request from change is just about consistency with the existing code. We hardly use `var`  unless there's a concrete reason to do so. If we declare and initialize at the same time we prefer to use the `:=` form.

cmd/metricswatcher/metrics.go
esdrasbeleza: It won't. It keeps running until you Ctrl-C the application.

cmd/metricswatcher/main.go
```diff
@@ -0,0 +1,50 @@
...
+		return err
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(2)
+
+	go healthcheck()
+	go checkMetrics(jsonQueryFile)
+
+	wg.Wait()
```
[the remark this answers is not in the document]
esdrasbeleza: I removed the waitgroup and I'm using error channels instead.
I keep some calls to `os.Exit` during the Cobra setup process because loadtest was doing it, so I decided to keep it consistent.

cmd/metricswatcher/main.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"os"
+	"sync"
+
+	"github.com/spf13/cobra"
+
+	"github.com/mattermost/mattermost-load-test-ng/config"
+)
+
+var configuration *config.Configuration
```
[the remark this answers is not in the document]
esdrasbeleza: Done!

cmd/metricswatcher/healthcheck.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	for {
+		healthcheckResult := healthCheck.Check()
+
+		if !healthcheckResult.Healthy && healthcheckResult.Error != nil {
+			mlog.Error("Prometheus is not healthy:", mlog.Err(healthcheckResult.Error))
+		}
+
+		time.Sleep(60 * time.Second)
```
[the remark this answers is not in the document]
esdrasbeleza: Done, it's configurable now.

logger/logger.go
```diff
@@ -0,0 +1,35 @@
...
+	EnableConsole bool
+	ConsoleJson   bool
+	ConsoleLevel  string
+	EnableFile    bool
+	FileJson      bool
+	FileLevel     string
+	FileLocation  string
+}
+
+func InitLogger(logSettings *LoggerSettings) {
```
streamer45: Since you made the `logger` package (great idea) I suggest we rename this to just `Init`. Then calling `logger.Init()` will look better than `logger.InitLogger()`.

cmd/metricswatcher/config/config.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		mlog.Error("Failed to get logging config:", mlog.Err(err))
+		os.Exit(1)
+	}
+
+	logger.InitLogger(&cfg.LogSettings)
+}
+
+func ReadConfig(configFilePath string) error {
+	viper.SetConfigName("config")
```
streamer45: We should be looking for `config.metricswatcher`.

config/config.metricscheck.default.json
```diff
@@ -0,0 +1,16 @@
...
+        "FileLevel": "INFO",
+        "FileJson": true,
+        "FileLocation": "metricscheck.log"
+    },
+    "PrometheusConfiguration": {
+        "PrometheusURL": "http://localhost:9090",
+        "MetricsUpdateIntervalInMS": 1000,
+        "HealthcheckUpdateIntervalInMS": 60000
+    }
+}
```
streamer45: I suggest we rename this to `config.metricswatcher.json` to keep consistency with the command that makes use of this.

config/prometheusqueries.sample.json
```diff
@@ -0,0 +1,20 @@
...
+        "Threshold": 0.0,
+        "Alert": false
+    },
+    {
+        "Description": "Average database store time",
+        "Query": "avg(rate(mattermost_db_store_time_sum[5m]))",
+        "Threshold": 0.05,
+        "Alert": true
+    }
+]
```
streamer45: What about including the queries in the main config file so we avoid having two separate things? Since this is required I'd like to make running this command as straightforward as possible.

cmd/metricswatcher/README.md
```diff
@@ -0,0 +1,21 @@
...
+    "PrometheusConfiguration": {
+        "PrometheusURL": "http://localhost:9090",
+        "UpdateIntervalInMS": 1000
+    }
+```
+
+## Running
+
+```
+metricswatcher --config config/config.default.json --queries config/prometheusqueries.sample.json
```
streamer45: This also will need to change of course.

cmd/metricswatcher/main.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+
+	persistentFlags := rootCmd.PersistentFlags()
+	persistentFlags.StringP("config", "c", "", "path to the configuration file to use")
+	persistentFlags.StringP("queries", "q", "", "path to the JSON file with Prometheus queries")
+	cobra.MarkFlagRequired(persistentFlags, "queries")
+
+	if err := rootCmd.Execute(); err != nil {
+		os.Exit(1)
+	}
```
streamer45: I suggest we log the error here. Running without the required config file will fail without notifying the user.

cmd/metricswatcher/main.go
```diff
@@ -0,0 +1,51 @@
...
+
+	errChan := make(chan error, 1)
+	defer close(errChan)
+
+	go healthcheck(errChan, configuration)
+	go checkMetrics(errChan, configuration, jsonQueryFile)
+
+	err = <-errChan
+
+	mlog.Error(err.Error())
```
streamer45: Makes sense to remove this and log this error in the upper layer since we return it.

.gitignore
```diff
@@ -19,6 +19,7 @@ _test
 *mattermost.log
 *npm-debug.log*
 loadtest.log
+metricscheck.log
```
streamer45: Let's rename this to `metricswatcher.log` for consistency.

cmd/metricswatcher/main.go
esdrasbeleza: I moved the logger call there. I also re-enabled the viper settings I had disabled for error output. Some important errors were being silenced, so I considered it's better to have duplicated error output when errors happened than to have no output at all.

config/prometheusqueries.sample.json
esdrasbeleza: Done. This also brought a positive side effect where some lines of code were removed, since we don't have to deal with reading and parsing the extra JSON file. 👌

cmd/metricswatcher/README.md
esdrasbeleza: Updated. I also updated the configuration part, just explaining where to set configuration instead of showing the possible settings (it became too big for a README file).

esdrasbeleza:
I'm trying to solve the conflict with `api/server.go`, but it seems there's something missing in the original master branch:

```
$ make build-osx
Build OSX amd64
env GOOS=darwin GOARCH=amd64 go install -mod=readonly -trimpath ./...
# github.com/mattermost/mattermost-load-test-ng/api
api/server.go:68:43: undefined: u
api/server.go:69:55: undefined: u
make: *** [build-osx] Error 2
```

esdrasbeleza:
Alright, I was able to change the contents of `api/server.go` and solve the conflict, but it seems the original master branch isn't compiling 🤔

streamer45:
> Alright, I was able to change the contents of `api/server.go` and solve the conflict, but it seems the original master branch isn't compiling

Yes, that's expected, we need to merge #99 to fix it.

metricscheck.log
```diff
@@ -0,0 +1,27 @@
+{"level":"info","ts":1579123328.0971992,"caller":"metricswatcher/metrics.go:75","msg":"Request duration = 0.00097710"}
```
streamer45: I think this was added by error.

api/server.go
```diff
@@ -6,15 +6,15 @@ import (
...
-	"github.com/gorilla/mux"
-
 	"github.com/mattermost/mattermost-load-test-ng/cmd/loadtest/config"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/simplecontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/user/userentity"
+
+	"github.com/gofrs/uuid"
```
streamer45: Also this is not needed anymore. We recently removed UUIDs.

esdrasbeleza: My branch still has them. I just moved it back to its original position (my editor moved it around...), so when we merge it'll be gone.

metricscheck.log
esdrasbeleza: Ops, you're right. Misuse of `git add -A` after I updated `.gitignore`.

esdrasbeleza:
There you go!

streamer45:
Great work @esdrasbeleza . Good to merge! :tada:

PR #94, opened 2020-01-13
MM-21676: Fix panic on closed channel

agnivade:
#### Summary

Re-initialize the status channel when starting a run.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21676

### DIFF
--- a/loadtest/loadtest.go
+++ b/loadtest/loadtest.go
@@ -141,6 +141,7 @@ func (lt *LoadTester) Run() error {
 	lt.status.NumUsersAdded = 0
 	lt.status.NumErrors = 0
 	lt.status.StartTime = time.Now()
+	lt.statusChan = make(chan control.UserStatus, lt.config.UsersConfiguration.MaxActiveUsers)
 	go lt.handleStatus()
 	for i := 0; i < lt.config.UsersConfiguration.InitialActiveUsers; i++ {
 		if err := lt.addUser(); err != nil {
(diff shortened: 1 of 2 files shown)


agnivade:
I had already tried to add that. :) Currently, it's not possible to do that without spinning up an active server, because without that, the status channel don't get populated.

streamer45:
> I had already tried to add that. :) Currently, it's not possible to do that without spinning up an active server, because without that, the status channel don't get populated.

Aren't we already doing that in `TestRun`?
Just hacking that test (adding a Run() call) made the panic happen for me.

agnivade:
It's not working for me. I can do that if you prefer.

streamer45:
> It's not working for me. I can do that if you prefer.

Interesting.

```patch
diff --git a/loadtest/loadtest_test.go b/loadtest/loadtest_test.go
index bd46208..79eed15 100644
--- a/loadtest/loadtest_test.go
+++ b/loadtest/loadtest_test.go
@@ -103,6 +103,12 @@ func TestRun(t *testing.T) {
 
        err = lt.Stop()
        require.NoError(t, err)
+
+       err = lt.Run()
+       require.NoError(t, err)
+
+       err = lt.Stop()
+       require.NoError(t, err)
 }
 
 func TestStop(t *testing.T) {
```

This before this PR would panic for me, after it won't.

agnivade:
Oh I'm dumb. :man_facepalming: nvm.

PR #95, opened 2020-01-13
MM-21675: Return error for invalid amount

agnivade:
#### Summary

Return an error when invalid amounts like 0, -2, "test" gets passed in the query param.

While at it, we also add some godocs and refactor the response function to include status code also so that we can pass error responses too. This is to be revamped completely in https://mattermost.atlassian.net/browse/MM-21677.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21675

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -6,22 +6,25 @@ import (
 	"net/http"
 	"strconv"
 
-	"github.com/gofrs/uuid"
-	"github.com/gorilla/mux"
 	"github.com/mattermost/mattermost-load-test-ng/config"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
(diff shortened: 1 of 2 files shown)


api/server.go
```diff
@@ -139,11 +141,16 @@ func (a *API) addUserHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	writeJsonResponse(w, map[string]interface{}{"message": fmt.Sprintf("%d users added", i+1), "error": addError, "status": lt.Status()})
+	writeJsonResponse(w, http.StatusOK, map[string]interface{}{"message": fmt.Sprintf("%d users added", i+1), "error": addError, "status": lt.Status()})
```
streamer45: `i+1` should really be just `i`. 
Same in `removeUserHandler`.

PR #96, opened 2020-01-13
MM-21677: Revamp HTTP API responses

agnivade:
#### Summary

- Add custom Marshal/UnMarshal methods for Status.
- Use a proper response struct for all responses.
- Always add status when possible.
- Return a more apt StatusCreated when load agent gets created.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21677

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -45,15 +45,9 @@ func (a *API) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if config.ConnectionConfiguration.ServerURL == "" {
+	if ok, err := config.IsValid(); !ok {
 		writeResponse(w, http.StatusBadRequest, &APIResponse{
-			Error: fmt.Errorf("ServerURL is not present in config"),
-		})
-		return
-	}
-	if config.ConnectionConfiguration.WebSocketURL == "" {
-		writeResponse(w, http.StatusBadRequest, &APIResponse{
-			Error: fmt.Errorf("WebSocketURL is not present in config"),
+			Error: err,
 		})
 		return
 	}
@@ -79,7 +73,7 @@ func (a *API) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 	a.agents[u.String()] = lt
 
 	writeResponse(w, http.StatusCreated, &APIResponse{
-		Message: fmt.Sprintf("Load-test started with agent id: %s", u.String()),
+		Message: fmt.Sprintf("load-test started with agent id: %s", u.String()),
 	})
 }
 
@@ -88,7 +82,7 @@ func (a *API) getLoadTestById(w http.ResponseWriter, r *http.Request) (*loadtest
 	id := vars["id"]
 	lt, ok := a.agents[id]
 	if !ok {
-		err := fmt.Errorf("Load-test agent with id %s not found", id)
+		err := fmt.Errorf("load-test agent with id %s not found", id)
 		writeResponse(w, http.StatusNotFound, &APIResponse{
 			Error: err,
 		})
@@ -103,14 +97,14 @@ func (a *API) runLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if err = lt.Run(); err != nil {
-		writeResponse(w, http.StatusInternalServerError, &APIResponse{
+		writeResponse(w, http.StatusBadRequest, &APIResponse{
 			Error:  err,
 			Status: lt.Status(),
 		})
 		return
 	}
 	writeResponse(w, http.StatusOK, &APIResponse{
-		Message: "Load-test agent started",
+		Message: "load-test agent started",
 		Status:  lt.Status(),
 	})
 }
@@ -128,7 +122,7 @@ func (a *API) stopLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	writeResponse(w, http.StatusOK, &APIResponse{
-		Message: "Load-test agent stopped",
+		Message: "load-test agent stopped",
 		Status:  lt.Status(),
 	})
 }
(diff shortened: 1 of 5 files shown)


api/server_test.go
```diff
@@ -34,12 +37,39 @@ func TestAPI(t *testing.T) {
...
+		Status(http.StatusCreated).JSON().Object()
+	rawMsg := obj.Value("message").String().Raw()
+	parts := strings.Split(rawMsg, ":")
+	require.Len(t, parts, 2)
+	ltId := strings.TrimSpace(parts[1])
 
 	e.POST(ltId + "/run").Expect().Status(http.StatusOK)
 	e.POST(ltId+"/user/add").WithQuery("amount", 10).Expect().Status(http.StatusOK)
 	e.POST(ltId+"/user/remove").WithQuery("amount", 3).Expect().Status(http.StatusOK)
+	e.POST(ltId+"/user/add").WithQuery("amount", 0).Expect().
```
streamer45: This is unrelated from the PR but since we are touching a lot of things I think it might be a good place to have a discussion.

I am thinking that `/loadagent/{id}/user/add` should really be `/loadagent/{id}/addusers` or even simply `/loadagent/{id}/add`. The reason is that `user` is not really an API entity. You can't get or even create a user. The only API entity is the loadagent. A `GET` to `/loadagent/{id}` should in fact return it.

@reflog, @agnivade I'd like your thoughts on this.

loadtest/status.go
```diff
@@ -0,0 +1,74 @@
...
+	"strings"
+	"time"
+)
+
+// State determines which state a loadtester is in.
+type State int
+
+// Different possible states of a loadtester.
+const (
+	StateUnknown State = iota
```
streamer45: Is there a reason we need this additional state? Isn't a newly created `LoadTester` stopped by default?

loadtest/status.go
```diff
@@ -0,0 +1,74 @@
...
+	case StateStopping:
+		res = "stopping"
+	}
+
+	return json.Marshal(res)
+}
+
+// Status contains various information about the load test.
+type Status struct {
+	State           State     // State of the the load test.
```
streamer45: Double `the`.

loadtest/status.go
agnivade: This is required because of the json marshal/unmarshal thing. We need to map a state with some other unknown string.

streamer45: > This is required because of the json marshal/unmarshal thing. We need to map a state with some other unknown string.

I see. Can't we just fail if it's not one of those four? Why encode something that we know is not valid?

api/server_test.go
reflog: fine by me :)

agnivade: I prefer `/loadagent/{id}/addusers`.

streamer45: Yes that version is definitely more precise. Sounds good to me. We can probably make that a separate ticket.

api/server.go
```diff
@@ -22,23 +22,47 @@ type API struct {
...
+	if config.ConnectionConfiguration.ServerURL == "" {
+		writeResponse(w, http.StatusBadRequest, &APIResponse{
+			Error: fmt.Errorf("ServerURL is not present in config"),
+		})
+		return
+	}
+	if config.ConnectionConfiguration.WebSocketURL == "" {
+		writeResponse(w, http.StatusBadRequest, &APIResponse{
+			Error: fmt.Errorf("WebSocketURL is not present in config"),
+		})
```
streamer45: Maybe it makes more sense to have a `LoadTestConfig.IsValid()` method in `config.go` and have that return an error if something is not valid.

api/server_test.go
agnivade: Created - https://mattermost.atlassian.net/browse/MM-21708

api/server.go
```diff
@@ -54,7 +78,9 @@ func (a *API) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 	lt := loadtest.New(&config, newSimpleController)
 	a.agents[u.String()] = lt
 
-	writeJsonResponse(w, http.StatusOK, map[string]string{"loadAgentId": u.String()})
+	writeResponse(w, http.StatusCreated, &APIResponse{
+		Message: fmt.Sprintf("Load-test started with agent id: %s", u.String()),
```
streamer45: I think this and following errors should be all lowercase.

loadtest/status.go
```diff
@@ -0,0 +1,77 @@
...
+// Different possible states of a loadtester.
+const (
+	StateStopped State = iota
+	StateStarting
+	StateRunning
+	StateStopping
+)
+
+// ErrUnknownState is returned when an unknown state variable is encoded/decoded.
+var ErrUnknownState = errors.New("unknown state")
```
streamer45: I would probably rename this to `ErrInvalidState`. It's a semantic detail really but there are no unknowns here. If it's not one of the declared ones then it's bad data.

api/server.go
```diff
@@ -75,10 +103,16 @@ func (a *API) runLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if err = lt.Run(); err != nil {
-		http.Error(w, err.Error(), http.StatusInternalServerError)
+		writeResponse(w, http.StatusInternalServerError, &APIResponse{
```
streamer45: This is another detail but if we try to run the agent twice without stopping I think it's more like a bad request from the api user and not a real problem with the server.

api/server.go
```diff
@@ -120,53 +165,65 @@ func getAmount(r *http.Request) (int, error) {
...
 	for ; i < amount; i++ {
 		if err := lt.AddUser(); err != nil {
-			addError = err.Error()
+			addError = err
 			break // stop on first error, result is reported as part of status
 		}
 	}
 
-	writeJsonResponse(w, http.StatusOK, map[string]interface{}{"message": fmt.Sprintf("%d users added", i), "error": addError, "status": lt.Status()})
+	writeResponse(w, http.StatusOK, &APIResponse{
```
streamer45: Here (and in `removeUserHandler`) we end up returning `http.StatusOK` even when `lt.AddUser()` fails. 
It might be okay since we specify an error object but if so I'd like to be consistent and do the same in `runLoadAgentHandler` and `stopLoadAgentHandler`.

agnivade: Can you expand what you meant by "do the same" ? AFAI see we are indeed returning statusok everywhere.

streamer45: If Stop fails we are still returning `500`.

PR #100, opened 2020-01-19
[MM-21647] First implementation for the load-test agent coordinator

streamer45:
#### Summary

This PR proposes the first draft implementation for the load-test agent coordinator. This is a separate command that will help in running a load-test across a cluster of load-test agents. It will be the main tool used when running load-tests for production. 

The implementation is made by three main components/packages:

- `coordinator` (the highest level component that handles a cluster of load-test agents and reacts to performance degradation events).
- `cluster` (the layer in charge of incrementing/decrementing the number of concurrently active users across all the available agents and provide aggregated status reports).
- `agent` (a component that acts as a simple HTTP API client that interfaces with the load-test API that we recently implemented.)

I apologize in advance for the complete lack of tests and for some missing documentation. I've had to take a couple of shortcuts within the code and left a few TODOs to try and keep the PR as short as possible and also move development under the time constraints that we are facing (really want to get to a MVP demo next week).

#### Ticket

https://mattermost.atlassian.net/browse/MM-21647

### DIFF
--- a/coordinator/agent/agent.go
+++ b/coordinator/agent/agent.go
@@ -61,10 +61,7 @@ func (a *LoadAgent) AddUsers(n int) error {
 	if err != nil {
 		return err
 	}
-	if err := a.apiRequest(req); err != nil {
-		return err
-	}
-	return nil
+	return a.apiRequest(req)
 }
 
 func (a *LoadAgent) RemoveUsers(n int) error {
@@ -73,10 +70,7 @@ func (a *LoadAgent) RemoveUsers(n int) error {
 	if err != nil {
 		return err
 	}
-	if err := a.apiRequest(req); err != nil {
-		return err
-	}
-	return nil
+	return a.apiRequest(req)
 }
 
 func (a *LoadAgent) Start() error {
@@ -96,8 +90,6 @@ func (a *LoadAgent) Start() error {
 		return err
 	}
 
-	fmt.Printf("agent: agent %s created\n", a.config.Id)
-
 	url = fmt.Sprintf("%s/loadagent/%s/run", a.config.ApiURL, a.config.Id)
 	req, err = http.NewRequest("POST", url, nil)
 	if err != nil {
@@ -106,6 +98,8 @@ func (a *LoadAgent) Start() error {
 	if err := a.apiRequest(req); err != nil {
 		return err
 	}
+
+	fmt.Printf("agent: agent %s created\n", a.config.Id)
 
 	return nil
 }
--- a/coordinator/agent/agent.go
+++ b/coordinator/agent/agent.go
@@ -11,6 +11,8 @@ import (
 
 	"github.com/mattermost/mattermost-load-test-ng/api"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
+
+	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
 // LoadAgent is the object acting as a client to the load-test agent
@@ -99,7 +101,7 @@ func (a *LoadAgent) Start() error {
 		return err
 	}
 
-	fmt.Printf("agent: agent %s created\n", a.config.Id)
+	mlog.Info("agent: agent created", mlog.String("agent_id", a.config.Id))
 
 	return nil
 }
@@ -113,7 +115,9 @@ func (a *LoadAgent) Stop() error {
 	if err := a.apiRequest(req); err != nil {
 		return err
 	}
-	fmt.Printf("agent: agent %s destroyed\n", a.config.Id)
+
+	mlog.Info("agent: agent destroyed", mlog.String("agent_id", a.config.Id))
+
 	return nil
 }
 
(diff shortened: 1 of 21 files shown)


coordinator/agent/agent.go
```diff
@@ -0,0 +1,128 @@
...
+	return nil
+}
+
+func (a *LoadAgent) AddUsers(n int) error {
+	url := fmt.Sprintf("%s/loadagent/%s/addusers?amount=%d", a.config.ApiURL, a.config.Id, n)
+	req, err := http.NewRequest("POST", url, nil)
+	if err != nil {
+		return err
+	}
+	if err := a.apiRequest(req); err != nil {
```
reflog: why not just return 'err' ?

coordinator/agent/agent.go
```diff
@@ -0,0 +1,128 @@
...
+func (a *LoadAgent) RemoveUsers(n int) error {
+	url := fmt.Sprintf("%s/loadagent/%s/removeusers?amount=%d", a.config.ApiURL, a.config.Id, n)
+	req, err := http.NewRequest("POST", url, nil)
+	if err != nil {
+		return err
+	}
+	if err := a.apiRequest(req); err != nil {
+		return err
+	}
+	return nil
```
reflog: why not just return 'err' ?

coordinator/agent/agent.go
```diff
@@ -0,0 +1,128 @@
...
+	url := fmt.Sprintf("%s/loadagent/create?id=%s", a.config.ApiURL, a.config.Id)
+	req, err := http.NewRequest("POST", url, bytes.NewReader(configData))
+	if err != nil {
+		return err
+	}
+	if err := a.apiRequest(req); err != nil {
+		return err
+	}
+
+	fmt.Printf("agent: agent %s created\n", a.config.Id)
```
reflog: this should be log and not printf

coordinator/agent/agent.go
```diff
@@ -0,0 +1,128 @@
...
+func (a *LoadAgent) Stop() error {
+	url := fmt.Sprintf("%s/loadagent/%s", a.config.ApiURL, a.config.Id)
+	req, err := http.NewRequest("DELETE", url, nil)
+	if err != nil {
+		return err
+	}
+	if err := a.apiRequest(req); err != nil {
+		return err
+	}
+	fmt.Printf("agent: agent %s destroyed\n", a.config.Id)
```
reflog: this should be log and not printf

coordinator/cluster/cluster.go
```diff
@@ -0,0 +1,95 @@
...
+		agents: agents,
+		config: config,
+	}, nil
+}
+
+// Run starts all the load-test agents available in the cluster.
+func (c *LoadAgentCluster) Run() error {
+	for _, agent := range c.agents {
+		err := agent.Start()
+		if err != nil {
```
reflog: if we abort on first error, probably should shutdown the already started agents

coordinator/cluster/cluster.go
```diff
@@ -0,0 +1,95 @@
...
+func (c *LoadAgentCluster) IncrementUsers(n int) error {
+	if len(c.agents) == 0 {
+		return nil
+	}
+	// TODO: Make this smarter. Implement an algorithm to make sure users are
+	// distributed evenly across the agents regardless of the input value and number of
+	// agents available.
+	inc := int(n / len(c.agents))
+	for i, agent := range c.agents {
+		fmt.Printf("cluster: adding %d users to %s\n", inc, c.config.Agents[i].Id)
```
reflog: log

coordinator/coordinator.go
```diff
@@ -0,0 +1,81 @@
...
+// load-test agents.
+type Coordinator struct {
+	config  *CoordinatorConfig
+	cluster *cluster.LoadAgentCluster
+	// monitor *Monitor
+}
+
+// Run starts a cluster of load-test agents.
+func (c *Coordinator) Run() error {
+	fmt.Printf("coordinator: ready to drive a cluster of %d load-test agents\n", len(c.config.ClusterConfig.Agents))
```
reflog: log

coordinator/agent/agent.go
streamer45: I guess it was me being extra cautious since I've been burnt by nil error interfaces before. In this case case though it looks like it's perfectly fine to return directly the result of `apiRequest`.

coordinator/agent/agent.go
streamer45: I haven't used `mlog` anywhere in these packages mainly to ease debugging (output looks much cleaner). I can easily convert though.

coordinator/cluster/cluster.go
streamer45: Yes, I like the idea of leaving the state as we found it so it makes sense to shutdown if an agent fails to start especially since we currently exit the program when that happens. That said, I'd prefer to leave this decision to the `coordinator` package.

PR #102, opened 2020-01-21
MM-21668 implemented controller channel creation

reflog:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-21668

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -220,12 +220,74 @@ func (c *SimpleController) removeReaction() control.UserStatus {
 }
 
 func (c *SimpleController) createGroupChannel() control.UserStatus {
-	channelId, err := c.user.CreateGroupChannel([]string{}) // TODO: populate memberIds parameter with other users
+	var userIds []string
+	users, err := c.user.Store().RandomUsers(3)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+	for _, user := range users {
+		userIds = append(userIds, user.Id)
+	}
+
+	channelId, err := c.user.CreateGroupChannel(userIds)
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
 
-	return c.newInfoStatus(fmt.Sprintf("group channel created, id %v", channelId))
+	return c.newInfoStatus(fmt.Sprintf("group channel created, id %v with users %+v", channelId, userIds))
+}
+
+func (c *SimpleController) createPublicChannel() control.UserStatus {
+	team, err := c.user.Store().RandomTeam()
+	if err != nil {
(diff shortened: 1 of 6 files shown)


loadtest/store/memstore/random_test.go
```diff
@@ -61,6 +61,27 @@ func TestRandomTeam(t *testing.T) {
...
+		{Id: id3},
+		{Id: id4},
+	})
+	require.NoError(t, err)
+	u, err := s.RandomUsers(2)
+	require.NoError(t, err)
+	require.Len(t, u, 2)
+
+	_, err = s.RandomUsers(5)
+	require.Error(t, err)
```
streamer45: Let's add a check that the returned error is in fact the one expected.

loadtest/store/memstore/random.go
```diff
@@ -12,8 +12,9 @@ import (
 )
 
 var (
-	ErrEmptyMap   = errors.New("memstore: cannot select from an empty map")
-	ErrEmptySlice = errors.New("memstore: cannot select from an empty slice")
+	ErrEmptyMap     = errors.New("memstore: cannot select from an empty map")
+	ErrEmptySlice   = errors.New("memstore: cannot select from an empty slice")
+	ErrSizeMismatch = errors.New("memstore: cannot select from a slice, not enough elements")
```
streamer45: Detail: I come from a [world](https://en.wikipedia.org/wiki/Sizeof) where `size` and `length` are very different concepts. I propose we change this since what we are really checking is the length. Also `s.users` is a `map` and not a `slice`.

loadtest/control/simplecontroller/actions.go
```diff
@@ -220,12 +220,74 @@ func (c *SimpleController) removeReaction() control.UserStatus {
...
+	}
+
+	channelId, err := c.user.CreateDirectChannel(user.Id)
+
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
 
-	return c.newInfoStatus(fmt.Sprintf("group channel created, id %v", channelId))
+	return c.newInfoStatus(fmt.Sprintf("public channel created, id %v", channelId))
```
streamer45: Should be `direct` channel.

loadtest/control/simplecontroller/actions.go
```diff
@@ -220,12 +220,74 @@ func (c *SimpleController) removeReaction() control.UserStatus {
...
+		Name:   model.NewId(),
+		TeamId: team.Id,
+		Type:   "P",
+	})
+
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+
+	return c.newInfoStatus(fmt.Sprintf("public channel created, id %v", channelId))
```
streamer45: Should be `private channel`.

PR #107, opened 2020-01-22
MM-21897: Fix failing actions in simpleController

agnivade:
#### Summary

- Create channels during the init phase.
- Populate teams and channels after login.
- Set the createAt time during creating a post, and fix the time search range
while searching for posts.
- Choose a channel randomly instead of the first channel.
- Added a viewUser action which populates the users in the store.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21897

### DIFF
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -186,7 +186,7 @@ func (c *SimpleController) createPost() control.UserStatus {
 
 func (c *SimpleController) addReaction() control.UserStatus {
 	// get posts from UserStore that have been created in the last minute
-	posts, err := c.user.Store().PostsSince(time.Now().Unix()*1000 - 60000)
+	posts, err := c.user.Store().PostsSince(time.Now().Add(-1*time.Minute).Unix() * 1000)
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
@@ -209,7 +209,7 @@ func (c *SimpleController) addReaction() control.UserStatus {
 
 func (c *SimpleController) removeReaction() control.UserStatus {
 	// get posts from UserStore that have been created in the last minute
-	posts, err := c.user.Store().PostsSince(time.Now().Unix()*1000 - 60000)
+	posts, err := c.user.Store().PostsSince(time.Now().Add(-1*time.Minute).Unix() * 1000)
 	if err != nil {
 		return c.newErrorStatus(err)
 	}
--- a/loadtest/control/simplecontroller/actions.go
+++ b/loadtest/control/simplecontroller/actions.go
@@ -45,6 +45,18 @@ func (c *SimpleController) login() control.UserStatus {
 
 	c.connect()
 
+	// Populate teams and channels.
+	teamIds, err := c.user.GetAllTeams(0, 100)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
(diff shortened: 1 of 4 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -32,11 +31,17 @@ func (ue *UserEntity) Login() error {
...
 	if resp.Error != nil {
 		return resp.Error
 	}
 
+	// We need to set user again because the user ID does not get set
+	// if a user is already signed up.
+	if err := ue.store.SetUser(loggedUser); err != nil {
+		return err
+	}
+
```
streamer45: I think this was the missing part to make the tests *resumable* without clearing the db every time.

loadtest/control/simplecontroller/actions.go
```diff
@@ -586,6 +593,34 @@ func (c *SimpleController) reload(full bool) control.UserStatus {
...
+	member, err := c.user.Store().RandomChannelMember(channel.Id)
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
+
+	// GetUsersByIds for that userid
+	_, err = c.user.GetUsersByIds([]string{member.UserId})
+	if err != nil {
+		return c.newErrorStatus(err)
+	}
```
streamer45: Do we need to implement `GetUser` or does the webapp actually call `GetUsersByIds` with one id only?

loadtest/user/userentity/actions.go
agnivade: Yep

loadtest/control/simplecontroller/actions.go
agnivade: Not sure. My grepping skills are failing me.

```
10:49:41-~/mattermost/mattermost-webapp$git grep "Client4\.getProfilesByIds"
10:49:45-~/mattermost/mattermost-webapp$git grep "Client4\.getUser"
components/suggestion/generic_user_provider.jsx:                    url={Client4.getUsersRoute() + '/' + item.id + '/image?_=' + (item.last_picture_update || 0)}
utils/utils.jsx:    return Client4.getUsersRoute() + '/' + userIdOrObject.id + '/image?_=' + (userIdOrObject.last_picture_update || 0);
utils/utils.jsx:    return Client4.getUsersRoute() + '/' + userId + '/image/default';
```

agnivade: From my search, I think it uses `GetUsersInChannel` and `GetUsersInTeam`. I think `GetUsersInChannel` is the one we are looking for here. But it seems out of scope for this PR ? Perhaps another PR ?

streamer45: > From my search, I think it uses `GetUsersInChannel` and `GetUsersInTeam`. I think `GetUsersInChannel` is the one we are looking for here. But it seems out of scope for this PR ? Perhaps another PR ?

I guess my doubt was about using:

https://godoc.org/github.com/mattermost/mattermost-server/model#Client4.GetUser

as opposed to:

https://github.com/mattermost/mattermost-server/blob/master/model/client4.go#L943

For fetching a single user like we do there.
But it's good for now, just curious.

agnivade: I got that part. My point was that I think the webapp does not fetch a single user at all. It either fetches all users in a channel, or all users in a team. That's what I gleaned.

PR #108, opened 2020-01-23
[MM-21896] Implement performance monitor and complete feedback loop

streamer45:
#### Summary

PR implements the missing performance monitoring component by extracting some very useful code from the `metricswatcher` tool.
It also adds an initial implementation for the feedback loop that will regulate the number of concurrently active users to run.

This is all **very** experimental still. The idea is that we want to try and keep the loop as simple as possible (hard thing to do).  
I've left various comments throughout the code to explain some decisions and oversimplifications I took.

#### Ticket

https://mattermost.atlassian.net/browse/MM-21896

### DIFF
--- a/cmd/metricswatcher/metrics.go
+++ b/cmd/metricswatcher/metrics.go
@@ -5,13 +5,13 @@ import (
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/cmd/metricswatcher/config"
-	"github.com/mattermost/mattermost-load-test-ng/cmd/metricswatcher/prometheushelper"
+	"github.com/mattermost/mattermost-load-test-ng/coordinator/performance/prometheus"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func checkMetrics(errChan chan error, configuration *config.MetricsWatcherConfiguration) {
-	prometheusHelper, err := prometheushelper.NewPrometheusHelper(configuration.PrometheusConfiguration.PrometheusURL)
+func checkMetrics(errChan chan error, config *config.MetricsWatcherConfiguration) {
+	helper, err := prometheus.NewHelper(config.PrometheusConfiguration.PrometheusURL)
 
 	if err != nil {
 		errChan <- err
@@ -19,15 +19,15 @@ func checkMetrics(errChan chan error, configuration *config.MetricsWatcherConfig
 	}
 
 	for {
-		checkQueries(prometheusHelper, configuration.Queries)
+		checkQueries(helper, config.Queries)
 
-		time.Sleep(time.Duration(configuration.PrometheusConfiguration.MetricsUpdateIntervalInMS) * time.Millisecond)
+		time.Sleep(time.Duration(config.PrometheusConfiguration.MetricsUpdateIntervalInMS) * time.Millisecond)
 	}
 }
 
-func checkQueries(prometheus *prometheushelper.PrometheusHelper, queries []config.PrometheusQuery) {
+func checkQueries(helper *prometheus.Helper, queries []prometheus.Query) {
 	for _, query := range queries {
-		value, err := prometheus.VectorFirst(query.Query)
+		value, err := helper.VectorFirst(query.Query)
 
 		if err != nil {
 			mlog.Error("Error while querying Prometheus:", mlog.String("query_description", query.Description), mlog.Err(err))
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -49,10 +49,10 @@ func (c *Coordinator) Run() error {
 	// to the speed at which metrics are changing.
 
 	// The value of users to be incremented at each iteration.
-	// It should be proportional to the maximum number of users expected to test.
+	// TODO: It should be proportional to the maximum number of users expected to test.
 	const incValue = 8
 	// The value of users to be decremented at each iteration.
-	// It should be proportional to the maximum number of users expected to test.
+	// TODO: It should be proportional to the maximum number of users expected to test.
 	const decValue = 8
 	// The timespan to wait after a performance degradation alert before
 	// incrementing or decrementing users again.
@@ -75,7 +75,7 @@ func (c *Coordinator) Run() error {
 		status := c.cluster.Status()
 		mlog.Debug("coordinator: cluster status:", mlog.Int("active_users", status.ActiveUsers), mlog.Int64("errors", status.NumErrors))
 
-		// supportedUsers should be estimated in a more clever way in the future.
+		// TODO: supportedUsers should be estimated in a more clever way in the future.
 		// For now we say that the supported number of users is the number of active users that ran
 		// for the defined timespan without causing any performance degradation alert.
 		if !lastAlertTime.IsZero() && !perfStatus.Alert && hasPassed(lastAlertTime, restTime) && hasPassed(lastActionTime, restTime) {
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -31,22 +31,16 @@ func (c *Coordinator) Run() error {
 	interruptChannel := make(chan os.Signal, 1)
 	signal.Notify(interruptChannel, os.Interrupt, syscall.SIGINT, syscall.SIGTERM)
 
+	defer c.cluster.Shutdown()
 	if err := c.cluster.Run(); err != nil {
 		mlog.Error("coordinator: running cluster failed", mlog.Err(err))
-		c.cluster.Shutdown()
 		return err
 	}
-	defer c.cluster.Shutdown()
 
-	monitorChan, err := c.monitor.Run()
-	if err != nil {
-		mlog.Error("coordinator: running monitor failed", mlog.Err(err))
-		return err
-	}
+	monitorChan := c.monitor.Run()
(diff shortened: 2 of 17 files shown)


coordinator/coordinator.go
```diff
@@ -35,34 +36,81 @@ func (c *Coordinator) Run() error {
...
+		}
+
+		mlog.Debug("coordinator: supported users", mlog.Int("supported_users", supportedUsers))
+
+		// We give the feedback loop some rest time in case of performance
+		// degradation alerts. We want metrics to stabilize before incrementing/decrementing users again.
+		if lastAlertT.IsZero() || lastActionT.IsZero() || hasPassed(lastActionT, restTime) {
+			if perfStatus.Alert {
+				if err := c.cluster.DecrementUsers(decValue); err != nil {
+					mlog.Error("coordinator: failed to decrement users", mlog.Err(err))
```
reflog: we are not doing anything if we failed to reduce the number of users? maybe delay and retry (if the server is choked)

coordinator/coordinator.go
```diff
@@ -35,34 +36,81 @@ func (c *Coordinator) Run() error {
...
+	var supportedUsers int
+
+	// For now we are keeping all these values constant but in the future they
+	// might change based on the state of the feedback loop.
+	// Ideally we want the value of users to increment/decrement to react
+	// to the speed at which metrics are changing.
+
+	// The value of users to be incremented at each iteration.
+	// It should be proportional to the maximum number of users expected to test.
+	const incValue = 8
```
reflog: should probably add a // TODO: here to catch this later

coordinator/coordinator.go
streamer45: This falls into the "keep it dead simple" part. If we fail here and performance don't improve we'll just wait and try again anyway so I was hesitant to introduce more logic.

cmd/metricswatcher/metrics.go
```diff
@@ -5,29 +5,29 @@ import (
...
-	prometheusHelper, err := prometheushelper.NewPrometheusHelper(configuration.PrometheusConfiguration.PrometheusURL)
+func checkMetrics(errChan chan error, config *config.MetricsWatcherConfiguration) {
+	helper, err := prometheus.NewHelper(config.PrometheusConfiguration.PrometheusURL)
 
 	if err != nil {
 		errChan <- err
 		return
 	}
 
 	for {
```
agnivade: Unrelated to this PR. But we are never exiting from this loop when this is run in a goroutine.

coordinator/performance/prometheus/config.go
```diff
@@ -0,0 +1,14 @@
+package prometheus
+
+type Configuration struct {
```
agnivade: Let's keep the good habit of putting godocs :P

coordinator/performance/prometheus/config.go
```diff
@@ -0,0 +1,14 @@
+package prometheus
```
agnivade: Missing license notice. I realize it is missing in a lot of places. We should do a full cleanup some time.

coordinator/coordinator.go
```diff
@@ -35,34 +36,81 @@ func (c *Coordinator) Run() error {
 		c.cluster.Shutdown()
 		return err
 	}
+	defer c.cluster.Shutdown()
```
agnivade: Why don't we move this above the Run call ? Then we can even remove the Shutdown call from inside the if condition.

coordinator/utils.go
```diff
@@ -0,0 +1,19 @@
+package coordinator
+
+import (
+	"time"
+)
+
+// hasPassed returns true if the provided duration has passed since the
```
agnivade: // hasPassed reports whether the provided duration
// added with the given time is before the current time or not.

coordinator/performance/monitor.go
```diff
@@ -0,0 +1,79 @@
...
+	helper     *prometheus.Helper
+	stopChan   chan struct{}
+	statusChan chan Status
+}
+
+func NewMonitor(config MonitorConfig) (*Monitor, error) {
+	if ok, err := config.IsValid(); !ok {
+		return nil, err
+	}
+	helper, err := prometheus.NewHelper(config.PrometheusURL)
```
agnivade: The word "helper" sounds odd to me. From the code, it seems like it wraps a subset of the prometheus APIs. How about `NewWrapper` ? And respectively changing all helpers to wrappers.

coordinator/coordinator.go
```diff
@@ -35,34 +36,81 @@ func (c *Coordinator) Run() error {
...
+
+	monitorChan, err := c.monitor.Run()
+	if err != nil {
+		mlog.Error("coordinator: running monitor failed", mlog.Err(err))
+		return err
+	}
+	defer c.monitor.Stop()
+
+	var lastActionT time.Time
```
agnivade: `var lastActionTime, lastAlertTime time.Time`

coordinator/coordinator.go
```diff
@@ -35,34 +36,81 @@ func (c *Coordinator) Run() error {
...
+				}
+				lastActionT = time.Now()
+			} else if lastAlertT.IsZero() || hasPassed(lastAlertT, restTime) {
+				if status.ActiveUsers < c.config.ClusterConfig.MaxActiveUsers {
+					inc := min(incValue, c.config.ClusterConfig.MaxActiveUsers-status.ActiveUsers)
+					mlog.Info("coordinator: incrementing active users", mlog.Int("num_users", inc))
+					if err := c.cluster.IncrementUsers(inc); err != nil {
+						mlog.Error("coordinator: failed to increment users", mlog.Err(err))
+					}
+					lastActionT = time.Now()
```
agnivade: I see that we are updating lastActionT irrespective of whether Increment/Decrement failed or not. What are the implications of IncrementUsers failing and lastActionT still getting updated ? Will there be any issues with it getting updated but we fail to reduce users, and now it is within the restTime and we don't take any action until then ?

coordinator/performance/config.go
```diff
@@ -0,0 +1,26 @@
+package performance
+
+import (
+	"fmt"
+
+	"github.com/mattermost/mattermost-load-test-ng/coordinator/performance/prometheus"
+)
+
+type MonitorConfig struct {
```
agnivade: godocs please :)

coordinator/performance/config.go
```diff
@@ -0,0 +1,26 @@
...
+	PrometheusURL  string
+	UpdateInterval int
```
agnivade: It's in ms I guess ? Let's mention that with UpdateIntervalMs.

coordinator/performance/monitor.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	return &Monitor{
+		config:     config,
+		helper:     helper,
+		stopChan:   make(chan struct{}),
+		statusChan: make(chan Status),
+	}, nil
+}
+
+func (m *Monitor) Run() (<-chan Status, error) {
```
agnivade: We are not returning any error anywhere. Do you foresee any future cases to return errors ?

coordinator/performance/monitor.go
```diff
@@ -0,0 +1,79 @@
...
+				return
+			case <-time.After(time.Duration(m.config.UpdateInterval) * time.Millisecond):
+			}
+		}
+	}()
+	return m.statusChan, nil
+}
+
+func (m *Monitor) Stop() {
+	close(m.stopChan)
```
agnivade: We should have a way to pass this signal into the runQueries loop. It seems likely that there may be 20-30 queries and the loop is stuck in runQueries and the stop signal never reaches until it exits from runQueries.

We can have a context variable to cancel the loop or pass the signal into the loop. I am fine with whatever you think is better.

coordinator/utils_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+func TestHasPassed(t *testing.T) {
+	tm := time.Now()
+	require.False(t, hasPassed(tm, 1*time.Second))
+	time.Sleep(1 * time.Second)
+	require.True(t, hasPassed(tm, 500*time.Millisecond))
+	require.False(t, hasPassed(tm, 2*time.Second))
+}
+
+func TestMin(t *testing.T) {
```
agnivade: lol :P

streamer45: :trollface:

coordinator/performance/monitor.go
streamer45: I guess running queries could result in some errors but will return those as part of the status so I think we can safely remove this.

coordinator/coordinator.go
streamer45: Worst case scenario we wait `restTime` before attempting to run the action again. 
I don't see any problem setting it only in case of a non error response though.

streamer45:
@agnivade Made the proposed changes. I haven't touched non PR related stuff which I think we can address later on.

coordinator/performance/prometheus/config.go
reflog: @agnivade - can you open a ticket for this please? feel free to assign to me

agnivade: Done https://mattermost.atlassian.net/browse/MM-22000.

PR #111, opened 2020-01-29
MM-21959: Use a middleware to pass the load test

agnivade:
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-21959

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -4,6 +4,7 @@
 package api
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


agnivade:
Code cleanliness-wise yes it's the same. But I think conceptually it makes things simpler.

Previously, each handler called the same function every time. Now it's just being called once. So the handlers themselves don't need to bother about how to get the loadAgentId. They can just access it from the context and get about doing their stuff. The middleware now does that job.

That was my thinking. Let me know what you think.

reflog:
Well, the function isn't "called once", it's called by the middleware for each request - so we just shifted the responsibility to middleware.
In general - I agree with your approach, but it seems it doesn't add anything in this case. I'm ok with merging this, nothing bad here, I just think it's a bit redundant

agnivade:
Fair enough. I'm okay with closing this if you prefer. LMK.

reflog:
I think we can close this. Seems to add more code with little benefit.

PR #113, opened 2020-02-02
MM-22104: First cut of using terraform

agnivade:
#### Summary

- Use a basic terraform config file
- Add the terraform variables to our config file
- Take the variables and apply the configuration

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22104

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -64,9 +64,11 @@ sessions.db
 config/config.json
 config/coordinator.json
 
+# terraform stuff
 /terraformenv
+.terraform
+terraform.tfstate*
+
 .idea
 mattermost-load-test.iml
-
-vendor/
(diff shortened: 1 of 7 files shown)


agnivade:
Actually, to add some color to why I hardcoded this, according to me tests [here](https://mattermost.atlassian.net/browse/MM-21917), there is some variability in instance types, and I found c5 instances to be the most consistent after i3 and p3. Hence I decided to use c5 for all of them.

I think for our load tests, our prime objective would be to get consistent results which was a big problem in the previous iteration. Hence, I would like to keep c5 as hardcoded for now. Later, if we would like to test on specific instances, we can always make it as an external parameter.

PR #115, opened 2020-02-04
MM-22101: reduce login/logout frequency

agnivade:
#### Summary

- During load testing, it was found that doing login/logout on every iteration has a huge CPU cost and essentially takes up the entire CPU profile. Keeping this logic would prevent us from looking into what bottlenecks do we have in the system because all of the time would be spent on hashing the user passwords.

Hence we reduce the frequency to allow other actions to be executed more frequently and allow us to look into what other parts can be optimized.
- Reordered login to be after logout in the loop
so that the frquency of logout/login can be reduced without
leaving the user in a logged out state.
- Doing logout/login once every 20 iterations of the loop.
- Doing signup and login in an initial phase before starting the loop
so that the user is in a logged in state when the loop starts.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22101

### DIFF
--- a/loadtest/control/simplecontroller/controller.go
+++ b/loadtest/control/simplecontroller/controller.go
@@ -190,7 +190,7 @@ func (c *SimpleController) Run() {
 	c.status <- c.signUp()
 	c.status <- c.login()
 
-	cycleCount := 0 // keeps a track of how many times the entire cycle of actions have been completed.
+	cycleCount := 1 // keeps a track of how many times the entire cycle of actions have been completed.
 	for {
 		for i := 0; i < len(actions); i++ {
 			if cycleCount%actions[i].runFrequency == 0 {
--- a/loadtest/control/simplecontroller/controller.go
+++ b/loadtest/control/simplecontroller/controller.go
@@ -71,8 +71,6 @@ func (c *SimpleController) Run() {
 		}
 	}()
 
-	i := 0
-	const loginLogoutFrequency = 20 // login/logout once every 20 times.
 	actions := []UserAction{
 		{
 			run:       c.signUp,
@@ -156,22 +154,14 @@ func (c *SimpleController) Run() {
 			waitAfter: 1000,
 		},
 		{
-			run: func() control.UserStatus {
-				if i%loginLogoutFrequency == 0 {
-					return c.logout()
-				}
-				return c.newInfoStatus("skipping logout")
-			},
-			waitAfter: 1000,
+			run:          c.logout,
+			waitAfter:    1000,
+			runFrequency: 20,
 		},
 		{
-			run: func() control.UserStatus {
-				if i%loginLogoutFrequency == 0 {
-					return c.login()
-				}
-				return c.newInfoStatus("skipping login")
-			},
-			waitAfter: 1000,
+			run:          c.login,
+			waitAfter:    1000,
+			runFrequency: 20,
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


loadtest/control/simplecontroller/controller.go
```diff
@@ -71,15 +71,13 @@ func (c *SimpleController) Run() {
 		}
 	}()
 
+	i := 0
```
streamer45: I am wondering if it would be better to name this with a slightly more meaningful name since it might not be obvious at first glance what is the purpose of it.

loadtest/control/simplecontroller/controller.go
```diff
@@ -158,18 +154,33 @@ func (c *SimpleController) Run() {
...
 
+	c.status <- c.signUp()
+	c.status <- c.login()
+
+	cycleCount := 0 // keeps a track of how many times the entire cycle of actions have been completed.
 	for {
 		for i := 0; i < len(actions); i++ {
-			c.status <- actions[i].run()
+			if actions[i].runFrequency == 0 ||
+				cycleCount%actions[i].runFrequency == 0 {
```
streamer45: I guess calling it *frequency* is a bit counter-intuitive since the greater the value is the less frequently we run the given action. But as long as we make it clear from the docs it shouldn't be a huge problem.

Regarding this check, if we keep this format I don't think 0 is a valid value and I expect a frequency of 1 to mean we run an action on every cycle iteration.

Also, I am wondering if we may want to add a check for `cycleCount == 0` to avoid executing everything on the first iteration which might slightly skew the tests.

agnivade: Is there a better word for "once every x iterations" ? :)

So you want runFrequency of 1 to be set on all actions ? I thought we discussed it would be a non-zero check.

streamer45: > Is there a better word for "once every x iterations" ? :)

Not sure, but I think it's fine as long as we make the behaviour clear. It was more of an afterthought, not a request for change.

> So you want runFrequency of 1 to be set on all actions ? 

I think it should be set to 1 for those actions we want to run on each iteration which is the most part of them at this time.

> I thought we discussed it would be a non-zero check.

Not sure I understood. What I meant above is that on the first iteration when `cycleCount` is 0 we end up running all the actions, regardless of their frequency since the modulo result will be 0.  
I am wondering if this is something we want. Another option to avoid the check is to just start with a `cycleCount` equal to 1. I'd like your thought on this.

agnivade: Ah ok, gotcha. Yes, setting 1 for all other cases is fine I think.

loadtest/control/simplecontroller/controller.go
```diff
@@ -146,30 +157,47 @@ func (c *SimpleController) Run() {
...
 	}
 
 	c.status <- control.UserStatus{ControllerId: c.id, User: c.user, Info: "user started", Code: control.USER_STATUS_STARTED}
 
 	defer c.sendStopStatus()
 
+	c.status <- c.signUp()
+	c.status <- c.login()
+
+	cycleCount := 0 // keeps a track of how many times the entire cycle of actions have been completed.
```
streamer45: This is the last detail I was referring to. I suggest we set `cycleCount` to be starting at 1 to avoid running all the actions on the first iteration.

PR #117, opened 2020-02-05
MM-22217: ability to destroy a terraform environment

agnivade:
#### Summary

We change the parent command to env and add 2 sub-commands
to create and destroy.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22217

### DIFF
--- a/cmd/loadtest/terraform.go
+++ b/cmd/loadtest/terraform.go
@@ -11,7 +11,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-func RunDeployCmdF(cmd *cobra.Command, args []string) error {
+func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 	config, err := loadtest.GetConfig()
 	if err != nil {
 		mlog.Error(err.Error())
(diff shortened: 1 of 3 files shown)


cmd/loadtest/terraform.go
```diff
@@ -26,13 +26,42 @@ func RunDeployCmdF(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func MakeDeployCommand() *cobra.Command {
+func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
+	config, err := loadtest.GetConfig()
+	if err != nil {
+		mlog.Error(err.Error())
+		return nil
```
reflog: any reason not to return `err` ?

terraform/destroy.go
```diff
@@ -0,0 +1,21 @@
...
+	err := t.preFlightCheck()
+	if err != nil {
+		return err
+	}
+
+	err = t.runCommand("destroy",
+		"-auto-approve",
+		"./terraform",
+	)
+	if err != nil {
```
reflog: why not just return `err`?

cmd/loadtest/terraform.go
agnivade: Yes, this is a problem in other places too. The issue with returning an error from a `cobra.Command` is that cobra thinks that there was an issue with the command flag or some incorrect param, and it spits out the entire command help section, which looks a bit odd to me.

It would give something like:
```
INFO[2020-02-05T17:52:18.200545+05:30]                                       caller="terraform/create.go:107"
ERRO[2020-02-05T17:52:18.200576+05:30] exit status 1                                 caller="loadtest/terraform.go:24"
Error: exit status 1
Usage:
  loadtest env create [flags]

Flags:
  -h, --help   help for create

Global Flags:
  -c, --config string   path to the configuration file to use

exit status 1
```

If we return nil, this help section does not get shown.

reflog: Yuck. Thanks for clarifying

PR #118, opened 2020-02-05
MM-22218: create an RDS instance with terraform

agnivade:
I am submitting it against the destroy branch as this is dependant on the subcommand change.

I will submit it against master once the other PR is merged.

#### Summary

- Take variables specifying postgres/mysql.
- Allow only the app servers to access the DB.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22218


terraform/cluster.tf
```diff
@@ -41,6 +29,26 @@ resource "aws_instance" "app_server" {
...
 }
 
+resource "aws_db_instance" "db" {
+  identifier           = "${var.cluster_name}-db-${count.index}"
+  allocated_storage    = 100
+  storage_type         = "io1"
+  iops                 = 1000
+  engine               = var.db_instance_engine
+  engine_version       = var.db_engine_version[var.db_instance_engine]
+  instance_class       = "db.r4.large"
```
reflog: can we make `instance_class` a variable?

terraform/cluster.tf
```diff
@@ -41,6 +29,26 @@ resource "aws_instance" "app_server" {
   ]
 }
 
+resource "aws_db_instance" "db" {
+  identifier           = "${var.cluster_name}-db-${count.index}"
+  allocated_storage    = 100
+  storage_type         = "io1"
+  iops                 = 1000
```
reflog: this is hardcoded on purpose I assume?

loadtest/config.go
```diff
@@ -48,6 +48,10 @@ type DeploymentConfiguration struct {
 	ClusterName      string // Name of the cluster.
 	AppInstanceCount int    // Number of application instances.
 	SSHPublicKey     string // Path to the SSH public key.
+	InstanceCount    int    // Number of DB instances.
+	InstanceEngine   string // Type of the DB instance - postgres or mysql.
```
reflog: we should validate the values passed (aurora-postgres, mysql, etc. if user sets it to some random stuff, errors will appear)

terraform/cluster.tf
agnivade: Yeah .. there is some relation with `iops` and `allocated_storage`, if one is 1000, other can't be less than 100. It will cause more confusion if we parameterize this.

loadtest/config.go
agnivade: done

terraform/cluster.tf
agnivade: done

PR #120, opened 2020-02-10
MM-22000 - added licenses

reflog:
https://mattermost.atlassian.net/browse/MM-22000


cmd/metricswatcher/config/config.go
```diff
@@ -1,3 +1,6 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
 // Copyright (c) 2019 Mattermost, Inc. All Rights Reserved.
 // See License.txt for license information
```
streamer45: This got doubled somehow.

reflog: @streamer45 - yeah, I think that's the custom :)
I'll fix the duplicate.
Also.. Welcome back!

streamer45:
> Wondering why we use the year 2015 for code started from scratch in 2019. Is it custom to do so?

Still curious about this :stuck_out_tongue:

PR #123, opened 2020-02-17
MM-22437 loadtest: Data race in SimpleController websocket event handler

reflog:
#### Summary
Added RW lock to Mutable Store

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22435

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -6,6 +6,7 @@ package memstore
 import (
 	"errors"
 	"sort"
+	"sync"
 
 	"github.com/mattermost/mattermost-server/v5/model"
 )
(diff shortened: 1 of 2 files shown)


loadtest/store/memstore/store.go
```diff
@@ -6,13 +6,15 @@ package memstore
...
 	"sort"
+	"sync"
 
 	"github.com/mattermost/mattermost-server/v5/model"
 )
 
 // MemStore is a simple implementation of MutableUserStore
 // which holds all data in memory.
 type MemStore struct {
+	mutex          sync.RWMutex
```
agnivade: Nit: Let's name it mut or lock for consistency.

PR #124, opened 2020-02-26
[MM-22440] Refactor user actions and start implementation of a simulative controller

streamer45:
#### Summary

PR attempts to minimize code repetition by extracting common user actions.
It also moves some utility functions that were not tied to the `SimpleController` implementation and can be now shared among different controllers.

This is in preparation for the upcoming `SimulController` implementation which is yet missing except for the basic methods which come with some unavoidable repetition.

#### Ticket

https://mattermost.atlassian.net/browse/MM-22440


loadtest/control/utils_test.go
```diff
@@ -0,0 +1,36 @@
...
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRandomizeUserName(t *testing.T) {
+
```
agnivade: unnecessary space

loadtest/control/actions.go
```diff
@@ -0,0 +1,420 @@
...
+	Info string
+	// Err contains an error when the action failed.
+	Err error
+}
+
+// UserAction is a function that simulates a specific behaviour for the provided
+// user.User. It returns a UserActionResponse.
+type UserAction func(user.User) UserActionResponse
+
+func SignUp(u user.User) UserActionResponse {
```
agnivade: These became public functions now and need godocs. We can do it later if you prefer.

streamer45: Yep, it's going to be a separate ticket given it's a non trivial amount of docs.

PR #125, opened 2020-02-28
Add golangci config and enable scopelint linter

streamer45:
#### Summary

PR adds `.golangci.yml` with [scopelint](https://github.com/kyoh86/scopelint) linter enabled. It also fixes the reported issue.
I've explicitly disabled [govet](https://golang.org/cmd/vet/) check for [variable shadowing](https://en.wikipedia.org/wiki/Variable_shadowing).


.golangci.yml
```diff
@@ -0,0 +1,65 @@
...
+    - varcheck
+    - scopelint
+
+issues:
+  exclude-rules:
+    - linters:
+      # ignore unused warnings from enterprise code
+      # add more as required.
+      - unused
+      text: "RedisSupplier|LocalCacheSupplier|Enterprise"
```
streamer45: I guess we can remove this.

.golangci.yml
```diff
@@ -0,0 +1,65 @@
+service:
+  golangci-lint-version: 1.23.6 # Keep this in sync with .circleci/config.yml and Jenkinsfile.pr
```
agnivade: 1.23.6 is not yet in the golangci service. You can use 1.23.3. I have a PR open to extend it till 1.23.6. Also remove the part about Jenkinsfile.pr.

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ commands:
     parameters:
       version:
         type: string
-        default: 1.21.0
+        default: 1.23.6
```
agnivade: Add a note to keep it in sync with .golangci.yml

loadtest/store/memstore/store.go
```diff
@@ -304,7 +304,7 @@ func (s *MemStore) SetChannelMembers(channelMembers *model.ChannelMembers) error
 		return errors.New("memstore: channelMembers should not be nil")
 	}
 
-	for _, member := range *channelMembers {
+	for idx, member := range *channelMembers {
 		// Initialize the maps as necessary.
```
agnivade: That's one way to fix this. The more idiomatic way is to just do a `member := member` inside the loop. We can ignore the idx.

.golangci.yml
agnivade: Yep

.golangci.yml
```diff
@@ -0,0 +1,65 @@
...
+  exclude-rules:
+    - linters:
+      # ignore unused warnings from enterprise code
+      # add more as required.
+      - unused
+      text: "RedisSupplier|LocalCacheSupplier|Enterprise"
+
+    - linters:
+      # ignore golint error for a lot of packages for now
+      - golint
```
agnivade: We can remove this too.

.golangci.yml
```diff
@@ -0,0 +1,65 @@
...
+run:
+  timeout: 5m
+  modules-download-mode: readonly
+
+linters-settings:
+  gofmt:
+    simplify: true
+  govet:
+    check-shadowing: false # set this to true from time to time to check for possible issues
+    disable-all: true
```
agnivade: Tested locally. Remove both `disable-all` and `enable` stanzas. That makes it run everything. Then toggle the `check-shadowing` switch.

loadtest/store/memstore/store.go
streamer45: Wouldn't that be an additional allocation? I like to use index cause we end up pointing to the original object.

.golangci.yml
streamer45: I think that's to avoid getting ids related errors which we are not fixing yet.

.golangci.yml
streamer45: Actually I kind of like the opt in version. We always know what we are enabling and we do the same below for the linters.

agnivade: Fine by me. I was just referring to the offline discussion we were having.

streamer45: Of course. I just wasn't sure what was enabled by default that way since I wasn't getting any error. I think this approach is probably better in the end since it gives us a bit more control over what to check.

loadtest/store/memstore/store.go
agnivade: By additional allocation, yes a new variable on the stack. It redeclares a new pointer which allows any part of the code to ues it freely and therefore prevents it getting pointed to the last item of the slice.

It's a standard way to doing this: https://github.com/golang/go/wiki/commonmistakes#using-reference-to-loop-iterator-variable.

.golangci.yml
agnivade: Oh my bad. I thought those exceptions were copied from the server repo, but now I see these are from this repo.

PR #126, opened 2020-03-02
MM-22701: Fix race in reading statusChan

agnivade:
#### Summary

While stopping a LoadTester, we just stopped reading from the channel
without ensuring the goroutine exited. This cause a race condition
when re-starting an existing loadtester.

We fix this by sending a quit signal which would actually exit
from the goroutine and ensure the channel read is no longer in memory.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22701

### DIFF
--- a/loadtest/loadtest.go
+++ b/loadtest/loadtest.go
@@ -21,6 +21,7 @@ type LoadTester struct {
 	statusChan    chan control.UserStatus
 	status        Status
 	newController NewController
+	quitChan      chan struct{}
 }
 
 // NewController is a factory function that returns a new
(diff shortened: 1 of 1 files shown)


loadtest/loadtest.go
```diff
@@ -19,6 +19,7 @@ type LoadTester struct {
 	config        *Config
 	wg            sync.WaitGroup
 	statusChan    chan control.UserStatus
+	quitChan      chan struct{}
```
streamer45: Interesting. My view of this race was that we just needed to make a safe copy of `statusChan` in `handleStatus()` by leveraging the existing `RWMutex` and range on that copy.  
The goroutine should already exit as a result of closing `statusChan` so adding an additional channel to quit seems counter-intuitive. What am I missing?

agnivade: Hmm .. I seriously need to stop over-thinking. Yeah, that works obviously.

loadtest/loadtest.go
```diff
@@ -32,24 +31,23 @@ type LoadTester struct {
...
-				mlog.Info(st.Err.Error(), mlog.Int("controller_id", st.ControllerId), mlog.String("origin", st.Err.Origin))
-				atomic.AddInt64(&lt.status.NumErrors, 1)
-				continue
-			} else if st.Code == control.USER_STATUS_FAILED {
-				mlog.Error(st.Err.Error())
-				continue
-			}
-			mlog.Info(st.Info, mlog.Int("controller_id", st.ControllerId))
+	var statusChan chan control.UserStatus
+	lt.mut.RLock()
```
streamer45: We need to unlock before calling this goroutine in `Run()` or it will deadlock.

PR #127, opened 2020-03-02
[MM-22438] Fix client side errors in failing actions

streamer45:
#### Summary

PR attempts to fix client side errors that were happening in user actions as a result of either missing or invalid data.

In the end it became a bit more involved than estimated due to some bugfix/additions that were required to make everything work properly.

PR also adds:

- a WebSocket event handler in the `userentity` package used to keep the `UserStore` in sync.
- a couple of random util functions to help picking a channel/team that the user is a member of.
- a fix for the action error origin that was broken by my previous refactor.

#### Ticket

https://mattermost.atlassian.net/browse/MM-22438

### DIFF
--- a/loadtest/control/actions.go
+++ b/loadtest/control/actions.go
@@ -185,8 +185,8 @@ func CreatePost(u user.User) UserActionResponse {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
 	channel, err := u.Store().RandomChannelJoined(team.Id)
-	if err == memstore.ErrNoChannelFound {
-		return UserActionResponse{Info: "no channel found"}
+	if errors.Is(err, memstore.ErrChannelStoreEmpty) {
+		return UserActionResponse{Info: "no channels in store"}
 	} else if err != nil {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
@@ -265,7 +265,7 @@ func RemoveReaction(u user.User) UserActionResponse {
 func CreateGroupChannel(u user.User) UserActionResponse {
 	var userIds []string
 	users, err := u.Store().RandomUsers(3)
-	if err == memstore.ErrLenMismatch {
+	if errors.Is(err, memstore.ErrLenMismatch) {
 		return UserActionResponse{Info: "not enough users to create group channel"}
 	} else if err != nil {
 		return UserActionResponse{Err: NewUserError(err)}
@@ -436,8 +436,8 @@ func ViewUser(u user.User) UserActionResponse {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
 	channel, err := u.Store().RandomChannelJoined(team.Id)
-	if err == memstore.ErrNoChannelFound {
-		return UserActionResponse{Info: "no channel found"}
+	if errors.Is(err, memstore.ErrChannelStoreEmpty) {
+		return UserActionResponse{Info: "no channels in store"}
 	} else if err != nil {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
--- a/loadtest/control/actions.go
+++ b/loadtest/control/actions.go
@@ -155,21 +155,25 @@ func LeaveChannel(u user.User) UserActionResponse {
 func JoinTeam(u user.User) UserActionResponse {
 	userStore := u.Store()
 	userId := userStore.Id()
-	teams, err := userStore.Teams()
+	teamIds, err := u.GetAllTeams(0, 100)
 	if err != nil {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
-	for _, team := range teams {
-		tm, err := userStore.TeamMember(team.Id, userId)
+	for _, teamId := range teamIds {
+		tm, err := userStore.TeamMember(teamId, userId)
 		if err != nil {
 			return UserActionResponse{Err: NewUserError(err)}
 		}
 		if tm.UserId == "" {
-			err := u.AddTeamMember(team.Id, userId)
+			err := u.AddTeamMember(teamId, userId)
 			if err != nil {
 				return UserActionResponse{Err: NewUserError(err)}
 			}
-			return UserActionResponse{Info: fmt.Sprintf("joined team %s", team.Id)}
+			err = u.GetChannelMembersForUser(userId, teamId)
+			if err != nil {
+				return UserActionResponse{Err: NewUserError(err)}
+			}
+			return UserActionResponse{Info: fmt.Sprintf("joined team %s", teamId)}
 		}
 	}
 	return UserActionResponse{Info: "no teams to join"}
--- a/loadtest/control/actions.go
+++ b/loadtest/control/actions.go
@@ -7,8 +7,10 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"time"
 
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/user"
 	"github.com/mattermost/mattermost-server/v5/model"
 )
@@ -38,7 +40,7 @@ func SignUp(u user.User) UserActionResponse {
 
 	err := u.SignUp(email, username, password)
 	if err != nil {
-		return UserActionResponse{Err: err}
+		return UserActionResponse{Err: NewUserError(err)}
 	}
 
 	return UserActionResponse{Info: fmt.Sprintf("signed up as %s", username)}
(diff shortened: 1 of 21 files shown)


loadtest/control/actions.go
```diff
@@ -190,73 +208,84 @@ func AddReaction(u user.User) UserActionResponse {
...
-	return UserActionResponse{Info: "removed reaction"}
+	return UserActionResponse{Info: "no reactions to remove"}
 }
 
 func CreateGroupChannel(u user.User) UserActionResponse {
 	var userIds []string
 	users, err := u.Store().RandomUsers(3)
-	if err != nil {
-		return UserActionResponse{Err: err}
+	if err == memstore.ErrLenMismatch {
```
agnivade: Maybe better to do "errors.Is()" now that errors can be wrapped ?

loadtest/control/actions.go
```diff
@@ -143,34 +155,40 @@ func LeaveChannel(u user.User) UserActionResponse {
 func JoinTeam(u user.User) UserActionResponse {
 	userStore := u.Store()
 	userId := userStore.Id()
-	teams, err := userStore.Teams()
+	teamIds, err := u.GetAllTeams(0, 100)
```
agnivade: What was the issue with userStore.Teams() ?

loadtest/control/actions.go
```diff
@@ -143,34 +155,40 @@ func LeaveChannel(u user.User) UserActionResponse {
...
+	team, err := u.Store().RandomTeamJoined()
 	if err != nil {
-		return UserActionResponse{Err: err}
+		return UserActionResponse{Err: NewUserError(err)}
 	}
-	channel, err := u.Store().RandomChannel(team.Id)
-	if err != nil {
-		return UserActionResponse{Err: err}
+	channel, err := u.Store().RandomChannelJoined(team.Id)
+	if err == memstore.ErrNoChannelFound {
```
agnivade: errors.Is here too ?

loadtest/control/actions.go
```diff
@@ -190,73 +208,84 @@ func AddReaction(u user.User) UserActionResponse {
...
 	if err != nil {
-		return UserActionResponse{Err: err}
+		return UserActionResponse{Err: NewUserError(err)}
 	}
 
-	return UserActionResponse{Info: "added reaction"}
+	return UserActionResponse{Info: fmt.Sprintf("added reaction to post %s", post.Id)}
 }
 
 func RemoveReaction(u user.User) UserActionResponse {
```
agnivade: Should we rename this to `RemoveReactions` then ? Since it is actually removing multiple reactions.

loadtest/control/actions.go
```diff
@@ -352,69 +391,71 @@ func UpdateProfileImage(u user.User) UserActionResponse {
...
+	team, err := u.Store().RandomTeamJoined()
 	if err != nil {
-		return UserActionResponse{Err: err}
+		return UserActionResponse{Err: NewUserError(err)}
 	}
-	channel, err := u.Store().RandomChannel(team.Id)
-	if err != nil {
-		return UserActionResponse{Err: err}
+	channel, err := u.Store().RandomChannelJoined(team.Id)
+	if err == memstore.ErrNoChannelFound {
```
agnivade: Same here too.

loadtest/control/simplecontroller/actions.go
```diff
@@ -181,68 +168,83 @@ func (c *SimpleController) reload(full bool) control.UserActionResponse {
...
 	}
 
-	if teamId != "" {
-		err = c.user.GetChannelsForTeam(teamId)
+	_, err = c.user.GetTeamsForUser(userId)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	team, err := c.user.Store().RandomTeamJoined()
```
agnivade: Yeah .. I was wondering about this too. Preferences never stored the team id. So how does the webapp work ?

loadtest/store/memstore/random.go
```diff
@@ -12,9 +12,13 @@ import (
...
 var (
-	ErrEmptyMap    = errors.New("memstore: cannot select from an empty map")
-	ErrEmptySlice  = errors.New("memstore: cannot select from an empty slice")
-	ErrLenMismatch = errors.New("memstore: cannot select from a map, not enough elements")
+	ErrEmptyMap       = errors.New("memstore: cannot select from an empty map")
+	ErrEmptySlice     = errors.New("memstore: cannot select from an empty slice")
+	ErrLenMismatch    = errors.New("memstore: cannot select from a map, not enough elements")
+	ErrTeamNotFound   = errors.New("memstore: team not found")
+	ErrUserNotSet     = errors.New("memstore: user is not set")
+	ErrNoTeamFound    = errors.New("memstore: no team found")
```
agnivade: ErrTeamNotFound and ErrNoTeamFound ? What's the difference ?

loadtest/store/memstore/random.go
```diff
@@ -12,9 +12,13 @@ import (
...
+	ErrNoChannelFound = errors.New("memstore: no channel found")
```
agnivade: We should be consistent - ErrChannelNotFound / memstore: channel not found.

loadtest/user/userentity/websocket.go
```diff
@@ -17,6 +18,32 @@ const (
 	maxWebsocketFails             = 7
 )
 
+// wsEventHandler handles the given WebSocket event by calling the appropriate
+// store methods to make sure the internal user state is kept updated.
+func (ue *UserEntity) wsEventHandler(ev *model.WebSocketEvent) error {
+	switch ev.EventType() {
```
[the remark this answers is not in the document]
streamer45: Yes. The same event can be handled at two separate levels.  
At the user level it's about keeping the state/store in sync. At the controller level it's about modelling the user's behaviour that could change in response to some event happening.

I think we want to have both handlers. I can improve a bit the documentation on both to make sure this point is clear.

loadtest/store/memstore/store.go
```diff
@@ -470,7 +480,7 @@ func (s *MemStore) SetUsers(users []*model.User) error {
 	s.lock.Lock()
 	defer s.lock.Unlock()
 
-	s.users = make(map[string]*model.User)
+	//s.users = make(map[string]*model.User)
```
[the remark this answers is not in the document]
streamer45: That commented line shows the struggle I am facing right now. I should probably add a `TODO` comment explaining what's the plan.

We currently have slightly different semantics for different store methods that should behave the same. Initially, `Set` meant replacing everything in the store with the given data. It then became more of a "update existing data and add the new one". 

The original reason to have the replace behaviour was to avoid pollution and try to keep the user's state in sync with the server. It then became clear that given that many API calls return partial data we cannot always replace but need a way to simply add/update.

I think the goal is to improve a bit the store API and make sure we have clear and consistent semantics across all store methods.

loadtest/control/actions.go
streamer45: The main issue was that `memstore.Teams()` returned the teams the user is in as consequence of a previous call to `GetTeamsForUser` (`memstore.SetTeams` is affected by the same problem `SetUsers` was).

Also, when joining a team from the webapp you go to the `/select_team` page which does an API call to GetAllTeams so I believe it's correct to make an initial call to fetch all teams as part of the action.

loadtest/control/actions.go
streamer45: I believe it's either removing one or none if it can't find any. We immediately return after removing.

loadtest/store/memstore/random.go
streamer45: There's a subtle difference but there might be a better way to express it. 

- `team not found` means we were explicitly looking for a team (we were passed the teamId) and weren't able to find it.
- `no team found` means we needed to pick one or more teams but we couldn't return any cause there was none to pick.

loadtest/store/memstore/random.go
streamer45: Same reasoning as above.

streamer45:
> LGTM  , just curious about why don't we return pointers in the `store/random.go`? Also, not sure if we want something like this but we can use named returns so that we don't need to `return model.Channel{}, nil`, we can simply just `return` when its appropriate.

Great observations @isacikgoz :+1: 

We don't return pointers because we don't want the caller to be able to modify the stored data. Obviously we could still return pointers of copies to make sure of that.
I am not a fan of returning empty structs either so that might be something to think about. 

I'll let @agnivade comment on the named returns suggestion since he's surely more informed on the subject.

loadtest/user/userentity/websocket.go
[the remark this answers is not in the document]
agnivade: Got it. Let's do that then.

loadtest/store/memstore/store.go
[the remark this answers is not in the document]
agnivade: We could have separate methods like SetUsers/PatchUsers ?

loadtest/control/actions.go
agnivade: Oh my bad. Yes that's right.

loadtest/store/memstore/random.go
agnivade: Maybe something like `empty teams`/`teams not set` is more apt for the 2nd case ?

loadtest/store/memstore/store.go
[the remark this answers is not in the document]
streamer45: Yes, that's a possibility. We've also have the option to pass an additional parameter to determine the behaviour of a `Set` operation if we don't want the store API to explode in the number of methods.

Ideally I'd like to get the naming right without causing too much confusion and that's why I haven't yet made any concrete proposal.

loadtest/store/memstore/random.go
streamer45: Sounds good.

agnivade:
It is certainly possible to use named returns. But there are some tradeoffs to consider. Here is the official stance on the matter: https://github.com/golang/go/wiki/CodeReviewComments#named-result-parameters.

To quote:

> Don't name result parameters just to avoid declaring a var inside the function; that trades off a minor implementation brevity at the cost of unnecessary API verbosity.

I don't see the 2 cases mentioned in that link satisfying here.
- We are not using repetetive params of the same type.
- We are not modifying the return param in a deferred closure.

Considering those points, I would suggest to keep the current code. It is slightly more verbose, but clear and immediately readable.

PR #128, opened 2020-03-03
MM-22442: Complete (*Config).IsValid

agnivade:
#### Summary

- Add more validation checks
- Run the validation while initializing a *loadtest.Tester.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22442

### DIFF
--- a/loadtest/config.go
+++ b/loadtest/config.go
@@ -13,14 +13,6 @@ import (
 	"github.com/spf13/viper"
 )
 
-type Config struct {
-	ConnectionConfiguration ConnectionConfiguration
-	InstanceConfiguration   InstanceConfiguration
-	UsersConfiguration      UsersConfiguration
-	DeploymentConfiguration DeploymentConfiguration
-	LogSettings             logger.Settings
-}
-
 type ConnectionConfiguration struct {
 	ServerURL                   string
 	WebSocketURL                string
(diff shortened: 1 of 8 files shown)


agnivade:
A problem with the current approach is that different components of the repo has different config requirements. The deployment configuration is not necessary for a loadtest to run, but it is necessary for a terraform deployment to happen.

For now, I have just checked for the essential fields needed to run a load test. Let me know if you had something else in mind.

loadtest/config.go
```diff
@@ -95,14 +95,31 @@ func GetConfig() (*Config, error) {
 	return cfg, nil
 }
 
-// IsValid checks whether a config is valid or not.
+// IsValid reports whether a config is valid or not.
 func (c *Config) IsValid() (bool, error) {
 	if c.ConnectionConfiguration.ServerURL == "" {
```
streamer45: I am wondering if it makes sense for each Config type to have its own `IsValid()` method.
For example we could have a `func (c ConnectionConfiguration) IsValid(bool, error)` and leverage that here.

loadtest/config.go
```diff
@@ -95,14 +95,31 @@ func GetConfig() (*Config, error) {
...
 		return false, fmt.Errorf("ServerURL is not present in config")
 	}
 	if c.ConnectionConfiguration.WebSocketURL == "" {
 		return false, fmt.Errorf("WebSocketURL is not present in config")
 	}
+	if c.ConnectionConfiguration.AdminEmail == "" {
```
streamer45: Should we also add checks for the missing fields like `DriverName` and `DataSource` for completeness sake?

streamer45:
> A problem with the current approach is that different components of the repo has different config requirements. The deployment configuration is not necessary for a loadtest to run, but it is necessary for a terraform deployment to happen.

I think the deployment configuration justifies having its own file (same as what happens with the coordinator). This can be part of that general refactoring we discussed.

loadtest/config.go
```diff
@@ -95,14 +95,31 @@ func GetConfig() (*Config, error) {
...
+	if c.InstanceConfiguration.NumTeams <= 0 {
+		return false, fmt.Errorf("NumTeams cannot be <= 0")
+	}
+
+	if c.UsersConfiguration.InitialActiveUsers < 0 {
+		return false, fmt.Errorf("InitialActiveUsers cannot be < 0")
+	}
+	if c.UsersConfiguration.MaxActiveUsers <= 0 {
+		return false, fmt.Errorf("MaxActiveUsers cannot be <= 0")
+	}
```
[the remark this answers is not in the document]
agnivade: done.

loadtest/config.go
agnivade: Actually those are not used at all I think. Maybe we can even remove them from the config.

loadtest/loadtest.go
```diff
@@ -177,16 +178,20 @@ func (lt *LoadTester) Status() *Status {
 // New creates and initializes a new LoadTester with given config. A factory
 // function is also given to enable the creation of UserController values from within the
 // loadtest package.
-func New(config *Config, nc NewController) *LoadTester {
+func New(config *Config, nc NewController) (*LoadTester, error) {
 	// TODO: add config validation
```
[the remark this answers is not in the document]
agnivade: Good spot.

loadtest/config.go
agnivade: Done.

loadtest/config.go
streamer45: I think originally they were needed to do some sort of bulkloading which then I tried to avoid as much as possible and instead rely on a single system admin to create the initial state.

There might still be some usefulness to have them for local testing, like creating the initial admin account without using the CLI or resetting the db when needed.

PR #131, opened 2020-03-05
MM-22861: Deploy mattermost with a service file

agnivade:
#### Summary

Deploy mattermost using a service file

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22861


agnivade:
It is based off the branch of https://github.com/mattermost/mattermost-load-test-ng/pull/130. The code works completely fine, but maybe a bit too hard to review. Feel free to wait until https://github.com/mattermost/mattermost-load-test-ng/pull/130 is merged.

ethervoid:
I've added @stylianosrigas that could help us with good practices about Terraform

terraform/create.go
```diff
@@ -96,8 +97,15 @@ func (t *Terraform) Create() error {
 
 			t.updateConfig(ip, sshc, output)
 
+			// Upload service file
+			rdr := strings.NewReader(strings.TrimSpace(serviceFile))
```
ethervoid: I'd use a [provisioner](https://www.terraform.io/docs/provisioners/file.html) for this task

terraform/strings.go
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package terraform
```
ethervoid: I'd create this as a file and copy it to the resources using a provisioner as I mentioned

terraform/create.go
```diff
@@ -96,8 +97,15 @@ func (t *Terraform) Create() error {
 
 			t.updateConfig(ip, sshc, output)
 
+			// Upload service file
```
ethervoid: One question. Why don't we launch every update of the instances concurrently using go routines?

agnivade: Keeping things simple for now. :) We can certainly move to this when we start deploying clusters.

terraform/create.go
agnivade: I wanted to do that initially. As you can see I am already using provisioner for the license. The issue with this is that this is a program asset, not any file passed as a config param. That means this needs to be packaged as part of the binary. Ideally, this would need a "go generate" step which generates the final asset from a file. But I have kept things simple for now. We could certainly move to an asset generation step when things start to get messy.

terraform/strings.go
agnivade: See comment above.

terraform/create.go
ethervoid: :+1:

terraform/create.go
ethervoid: IMHO I think this and the `updateConfig` part is something that should be done by a provisioner that could be a file one, or even something like Chef. 

Terraform should be responsible for firing up the infrastructure and after that, a provisioner should be the one that performs all the necessary customizations per machine. I'm not an expert of course and this is just my opinion but having to code our own provisioner using SSH à la Ansible makes me think if is not better to use something that is provided by Terraform

As I said, I'm not an expert and if you think this is the right call, it works for me 100% :)

agnivade: > Terraform should be responsible for firing up the infrastructure and after that, a provisioner should be the one that performs all the necessary customizations per machine.

Yes, I thought that too initially. And I totally agree with you that if we had an easy option, I would gladly offload these mundane tasks of creating and managing ssh connections to a 3rd party tool. It seemed very strange to me reading the old load-test codebase, to think why are we managing ssh connections manually.

But terraform itself cautions against using provisioners for everything: https://www.terraform.io/docs/provisioners/#provisioners-are-a-last-resort.

> The following sections describe some situations which can be solved with provisioners in principle, but where better solutions are also available. We do not recommend using provisioners for any of the use-cases described in the following sections. Even if your specific use-case is not described in the following sections, we still recommend attempting to solve it using other techniques first, and use provisioners only if there is no other option.

We could certainly use provisioners to upload this service file and the config file. But like I said before, it would need to be yet another config option passed to terraform.
- For the config file to be an option, it is a possibility, and something which @streamer45 has already raised. But there are some things to consider like: how many options do we override from the code, and how many do we let the user configure. This also means we have to document the options we override much more prominently.
Another solution might be to get a `model.Config` object, override the defaults, write it to a temp file, and pass it to the terraform deployment. (This can indeed be a decent solution, and I like it better than the bunch of `jq` hacks that we do. I might investigate this some time later).
- Regarding the service file, for it to work with provisioner, (assuming we don't want to expose a config param for it) we would need to use the same temp file hack as above.

Judging the factors, I actually don't mind using the temp file hack to upload the service file. But the file upload code seems pretty simple and readable to me. Open to @streamer45's suggestions.

The updateConfig part is indeed hacky and I dislike the series of `jq` calls which are there. I actually looked into the best way to edit a json file, and came to the same conclusion as the old codebase, which are these hacky jq calls. Now that I have this model.Config -> temp file idea, I will look into it.

streamer45: My reasoning here would be something like:

"as long as it works and it doesn't take too much effort to make it so". 

It's a utility part of the project not connected to the actual code doing the load-test so we are a bit less concerned about being *perfect*.

If that wasn't the case I would argue that running terraform binary using Go is just as bad as the hacky config part. We are essentially using Go as a scripting language. 
Originally I thought terraform provided an API to do everything the binary does but that doesn't seem to be the case. I am pretty sure @agnivade  took inspiration from the old load-test repository and given the time requirements I think it was the right choice.

My opinion is that we keep on making this work the less optimal (but quicker) way first and at a later time we can all review it and evaluate more idiomatic options.

ethervoid: Good for me ;)

agnivade: > Originally I thought terraform provided an API to do everything

Yep, same here. But I think it exposes that only for its enrerprise customers. For us, invoking the binary is the way to go.

Great discussion guys. :+1:

agnivade:
> To avoid this I propose we add a function to the ssh package that checks for a valid connection to the ssh-agent and also makes sure that agent.List() contains at least one key.

Sounds good to me. Mind if I move this to a new PR ? Seems like out of scope for this one.

> I don't know if this is easily avoidable or not, just thought it was worth mentioning it.

Intersting. I will look into it.

PR #133, opened 2020-03-06
[MM-22884] Handle post related ws events in userentity

streamer45:
#### Summary

PR adds a handler function in `userentity` for the following WebSocket events:

- `model.WEBSOCKET_EVENT_POSTED`
- `model.WEBSOCKET_EVENT_POST_EDITED`
- `model.WEBSOCKET_EVENT_POST_DELETED`

#### Ticket

https://mattermost.atlassian.net/browse/MM-22884

### DIFF
--- a/loadtest/user/userentity/websocket.go
+++ b/loadtest/user/userentity/websocket.go
@@ -19,15 +19,22 @@ const (
 )
 
 func (ue *UserEntity) handlePostEvent(ev *model.WebSocketEvent) error {
+	var data string
+	if el, ok := ev.Data["post"]; !ok {
+		return fmt.Errorf("post data is missing")
+	} else if data, ok = el.(string); !ok {
+		return fmt.Errorf("post data not of type string")
+	}
+
 	var post *model.Post
-	if err := json.Unmarshal([]byte(ev.Data["post"].(string)), &post); err != nil {
+	if err := json.Unmarshal([]byte(data), &post); err != nil {
 		return err
 	}
 
-	evType := ev.EventType()
-	if evType == model.WEBSOCKET_EVENT_POSTED || evType == model.WEBSOCKET_EVENT_POST_EDITED {
+	switch ev.EventType() {
+	case model.WEBSOCKET_EVENT_POSTED, model.WEBSOCKET_EVENT_POST_EDITED:
 		return ue.store.SetPost(post)
-	} else if evType == model.WEBSOCKET_EVENT_POST_DELETED {
+	case model.WEBSOCKET_EVENT_POST_DELETED:
 		return ue.store.DeletePost(post.Id)
 	}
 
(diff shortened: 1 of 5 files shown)


loadtest/user/userentity/websocket.go
```diff
@@ -41,6 +57,12 @@ func (ue *UserEntity) wsEventHandler(ev *model.WebSocketEvent) error {
 		if _, err := ue.store.DeleteReaction(reaction); err != nil {
 			return err
 		}
+	case model.WEBSOCKET_EVENT_POSTED:
```
agnivade: You can just write them as comma separated:

````
case <a>, <b>, <c>: 
// handle it
```

loadtest/user/userentity/websocket.go
```diff
@@ -18,6 +18,22 @@ const (
 	maxWebsocketFails             = 7
 )
 
+func (ue *UserEntity) handlePostEvent(ev *model.WebSocketEvent) error {
+	var post *model.Post
+	if err := json.Unmarshal([]byte(ev.Data["post"].(string)), &post); err != nil {
```
agnivade: I would be extra-cautious and do a check of whether "post" exists or not and whether it is castable to string or not. There have been cluster bugs where map properties don't get serialized properly.

loadtest/user/userentity/websocket.go
```diff
@@ -18,6 +18,22 @@ const (
...
+		return err
+	}
+
+	evType := ev.EventType()
+	if evType == model.WEBSOCKET_EVENT_POSTED || evType == model.WEBSOCKET_EVENT_POST_EDITED {
```
agnivade: A switch-case is preferred here to avoid the repetetive "evType ==" checks.

loadtest/user/userentity/websocket.go
streamer45: That's great :+1:

loadtest/user/userentity/websocket.go
streamer45: I could argue that panicking is our best chance of catching the bug. A rare error here might easily get unnoticed in the huge stream of logs that this tool will generate.

loadtest/user/userentity/websocket.go
streamer45: It's two cases/branches. I usually start considering a `switch` if I am dealing with at least 3.  
I guess it depends whether it's more annoying the repeated comparison or the added syntax.  
To me it's the latter but totally understand your reasoning :+1:

agnivade: Yes I have the threshold of 3 too. I actually count the comparisons also. So that way, it does cross 3. Thats why I suggested. Not a big deal. Feel free to take a call.

streamer45: Thinking more about it I believe the switch works fine in this particular case since the block would only consist of a single call to the store function. Also, now that I know you can handle multiple cases on the same `case` line as you suggested above it does indeed look like the cleanest approach.

loadtest/user/userentity/websocket.go
agnivade: Hmm .. but do we want to crash the entire load test tool just to catch a server bug ? I would think we would just log the error and move on. Panicking is usually an irrecoverable condition for a process, and I dont think thats the case here.

We can come up with ways of making errors more prominent later by using metrics or other log analysis tools, but I think thats an orthogonal concern.

streamer45: Of course and I guess that's the point I was trying to raise with my slightly *trolling* argument. We'll need to be extra careful about errors. I was trying to give a different point of view since we discovered several bugs in the main server code thanks to unexpected panics.

PR #139, opened 2020-03-11
MM-22100  loadtest: implement a NOOP controller

reflog:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-22100

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/noopcontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/simplecontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/simulcontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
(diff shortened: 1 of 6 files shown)


streamer45:
I remember blocking this ticket a while ago cause I didn't fully understand its purpose.

@agnivade 
Could you briefly illustrate how we would use this?
My main doubt is: what does this controller give us more than an idle instance?

agnivade:
I tried to explain all I could in the JIRA ticket. An idle instance is not responding to http requests and not communicating with the DB. There are also not websocket connections.

This controller is about doing just the bare minimum possible for a mattermost server to do - serve one http request which touches the DB, and have one websocket connection open.

The whole idea is to calculate what is the ideal performance of a mattermost server if it practically does nothing of use. This lets us know how close are we to a (theoretical) target, which is our highest achievable performance.

I got the idea from a CDN company that I worked for in the past. Let's say you have a custom build of a web server. How do you know the absolute limit of the performance of that ? Run the tests without any custom stuff. That's essentially the gap caused by adding the custom code.

So basically, let's say we have a base MM server (X). This X is the stage where it's doing the bare minimum possible. And usually we have Y, which is X + `<sth>`, which is the normal state of the server. The objective is to benchmark X and Y, and find out `<sth>`, and get an idea of how much that `<sth>` is.

That's it in short.

streamer45:
Thanks for the additional explanation @agnivade. That helped clearing my initial doubt.

My next question is: do we really need a custom controller to do this? Couldn't we do the same with the `SimpleController` running only a subset of actions? (https://mattermost.atlassian.net/browse/MM-22899 will happen soon).  
In other words, does the logic of user's behaviour change so much to justify a custom implementation?

agnivade:
> My next question is: do we really need a custom controller to do this? Couldn't we do the same with the SimpleController running only a subset of actions? 

Ah I see, didn't know about that ticket.

We could. But I thought having a custom controller might be cleaner. From a user perspective, they don't want to configure anything. Just run one small endpoint and an open ws connection. But a large portion of the code is indeed a copy of SimpleController. Maybe it's better to have it via SimpleController itself. In that case, I am extremely sorry to have wasted @reflog's time. I should have discussed it with you first.

reflog:
@agnivade - thoughts? merge it?

PR #142, opened 2020-03-13
MM-22773: provision a local mattermost build

agnivade:
#### Summary

Overload the `MattermostDownloadURL` to have a file:// prefix
which indicates a local build. In that case, we can download everything
from a stable release, and just copy the binary from local.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-22773

### DIFF
--- a/loadtest/config.go
+++ b/loadtest/config.go
@@ -134,7 +134,7 @@ type DeploymentConfiguration struct {
 	DBInstanceEngine string // Type of the DB instance - postgres or mysql.
 	DBUserName       string // Username to connect to the DB.
 	DBPassword       string // Password to connect to the DB.
-	// URL from where to download Mattermost distribution.
+	// URL from where to download Mattermost release.
 	// This can also point to a local binary path if the user wants to run loadtest
 	// on a custom build. The path should be prefixed with "file://". In that case,
 	// only the binary gets replaced, and the rest of the build comes from the latest
--- a/loadtest/config.go
+++ b/loadtest/config.go
@@ -126,15 +126,20 @@ func (uc *UsersConfiguration) IsValid() error {
 // DeploymentConfiguration contains the necessary data
 // to deploy and provision a load test environment through terraform.
 type DeploymentConfiguration struct {
-	ClusterName           string // Name of the cluster.
-	AppInstanceCount      int    // Number of application instances.
-	SSHPublicKey          string // Path to the SSH public key.
-	DBInstanceCount       int    // Number of DB instances.
-	DBInstanceClass       string // Type of the DB instance.
-	DBInstanceEngine      string // Type of the DB instance - postgres or mysql.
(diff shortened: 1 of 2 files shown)


loadtest/config.go
```diff
@@ -126,15 +126,20 @@ func (uc *UsersConfiguration) IsValid() error {
...
-	MattermostDownloadURL string // URL from where to download Mattermost distribution.
+	ClusterName      string // Name of the cluster.
+	AppInstanceCount int    // Number of application instances.
+	SSHPublicKey     string // Path to the SSH public key.
+	DBInstanceCount  int    // Number of DB instances.
+	DBInstanceClass  string // Type of the DB instance.
+	DBInstanceEngine string // Type of the DB instance - postgres or mysql.
+	DBUserName       string // Username to connect to the DB.
+	DBPassword       string // Password to connect to the DB.
+	// URL from where to download Mattermost distribution.
```
streamer45: Would maybe change "distribution" with "release".

terraform/create.go
```diff
@@ -57,6 +62,22 @@ func (t *Terraform) Create() error {
 		return err
 	}
 
+	uploadBinary := false
+	binaryPath := ""
```
streamer45: Should we use `var` for this?

terraform/create.go
```diff
@@ -104,6 +125,20 @@ func (t *Terraform) Create() error {
...
+				f, err := os.Open(binaryPath)
+				if err != nil {
+					mlog.Error("error opening file", mlog.String("file", binaryPath), mlog.Err(err))
+					return
+				}
+				defer f.Close()
+				if err := sshc.Upload(f, false, "/opt/mattermost/bin/mattermost"); err != nil {
+					mlog.Error("error uploading file", mlog.String("file", binaryPath), mlog.Err(err))
+					return
+				}
```
streamer45: What about making a function out of this? Uploading a file given a path and a `ssh.Client` if we want to keep it here or otherwise even a client method inside `ssh` that only takes the path.

loadtest/config.go
agnivade: Sure.

terraform/create.go
agnivade: Both are okay. We should just be consistent throughout. Have we been using var elsewhere ?

terraform/create.go
agnivade: Good point. I will do this in https://mattermost.atlassian.net/browse/MM-22977.

terraform/create.go
streamer45: It's just because you are setting the defaults. Otherwise the short form is what we always used throughout the repository I believe.

loadtest/config.go
agnivade: done.

terraform/create.go
agnivade: done

PR #147, opened 2020-03-17
[MM-23372] Add currentChannel/currentTeam in memstore

streamer45:
#### Summary

PR adds information about the current team/channel to the `memstore`.
I've also included a couple of fixes to issues with populating the store.
There might be some more places where to use `currentChannel`/`currentTeam` but I'll leave it for another PR.

#### Ticket

https://mattermost.atlassian.net/browse/MM-23372

### DIFF
--- a/loadtest/control/actions.go
+++ b/loadtest/control/actions.go
@@ -175,12 +175,13 @@ func JoinTeam(u user.User) UserActionResponse {
 			return UserActionResponse{Err: NewUserError(err)}
 		}
 		if tm.UserId == "" {
-			err := u.AddTeamMember(teamId, userId)
-			if err != nil {
+			if err := u.AddTeamMember(teamId, userId); err != nil {
 				return UserActionResponse{Err: NewUserError(err)}
 			}
-			err = u.GetChannelMembersForUser(userId, teamId)
-			if err != nil {
+			if err := u.GetChannelsForTeam(teamId); err != nil {
+				return UserActionResponse{Err: NewUserError(err)}
+			}
+			if err := u.GetChannelMembersForUser(userId, teamId); err != nil {
 				return UserActionResponse{Err: NewUserError(err)}
 			}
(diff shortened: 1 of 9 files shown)


loadtest/control/utils_test.go
```diff
@@ -36,7 +36,7 @@ func TestGetErrOrigin(t *testing.T) {
 }
 
 func TestEmulateUserTyping(t *testing.T) {
-	search := "Lorem ipsum dolor sit amet, consectetur adipiscing elit."
+	search := "this is long enough"
```
streamer45: Nothing to do with this PR. It was getting annoying to have to wait 20 seconds just for this.

loadtest/control/actions.go
```diff
@@ -191,9 +192,11 @@ func JoinTeam(u user.User) UserActionResponse {
 
 // CreatePost creates a new post in a random channel by the given user.
 func CreatePost(u user.User) UserActionResponse {
-	team, err := u.Store().RandomTeamJoined()
+	team, err := u.Store().CurrentTeam()
```
agnivade: Curious to know what was the issue here. It's changing from posting to any random team that the user is part of, to posting only in the current team. But that shouldn't be a bug. We can do it, but is it a bug though ?

streamer45: Looking at it now we should probably use the current channel directly. In this case it was more of a logic issue. It doesn't make much sense for a user to post to a team/channel he/she's not currently viewing.
I can update this.

streamer45: On a second thought I think we might want to revert these changes (using the current channel/team in these actions) until I figure out how to properly do this in the `SimulController` case.  
Keeping track of this data might not be trivial and also might break the `SimpleController` as soon as we order actions differently.

streamer45:
So, I actually removed calls to `SetCurrentXXX` from `userentity` and exposed that functionality to the controller. This was done because setting the current team/channel is not really a result of any single API call but of a higher level logic which lies at the `UserController` level.

PR #148, opened 2020-03-18
MM-23127: Fix JSON parse error with TF_LOG set

agnivade:
#### Summary

When TF_LOG would be set, there would be another
stray log line which would be part of the JSON output for
"terraform output -json". This would fail the JSON parsing.

We fix this by removing anything after the last }. This is a hack
and we have to resort to this because we are forced to parse
output from the CLI rather than go through an API.

This will inevitably break if the TF_LOG output contains a } too.
Then we can resort to even nastier hacks.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23127

### DIFF
--- a/deployment/terraform/cluster.tf
+++ b/deployment/terraform/cluster.tf
@@ -46,26 +46,6 @@ resource "aws_instance" "app_server" {
   }
 }
 
-resource "aws_db_instance" "db" {
-  identifier          = "${var.cluster_name}-db"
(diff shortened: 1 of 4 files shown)


agnivade:
We discussed offline and it was decided that it was getting too hacky. And for now we will document that setting any TF_LOG level other than `TRACE` is not advised. Setting the value to `TRACE` doesn't give this warning.

Closing.

PR #152, opened 2020-03-18
MM-23370 - implemented asset fetching action

reflog:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23383


loadtest/control/actions.go
```diff
@@ -490,6 +490,15 @@ func SearchPosts(u user.User) UserActionResponse {
 	return UserActionResponse{Info: fmt.Sprintf("found %d posts", len(list.Posts))}
 }
 
+// FetchStaticAssets parses index.html and fetches static assets mentioned in it
```
agnivade: Nit: period at the end.

loadtest/user/userentity/user.go
```diff
@@ -96,6 +100,76 @@ func (ue *UserEntity) Connect() <-chan error {
...
+func downloadFile(url string) ([]byte, error) {
+	resp, err := http.Get(url)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New("File not found: " + url)
+	}
+
+	defer resp.Body.Close()
```
agnivade: This should be right after the `err != nil` check.

loadtest/user/userentity/user.go
```diff
@@ -96,6 +100,76 @@ func (ue *UserEntity) Connect() <-chan error {
...
+	for {
+		switch tokenizer.Next() {
+		case html.ErrorToken:
+			// End of the document, we're done
+			return links
+		case html.StartTagToken:
+			tag, _ := tokenizer.TagName()
+			nodeType := ""
+			nodeValue := ""
+			if string(tag) != "script" {
```
agnivade: Any reason we are downloading only scripts and not css/images/other stuff ? The browser will download all that's there.

loadtest/user/userentity/user.go
```diff
@@ -96,6 +100,76 @@ func (ue *UserEntity) Connect() <-chan error {
...
+					nodeValue = string(val)
+				}
+				if string(key) == "type" {
+					nodeType = string(val)
+				}
+				if !more {
+					break
+				}
+			}
+			if nodeType == "text/javascript" {
```
agnivade: Actually, "text/javascript" is optional and not required to be present. We can remove this.

reflog:
@agnivade obviously using colly or similar is much better :) I just didn't feel like bringing in more libraries. But I really don't mind, and since I'll be addressing your other comments, I might just as well use colly 😎

streamer45:
> @agnivade obviously using colly or similar is much better :) I just didn't feel like bringing in more libraries. But I really don't mind, and since I'll be addressing your other comments, I might just as well use colly

Yes, ideally we want to avoid dependencies so your thought was the right one especially if it is for just one function. That said, if it helps a lot of with this effort in terms of code quality and time I am okay with bringing it in.

loadtest/user/userentity/user.go
```diff
@@ -96,6 +100,76 @@ func (ue *UserEntity) Connect() <-chan error {
...
+				links = append(links, nodeValue)
+			}
+		}
+	}
+}
+
+// FetchStaticAssets parses index.html and fetches static assets mentioned in it
+func (ue *UserEntity) FetchStaticAssets() error {
+	indexHTML, err := downloadFile(ue.client.Url)
```
streamer45: I am wondering if we might need to pass more than this. Do the requests need to be authenticated in order to fetch the proper page/assets?

reflog: @streamer45 - nope. non authenticated.
the stuff that comes after (avatars) needs authentication

streamer45: We might want to use the same behaviour when switching channels/teams so support for auth would be nice to have. I think we should have everything we need in the `UserEntity`.

reflog: I'm going to use `colly` for html scraping, so I'm deleting my `downloadFile` function for now.
When we add asset fetching to authenticated parts - I'll handle it properly :)

streamer45: Can you make a follow-up Jira ticket?

reflog: Done. https://mattermost.atlassian.net/browse/MM-23432

loadtest/user/userentity/user.go
```diff
@@ -96,6 +97,21 @@ func (ue *UserEntity) Connect() <-chan error {
...
+	c.OnHTML("link[href]", func(e *colly.HTMLElement) {
+		link := e.Attr("href")
+		c.Visit(e.Request.AbsoluteURL(link))
+	})
+	c.OnHTML("script[src]", func(e *colly.HTMLElement) {
+		link := e.Attr("src")
+		c.Visit(e.Request.AbsoluteURL(link))
+	})
+	return c.Visit(ue.client.Url)
+}
```
streamer45: My concern here is that on a real browser most of the stuff is downloaded after the main JS is parsed and executed.
Maybe we can hard-code a few things that we cannot figure out by just looking at the static html.
One example would be fetching config and license.

These are some calls that we make:
```
http://localhost:8065/api/v4/config/client?format=old
http://localhost:8065/api/v4/license/client?format=old
http://localhost:8065/api/v4/plugins/webapp
```

And then as consequence we also fetch all the plugins bundles.

I am okay if you prefer addressing this in a new PR but we need to think about it otherwise we are are not doing a great job at simulating this behaviour.

agnivade: Why can't we just do a maxDepth of 2 if we want to visit links inside the JS ?

streamer45: > Why can't we just do a maxDepth of 2 if we want to visit links inside the JS ?

Mmm. Does it parse JS? I am thinking there might be concatenated URLs or other non primitive format. We can surely try though and see how much it scrapes.

reflog: Guys, I think you are digging too deep :) I sincerely doubt that the static asset fetching is our bottleneck...

PR #157, opened 2020-03-20
MM-23376: provision a Grafana server with terraform

agnivade:
#### Summary

provision a Grafana server with Terraform and setup a datasource pointing to Prometheus. This does not add any dashboards for now. We can do it at a later stage.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23376

### DIFF
--- a/deployment/terraform/metrics.go
+++ b/deployment/terraform/metrics.go
@@ -4,13 +4,19 @@
 package terraform
 
 import (
+	"bytes"
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"net/http"
 	"strings"
 
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
+
+const defaultGrafanaUsernamePass = "admin:admin"
 
 func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent, output *terraformOutput) error {
 	// Updating Prometheus config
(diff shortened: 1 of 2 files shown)


deployment/terraform/metrics.go
```diff
@@ -4,14 +4,20 @@
...
+	"io/ioutil"
+	"net/http"
 	"strings"
 
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
+const defaultGrafanaUsernamePass = "admin:admin"
```
streamer45: Should this come from `deployer.json`?

agnivade: Not really. This is the initial user/pass which is always fixed. The user needs to change the password in the first login.

streamer45: I was suggesting disabling that behaviour with `disable_initial_admin_creation` and set admin username/password as defined by the user.

https://grafana.com/docs/grafana/latest/installation/configuration/#security

Just a thought, especially since this instance is probably going to be accessed from the outside.

agnivade: I haven't given too much thought on security till now. Even prometheus is exposed to the world, and port 8067 too. We can do what you are suggesting for sure. But I want to take a comprehensive sweep of the security side of this once we are done with the MVP. At that time, we can take care of this amongst other things.

streamer45: Sounds like a plan :+1:

PR #158, opened 2020-03-20
MM-23422: Fix apt-get update not working

agnivade:
#### Summary

This happens because cloud init wouldn't have been complete before the machine would be
termed ready. We fix this by checking for the presence of the file `/var/lib/cloud/instance/boot-finished` and then going ahead with the rest of the operations.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23422


agnivade:
PS: Wait for the Grafana PR to get merged.

agnivade:
Fixed the root case. Thanks to @streamer45. Ready for review now.

PR #161, opened 2020-03-21
MM-23141: Upgrade mattermost-server to 5.21

agnivade:
#### Summary

And also fix code breakages that came with it.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23141


agnivade:
@isacikgoz - This is _exactly_ what I was talking about. I upgraded the dependency to 5.21. And boom, my code broke. This should _never_ happen. What if somebody else used the client API ?

loadtest/user/userentity/actions.go
```diff
@@ -335,7 +335,7 @@ func (ue *UserEntity) SearchChannels(teamId string, search *model.ChannelSearch)
 }
 
 func (ue *UserEntity) GetChannelsForTeamForUser(teamId, userId string) ([]*model.Channel, error) {
-	channels, resp := ue.client.GetChannelsForTeamForUser(teamId, userId, "")
+	channels, resp := ue.client.GetChannelsForTeamForUser(teamId, userId, false, "")
```
streamer45: I think we need this `bool` to be a parameter of the function. The webapp client currently calls this with `true`.

At least here we can break the API for now :stuck_out_tongue:

loadtest/user/userentity/actions.go
```diff
@@ -318,7 +318,7 @@ func (ue *UserEntity) GetChannelsForTeam(teamId string) error {
 	if err != nil {
 		return err
 	}
-	channels, resp := ue.client.GetChannelsForTeamForUser(teamId, user.Id, "")
+	channels, resp := ue.client.GetChannelsForTeamForUser(teamId, user.Id, false, "")
```
streamer45: Same as below, let's have the controller decide whether to include or not deleted channels.

ethervoid: Agree

PR #165, opened 2020-03-23
Bump to 1.14.1

agnivade:
#### Summary

Bump to 1.14.1

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,7 +16,7 @@ commands:
     parameters:
       version:
         type: string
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ commands:
     parameters:
       version:
         type: string
-        default: 1.23.3 # Keep this in sync with .golangci.yml
+        default: 1.23.6 # Keep this in sync with .golangci.yml
```
streamer45: Do we still need the comment?

agnivade: We don't. Will remove.

PR #176, opened 2020-03-26
MM-23465: Remove jq calls for config update

agnivade:
#### Summary

We create a model.Config struct and override each setting.
Then we marshal that to JSON and upload the file to the server.

We also fix a minor bug in the switch case where aurora-postgresql
was mispelled.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23465


deployment/terraform/create.go
```diff
@@ -244,33 +245,40 @@ func (t *Terraform) updateAppConfig(ip string, sshc *ssh.Client, output *terrafo
...
+
+	b, err := json.MarshalIndent(cfg, "", "  ")
+	if err != nil {
+		return fmt.Errorf("error in marshalling config: %w", err)
 	}
+
+	err = sshc.Upload(bytes.NewReader(b), "/opt/mattermost/config/config.json", false)
+	if err != nil {
+		return fmt.Errorf("error uploading config.json: %w", err)
+	}
```
streamer45: nit: maybe use the short form since we don't use `err` after this.

agnivade: done. Sorry I mistakenly amended to my previous commit.

PR #178, opened 2020-03-26
MM-23488 - added license file existence check

reflog:
#### Ticket Link
https://mattermost.atlassian.net/browse/MM-23488

### DIFF
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -38,10 +38,8 @@ type Config struct {
 
 // IsValid reports whether a given deployment config is valid or not.
 func (c *Config) IsValid() error {
-	if c.MattermostLicenseFile != "" {
-		if _, err := os.Stat(c.MattermostLicenseFile); err != nil {
-			return fmt.Errorf("license file %s doesn't exist", c.MattermostLicenseFile)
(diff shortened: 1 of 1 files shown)


deployment/config.go
```diff
@@ -37,6 +38,12 @@ type Config struct {
 
 // IsValid reports whether a given deployment config is valid or not.
 func (c *Config) IsValid() error {
+	if c.MattermostLicenseFile != "" {
+		if _, err := os.Stat(c.MattermostLicenseFile); err != nil {
```
agnivade: Maybe better to follow the recommended example if we want to exactly match the error message:
```go
if _, err := os.Stat(filename); os.IsNotExist(err) {
	...
}
```

https://golang.org/pkg/os/#IsNotExist

Otherwise, we can just say "error trying to access license file %s". Either is fine by me.

agnivade: Wait .. I almost forgot. This will trigger only if it is not empty. We should also raise an error if it is empty. I think it might be simpler to just check the variable directly without doing the empty check. And then use `%q` in the error string to help in case it's an empty string.

PR #181, opened 2020-03-27
[MM-23628] Handle login failures in SimulController

streamer45:
#### Summary

PR adds code to handle login failures in `SimulController` with a very simple re-try mechanism.
This will avoid having users continuously failing to run actions cause the initial login didn't succeed.

#### Ticket

https://mattermost.atlassian.net/browse/MM-23628

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -6,6 +6,7 @@ package simulcontroller
 import (
 	"errors"
 	"fmt"
+	"math"
 	"math/rand"
 	"time"
 
@@ -65,17 +66,27 @@ func (c *SimulController) reload(full bool) control.UserActionResponse {
 	return loadTeam(c.user, team)
 }
 
-func (c *SimulController) login() control.UserActionResponse {
-	resp := control.Login(c.user)
-	if resp.Err != nil {
-		return resp
-	}
+func (c *SimulController) login(u user.User) control.UserActionResponse {
+	for {
+		resp := control.Login(u)
+		if resp.Err == nil {
+			err := c.connect()
+			if err == nil {
+				return resp
+			}
+			c.status <- c.newErrorStatus(err)
+		}
 
-	if err := c.connect(); err != nil {
-		return control.UserActionResponse{Err: control.NewUserError(err)}
-	}
+		c.status <- c.newErrorStatus(resp.Err)
 
-	return resp
+		idleTimeMs := time.Duration(math.Round(1000 * c.rate))
+
+		select {
+		case <-c.stop:
+			return control.UserActionResponse{Info: "login canceled"}
+		case <-time.After(idleTimeMs * time.Millisecond):
+		}
+	}
 }
 
 func (c *SimulController) joinTeam(u user.User) control.UserActionResponse {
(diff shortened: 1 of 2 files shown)


streamer45:
> One question: why does login fail in the first place ? If it fails once, is it expected to succeed in the next try ?
> 
> Also, let's apply this to simplecontroller too ?

It usually happens under stress when the database cannot handle that many connections (`Error 1040`).
Also, it's almost never the actual login call to fail but something else in the login action (getting teams for example).

I wanted to treat the whole action as atomic since the user needs that data after the login to operate and also the websocket connection happens only on a successful login action. 

We expect the user to be able to login eventually. We are not disabling users for now.
From my local tests, the user always succeeds on the second attempt if it failed on the first. It's temporary and mostly happens when launching many controllers at the same time.

The other options were:

- keep the user running in a half logged in state (logged in with no websocket connection and/or insufficient data). Until the next login (not very frequent) we would be failing a lot of actions.
- stop the controller if the user fails to log-in.

The second one might be a valid option but requires more thinking and code changes and I believe it can be avoided for now.

Finally, the reason I have not added this to `SimpleController` is because I wasn't sure the added logic was necessary in that scenario. Same goes for `NoopController`.

agnivade:
> and mostly happens when launching many controllers at the same time.

Ah ok, so I think that's the key here. Adding too many users concurrently brings down the system. What if we add a counting semaphore here

```go
go func() {
	controller.Run()
}()
```

 to limit the number of concurrent Run methods from being executed ? We can change the AddUser signature to take the number of users and then pass that info to addUser, have addUser execute controller.Run with a concurrency of say 10 or so. I think that might help the system from being overloaded.

Sorry I do realize I just suggested something completely different from what you submitted. But in my experience, retries almost always point to a more deeper problem. And in a lot of cases, exacerbates the problem rather than abating it. In this case, the DB is already giving error 1040, and we are retrying on top of that, adding yet more load to it, so that it works in the next try. I think the solution might be to reduce the DB load in the first place.

streamer45:
I think you might be focusing on this problem from the opposite side of things (the one of a wise back-end engineer).

The load-test, especially this controller, is also about getting the system to a breaking point. 
We are testing the system we are given, we should not be lenient. If the system doesn't have a way to limit the rate of incoming calls when under heavy load how is this our fault? (it actually might be our fault but here I am speaking from the load-test developer perspective :p).

I imagine simulating a 100k users installation: many users could easily login all at once if say, some event happened (a `@here` in a channel, an alert of some kind). If some of them somehow failed to log-in, are they going to re-try or just back off and call it a day?

With that said, the main problem I see with your proposal is that we lose a lot of control. The decision of how to add/remove users as far as timing and quantity is currently left to the `coordinator` and for a good reason. It needs tight control over the load applied to the target instance.

What we could do (in a separate PR) is implement different mechanisms to add/remove users in the coordinator or, if we really want to avoid the latency involved in API calls, we might have the load-agent itself provide ways to add/rm N users with different timing options. However, I see that as a separate issue from the one this PR is trying to solve.

This is really about avoiding to run a user with an incomplete state.
Should we do a better job at simulating the user re-trying the login? After N failures the user could just give up. Or maybe, as suggested above, we stop the controller. Either way, I think we should not preemptively limit the concurrency.

agnivade:
Ah thanks for getting this discussion back on track. You are right that we lose control, and the decision should be left to coordinator. And it really happens when we are trying to add 100 users at once.

> Should we do a better job at simulating the user re-trying the login? 

This is the crucial point here, and that's why retrying makes sense to me. Thanks.

PR #205, opened 2020-04-06
[MM-23895] Specify required Terraform version

streamer45:
#### Summary

PR adds version requirement for Terraform in documentation.

#### Ticket

https://mattermost.atlassian.net/browse/MM-23895

### DIFF
--- a/docs/terraform_loadtest.md
+++ b/docs/terraform_loadtest.md
@@ -6,7 +6,7 @@ This is the recommended way to load-test a Mattermost instance for production.
 
 ## Prerequisites
 
-- [Terraform](https://learn.hashicorp.com/terraform/getting-started/install)
+- [Terraform](https://learn.hashicorp.com/terraform/getting-started/install). Version 0.12 is required.
 - AWS credentials to be used as described [here](https://www.terraform.io/docs/providers/aws/index.html#authentication)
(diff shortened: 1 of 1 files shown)


streamer45:
> Oh I thought we were going to add a check from the code to verify the version ? We just need to add that in `preFlightCheck`.

We can surely do both but this is about documentation and preventing the user from getting an error by being explicit on the requirements. The code check will be needed if the user didn't read/follow the documentation. It's like a second layer of *security* from misusage.

agnivade:
Yeah, we can do it in a separate PR for sure. But it is good to make the tool more robust.

I am thinking of it as our config validation code. We do document the options which are allowed for the DB like - aurora-mysql and aurora-postgresql. But we validate them anyways to prevent the deployment from failing.

streamer45:
> Yeah, we can do it in a separate PR for sure. But it is good to make the tool more robust.
> 
> I am thinking of it as our config validation code. We do document the options which are allowed for the DB like - aurora-mysql and aurora-postgresql. But we validate them anyways to prevent the deployment from failing.

Yes, :+1: for me.

streamer45:
@agnivade 

Created https://mattermost.atlassian.net/browse/MM-23916 for the additional in-code check.

PR #206, opened 2020-04-06
[MM-23897] Fallback on defaults if expected config is not found

streamer45:
#### Summary

PR adds the behaviour that the documentation mentions. If the expected config file is not found we rely on the default one.

#### Ticket

https://mattermost.atlassian.net/browse/MM-23897

### DIFF
--- a/coordinator/config.go
+++ b/coordinator/config.go
@@ -7,11 +7,11 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/mattermost/mattermost-load-test-ng/config"
 	"github.com/mattermost/mattermost-load-test-ng/coordinator/cluster"
 	"github.com/mattermost/mattermost-load-test-ng/coordinator/performance"
 	"github.com/mattermost/mattermost-load-test-ng/logger"
 
-	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/spf13/viper"
 )
 
@@ -37,7 +37,8 @@ type Config struct {
 func ReadConfig(configFilePath string) (*Config, error) {
 	v := viper.New()
 
-	v.SetConfigName("coordinator")
+	configName := "coordinator"
+	v.SetConfigName(configName)
 	v.AddConfigPath(".")
 	v.AddConfigPath("./config/")
 	// This is needed for the calls from the terraform package to find the config.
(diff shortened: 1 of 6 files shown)


streamer45:
> This PR looks good, but we can't be sure if default config files are OK or exists. How about hard-coding default values into the code? If we can't find the file, we can simply log that command will run with default values (we can also print config here). And we can make all `*.default.json` files `*.sample.json`.

Yes. I agree on all fronts.
This PR is mainly the quickest way to have the code behave consistently with what the documentation says and it will help new users get things started without much editing.

I've created https://mattermost.atlassian.net/browse/MM-23908 to seek an optimal solution and do a good review of all things configurations related.

coordinator/config.go
```diff
@@ -50,7 +50,16 @@ func ReadConfig(configFilePath string) (*Config, error) {
 	}
 
 	if err := v.ReadInConfig(); err != nil {
-		return nil, errors.Wrap(err, "unable to read configuration file")
+		// If we can't find the config let's rely on the default one.
+		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
```
agnivade, line 54: Let's use errors.As ?

loadtest/config.go
```diff
@@ -87,7 +87,7 @@ func (ucc *UserControllerConfiguration) IsValid() error {
 		return fmt.Errorf("could not validate configuration: %w", err)
 	}
 	if ucc.Rate < 0 {
-		return errors.New("rate cannot be < 0")
+		return fmt.Errorf("rate cannot be < 0")
```
agnivade, line 90: I would keep using the stdlib `errors.New` and keep `fmt.Errorf` only for cases when formatting verbs are required.

coordinator/config.go
```diff
@@ -50,7 +50,16 @@ func ReadConfig(configFilePath string) (*Config, error) {
...
+			mlog.Info("falling back to default configuration file")
```
agnivade, line 55: I would probably make that a warning.

coordinator/config.go
```diff
@@ -50,7 +50,16 @@ func ReadConfig(configFilePath string) (*Config, error) {
 	}
 
 	if err := v.ReadInConfig(); err != nil {
-		return nil, errors.Wrap(err, "unable to read configuration file")
+		// If we can't find the config let's rely on the default one.
```
agnivade, line 53: Should we just refactor this to a helper function ? The same code kinda gets repeated everywhere. We probably just need to pass the viper instance and the config name.

streamer45: I've considered this but as soon as I thought about having it in `config` alongside the JSON files I changed my mind.

streamer45:
> What about simulcontroller.json ? Doesn't it need the same treatment too ?

Yes, it won't hurt.

loadtest/config.go
streamer45: Sure, somehow I thought we decided to use `fmt` for everything. Will update.

streamer45: We actually do use `fmt.Errorf` quite a lot for simple strings.

```sh
rg "fmt\.Errorf\(\"[^%]*\"\)"
```

If we want to keep consistency we might need a separate PR.

agnivade: Ok, sounds good. :+1:

streamer45:
> Ah you went for the config package anyways. :)

Yes. Any other inclusion was not ideal.

loadtest/config.go
streamer45: Created https://mattermost.atlassian.net/browse/MM-23914

PR #207, opened 2020-04-07
MM-23873: Add profile handlers

agnivade:
#### Summary

- Reset the MemStore on various occasions.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23873

### DIFF
--- a/loadtest/control/simplecontroller/controller.go
+++ b/loadtest/control/simplecontroller/controller.go
@@ -67,6 +67,7 @@ func (c *SimpleController) Run() {
 			c.status <- c.newErrorStatus(err)
 		}
 		c.user.Cleanup()
+		c.user.ClearUserData()
 		c.sendStopStatus()
 		close(c.stopped)
 	}()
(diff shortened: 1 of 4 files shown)


loadtest/control/simulcontroller/controller.go
```diff
@@ -58,6 +58,7 @@ func (c *SimulController) Run() {
 			c.status <- c.newErrorStatus(err)
 		}
 		c.user.Cleanup()
+		c.user.ClearUserData()
```
streamer45, line 61: That's an interesting one. Is this actually needed or does it simply help the GC to re-collect the memory more quickly?

loadtest/control/simplecontroller/controller.go
```diff
@@ -154,7 +155,14 @@ func (c *SimpleController) createActions(definitions []actionDefinition) error {
...
+			resp := control.Logout(u)
+			if resp.Err != nil {
+				return resp
+			}
+			u.ClearUserData()
+			return resp
+		},
 		"EditPost": control.EditPost,
 		"Reload": func(u user.User) control.UserActionResponse {
 			return c.reload(false)
```
streamer45, line 168: Can we also fix this since we are here? It should be called with `true`.

agnivade: Sure

loadtest/control/simulcontroller/controller.go
agnivade: This is needed in cases when the coordinator stops an agent, and then starts it again. So the agent never shuts down, and unless the store is cleared, will leak memory.

agnivade: Oh you mean the entire controller struct goes out of scope, is it ? Yeah then I guess it's not needed. But maybe good to add it anyways.

streamer45: Yeah, fine with keeping it. Just thought that when we are done with the controller we remove it from the slice and the GC should do the rest.

PR #215, opened 2020-04-08
Use latest release URL to download mattermost

agnivade:
#### Summary

Use latest release URL to download mattermost

Thanks to @jaydeland for setting this up.


deployment/terraform/create.go
```diff
@@ -22,9 +22,8 @@ import (
 
 const cmdExecTimeoutMinutes = 10
 
-// TODO: fetch this dynamically. See IS-327.
 const (
-	latestReleaseURL           = "https://releases.mattermost.com/5.20.1/mattermost-5.20.1-linux-amd64.tar.gz"
+	latestReleaseURL           = "https://latest.mattermost.com/mattermost-enterprise-linux"
 	defaultLoadTestDownloadURL = "https://github.com/mattermost/mattermost-load-test-ng/releases/download/v0.1.0-alpha/mattermost-load-test-ng-v0.1.0-alpha-linux-amd64.tar.gz"
```
streamer45, line 27: Not strictly related but is this actually needed? I didn't know we had this hard-coded here or I would have updated it. Seems to me like an unneeded duplication since we have it in `deployer.default.json`. Thoughts?

agnivade: This is needed because when we use file:// prefix, we just upload the binary. That means we need a release to be installed for it to work.

streamer45: Alright, so we either update it and keep it sync or we need to read the default config for deployer and use `LoadTestDownloadURL` from there.

agnivade: Yes. If the loadtest url is a non-file one, we use that. If not, then we replace it with the latest url to install the latest MM server and then replace the local binary.

PR #218, opened 2020-04-09
MM-23919: Add a default dashboard

agnivade:
#### Summary

Add a default dashboard
#### Ticket Link

https://mattermost.atlassian.net/browse/MM-23919


streamer45:
Very good! Tested and works nicely :+1:

agnivade:
ping @ethervoid :)

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "rate(mattermost_http_request_duration_seconds_sum{instance=~\"$server\"}[5m]) /   rate(mattermost_http_request_duration_seconds_count{instance=~\"$server\"}[5m])",
```
ethervoid, line 664: In case you're showing this in a graphic is better to use `increase` instead of `rate` in order to hve a proper Y axis measurement

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "mattermost_db_master_connections_total{instance=~\"$server\"}",
```
ethervoid, line 558: In case you want to show a counter you should check the `Instant value` to get the current one

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+      "targets": [
+        {
+          "expr": "mattermost_http_websockets_total{instance=~\"$server\"}",
+          "format": "time_series",
+          "interval": "5s",
+          "intervalFactor": 1,
+          "legendFormat": "{{instance}}",
+          "metric": "mattermost_http_websockets_total",
+          "refId": "A",
+          "step": 5
```
ethervoid, line 460: Step should be at least equal the `scraping time` in order to reduce load and avoid squared graphs. See [this ticket](https://mattermost.atlassian.net/browse/MM-24114) where I include all the information about this change

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "irate(mattermost_http_requests_total{instance=~\"$server\"}[1m])",
```
ethervoid, line 347: Please use increase and 5m as time range

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+          "format": "time_series",
+          "interval": "5s",
```
ethervoid, line 349: Remove interval, once we have the `Scrape time` set in the prometheus configuration this interval is set by default to that scrape time

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "irate(mattermost_post_total{instance=~\"$server\"}[1m])",
```
ethervoid, line 241: Same here

deployment/terraform/assets/datasource.yaml
```diff
@@ -0,0 +1,14 @@
+# config file version
+apiVersion: 1
+
+# list of datasources to insert/update depending
+# what's available in the database
+datasources:
+- name: loadtest-source
```
ethervoid, line 7: Please include here the `scrape time` configuration

agnivade:
Thanks for the great review. Let me address those.

deployment/terraform/assets/datasource.yaml
agnivade: Did you mean to set that as part of the `jsonData` field here: https://grafana.com/docs/grafana/latest/administration/provisioning/#example-datasource-config-file ? Because there's no direct field called `scrape time`.

deployment/terraform/assets/dashboard_data.json
agnivade: According to your another comment, the time range should be 2-3 times of the scrape interval. So a time range of 1m should be fine for scrape_interval of 5 seconds. What do you think ?

deployment/terraform/assets/dashboard_data.json
agnivade: I think this is better suited as a line graph to get the pattern of count over time.

deployment/terraform/assets/dashboard_data.json
agnivade: Gotcha.

deployment/terraform/assets/dashboard_data.json
```diff
@@ -0,0 +1,1162 @@
...
+      "pointradius": 5,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "irate(mattermost_process_cpu_seconds_total{instance=~\"$server\"}[1m])* 100",
```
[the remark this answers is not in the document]
agnivade: Scrape time is 5 seconds. So I guess 1m is fine ? Let me know.

deployment/terraform/assets/dashboard_data.json
agnivade: scrape_interval is indeed 5 seconds.

deployment/terraform/assets/dashboard_data.json
agnivade: I will need check it. But does grafana automatically parse jsonData fields ? Is there a known convention that it looks for the `scrapeInterval` field ? Or do I need to specify it somewhere ?

deployment/terraform/assets/dashboard_data.json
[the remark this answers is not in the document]
ethervoid: Yes, 1 minute is fine in that case :)

deployment/terraform/assets/dashboard_data.json
ethervoid: Ok. So you if you want to check "in real time" how many connections you have to the db its ok ;)

deployment/terraform/assets/dashboard_data.json
ethervoid: Ok, didn't know so its ok :)

deployment/terraform/assets/dashboard_data.json
ethervoid: Yes, it's ok :)

deployment/terraform/assets/datasource.yaml
ethervoid: Yes I think it should go in the jsonData and should be calle something like `scrape_time`. Haven't tried yet

deployment/terraform/assets/dashboard_data.json
ethervoid: In the configuration, see below

deployment/terraform/assets/datasource.yaml
agnivade: Seems to be called `timeInterval`.

PR #220, opened 2020-04-09
[MM-24005] Increase limits for ltagent

streamer45:
#### Summary

PR raises some limits, especially the max number of open file descriptors, to avoid the `ltagent` crashing when running more than 500 users.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24005


streamer45:
> Looking great. I have a couple more suggestions to make things better:
> 
> 1. Let's reduce the idle timeout to 15 seconds to recycle connections faster as you suggested earlier:
> 
> ```go
> transport := &http.Transport{
> 		Proxy: http.ProxyFromEnvironment,
> 		DialContext: (&net.Dialer{
> 			Timeout:   30 * time.Second,
> 			KeepAlive: 30 * time.Second,
> 			DualStack: true,
> 		}).DialContext,
> 		MaxIdleConns:          1000,
> 		MaxIdleConnsPerHost:   1000,
> 		IdleConnTimeout:       30 * time.Second,
> 		TLSHandshakeTimeout:   10 * time.Second,
> 		ExpectContinueTimeout: 1 * time.Second,
> 	}
> ```
> 
> 1. Let's also add upload the /etc/sysctl.conf file to the agents too.
> 2. I was also thinking of doing some config validation around `MaxActiveUsers`. Would it be possible to do some basic sanity testing with your current changes+1+2 applied and see what is the max users it supports, and put a cap on that number and validate it from the config, so that the user can never exceed that number ? (This is not for this PR ofcourse, just something to chew on for later)
> 3. Also maybe bump up ltagents from t2.medium to t2.large, to allow agents to run tests for a longer duration of time ?

I like this kind of reviews. Thanks :+1: 

1. During my tests I tried several configurations for that but it didn't seem to have any major effect on the number of opened fds (it's always 2 open fds per controller). I guess the client keeps the connection open for later requests anyway. That said, we can lower them since huge timeouts are not really what we expect if everything is running properly.
2. Thought about it but opted to just modify what we know will help fixing the issue at hand. The default port range `32768 60999` seems enough for our purposes. We can update though, if you think it might help to avoid similar issues in the future.
3. This is something I've been thinking about a lot. I've thought to just set the `MaxActiveUsers` for the coordinator to be the sum of the `MaxActiveUsers` configured for each agent. What I don't like about that is that the `MaxActiveUsers` for the ltagent tells you the max number of users it can simulate (a capacity if you want) while the one in coordinator's config is more like "the target maximum I am interested in". 
4. I've created https://mattermost.atlassian.net/browse/MM-24041 to somewhat address that. Okay with upgrading the default though.

agnivade:
Hmm .. okay 3 is a bit more tricky. Let's do 1 and 2 then.

1. Yes the client keeps the connection open, but it cannot reuse it until it's free. Reducing the timeout will help it do that.
2. Yes I think it will help in making things a bit more robust in situations where someone sets crazy request rates.

streamer45:
> What are your thoughts on making `uploadBatch` a method of sshc ? Not needed in this PR ofcourse. But wanted to see what you think.

That was my original idea but didn't want to include `mlog` in there so ended up doing it as a util.
If we are willing to sacrifice the file specific logging then I can convert, otherwise I prefer to have it as a util until it's really needed somewhere outside the package.

agnivade:
I _think_ we can do a concurrent file upload, and either return on the first error or combine all errors. But let's see later.

PR #241, opened 2020-04-22
Change increase for irate in the cpu usage

ethervoid:
#### Summary

[Here](https://stackoverflow.com/a/55628595) is a wonderful explanation about why should we use rate/irate

#### Ticket Link

You can see the differences in these two screenshots

With increase (check the Y-axis)
Screenshot from 2020-04-22 22-30-20

With irate
Screenshot from 2020-04-22 22-30-10


streamer45:
@ethervoid  probably need to run `make assets` to fix the build.

ethervoid:
@streamer45 Thank you. Fixed the asset problem :)

Hum, irate will show you the "instant" value, last two values, and that works well for CPU given that you probably want to check for peaks of CPU usage but in with the requests is better to use rate instead to check for a more smooth curve.

Also having spikes in the request alerts would lead to having false positives. I'd rather prefer, IMHO, to check for trends that maintain in the time and for that rate is better.

streamer45:
> Hum, irate will show you the "instant" value, last two values, and that works well for CPU given that you probably want to check for peaks of CPU usage but in with the requests is better to use rate instead to check for a more smooth curve.
> 
> Also having spikes in the request alerts would lead to having false positives. I'd rather prefer, IMHO, to check for trends that maintain in the time and for that rate is better.

Sounds good :+1:

PR #244, opened 2020-04-23
MM-24393: Set GOGC=50 for ltagent service

agnivade:
Setting an aggresive GC rate has shown good results when the agent
is under memory pressure. So we use this to increase the longevity
of the agent.

gogc

As we can see from above, the memory climb is much slower in the second case.


agnivade:
Agree.

PR #245, opened 2020-04-23
[MM-24396] Limit amount of stored data in UserStore

streamer45:
#### Summary

In the end stored data turned out to be not the main problem but still I thought a way to limit it (especially posts) would make sense and help with having long running load-testing sessions.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24396

### DIFF
--- a/loadtest/store/memstore/cqueue.go
+++ b/loadtest/store/memstore/cqueue.go
@@ -9,27 +9,27 @@ import (
 
 // CQueue is a basic implementation of a circular queue of fixed size.
 type CQueue struct {
-	data []interface{}
-	new  func() interface{}
-	size int
-	next int
+	data  []interface{}
+	newEl func() interface{}
+	size  int
+	next  int
 }
 
 // NewCQueue creates and returns a pointer to a queue of the given size.
-// The passed new parameter is a function used to allocate an element if data
+// The passed newEl parameter is a function used to allocate an element if data
 // for it is not yet present in the queue.
-func NewCQueue(size int, new func() interface{}) (*CQueue, error) {
+func NewCQueue(size int, newEl func() interface{}) (*CQueue, error) {
 	if size <= 0 {
 		return nil, errors.New("size should be > 0")
 	}
-	if new == nil {
+	if newEl == nil {
 		return nil, errors.New("new should not be nil")
 	}
 	q := &CQueue{
-		data: make([]interface{}, 0, size),
-		new:  new,
-		size: size,
-		next: 0,
+		data:  make([]interface{}, 0, size),
+		newEl: newEl,
+		size:  size,
+		next:  0,
 	}
 	return q, nil
 }
@@ -41,7 +41,7 @@ func (q *CQueue) Get() interface{} {
 		q.data = append(q.data, nil)
 	}
 	if q.data[q.next] == nil {
-		q.data[q.next] = q.new()
+		q.data[q.next] = q.newEl()
 	}
 	el := q.data[q.next]
 	q.next++
--- a/loadtest/store/memstore/cqueue.go
+++ b/loadtest/store/memstore/cqueue.go
@@ -9,10 +9,10 @@ import (
 
 // CQueue is a basic implementation of a circular queue of fixed size.
 type CQueue struct {
-	data  []interface{}
-	newEl func() interface{}
-	size  int
-	next  int
+	data  []interface{}      // The backing data slice.
+	newEl func() interface{} // The factory function used to allocate space for new elements.
+	size  int                // The maximum size (capacity) of the backing data slice.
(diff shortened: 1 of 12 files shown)


loadtest/store/memstore/cqueue.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"errors"
+)
+
+// CQueue is a basic implementation of a circular queue of fixed size.
+type CQueue struct {
+	data []interface{}
+	new  func() interface{}
+	size int
+	next int
```
agnivade, line 15: A small one-liner on what this is would be helpful.

loadtest/store/memstore/store.go
```diff
@@ -615,7 +722,17 @@ func (s *MemStore) SetStatus(userId string, status *model.Status) error {
 		return errors.New("memstore: status should not be nil")
 	}
 
-	s.statuses[userId] = status
+	if userId != status.UserId {
```
agnivade, line 725: Wondering if this check relevant to this PR ? Or just something we did while we are here. No issues either ways, just checking.

loadtest/store/memstore/store.go
```diff
@@ -50,18 +70,69 @@ func (s *MemStore) Clear() {
...
 	s.statuses = map[string]*model.Status{}
+	s.statusesQueue.Reset()
 	s.reactions = map[string][]*model.Reaction{}
 	s.roles = map[string]*model.Role{}
 	s.license = map[string]string{}
 	s.channelViews = map[string]int64{}
 }
 
+func (s *MemStore) setupQueues(config *Config) error {
+	setups := []struct {
```
agnivade, line 90: I was thinking if there was a better way to do this. And I'm sure you have already thought about it. Right now, if one needs to add a new queue, they would need to declare a queue field, a new config max field, and add the setup struct here. Wondering what alternative methods you thought about and decided on this one ?

loadtest/store/memstore/store.go
```diff
@@ -396,7 +475,15 @@ func (s *MemStore) SetChannelMembers(channelMembers *model.ChannelMembers) error
 		if s.channelMembers[cm.ChannelId] == nil {
 			s.channelMembers[cm.ChannelId] = make(map[string]*model.ChannelMember)
 		}
-		s.channelMembers[cm.ChannelId][cm.UserId] = cm
+
+		c := s.channelMembersQueue.Get().(*model.ChannelMember)
```
agnivade, line 479: A quick one-liner before these sections would be helpful for future readers to clarify what this is doing. Something like 
`// We check if the tail of the queue is already full. And if so, we delete the item from the map. This is done to keep the number of items in the map equal to the queue size and prevent unbounded memory growth.`

loadtest/store/memstore/store.go
streamer45: Yeah, nothing to do with the PR, just helps keeping consistent data.

loadtest/store/memstore/store_test.go
```diff
@@ -4,22 +4,30 @@
...
+	"fmt"
 	"testing"
 
 	"github.com/mattermost/mattermost-server/v5/model"
 
 	"github.com/stretchr/testify/require"
 )
 
+func newStore(t *testing.T) *MemStore {
+	s, err := New(nil)
```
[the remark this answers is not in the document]
streamer45: Yeah, just though the overhead was not required also because we almost always need to create a new store. I can still add it for future needs though.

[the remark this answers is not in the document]
streamer45: Oh I think you maybe meant to use `https://golang.org/pkg/testing/#T.Helper`.

loadtest/store/memstore/store.go
streamer45: I've thought about it for a while. In the end it's not like I've saved much repetition but at least we call the queue factory function and do the error checking in one place only, which was my goal.

The easiest approach would be to simply repeat everything for the 4 queues.  
I can revert to that one if the abstraction is confusing. Just didn't think to invest too much effort on an internal setup function.

loadtest/store/memstore/store_test.go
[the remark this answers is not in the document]
agnivade: Yes :)

loadtest/store/memstore/store.go
```diff
@@ -15,31 +16,50 @@ import (
...
+	currentTeam         *model.Team
+	channelViews        map[string]int64
+}
+
+// New returns a new instance of MemStore with the given config.
+// If config is nil, defaults will be used.
+func New(config *Config) (*MemStore, error) {
+	if config == nil {
+		config = &Config{}
+		config.SetDefaults()
```
isacikgoz, line 48: 😎

PR #246, opened 2020-04-23
MM-24313: Bump up instance class of agents

agnivade:
Load test agents require much more CPU and memory than a t2.medium can provide.
Let's bump it higher.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24313


agnivade:
Yep. Added it to my notes.

PR #252, opened 2020-04-24
MM-24397: Reusing the read buffer while reading messages from websockets

agnivade:
#### Summary

The core problem was that conn.ReadMessage allocated a buffer every time it was read. This created heavy amount of allocations _every single time_ we read a message from the websocket.

To avoid this, we bypass the ReadMessage which was more of a helper method, and actually call the NextReader which returns a reader object. We can then reuse a single byte.Buffer instance to read the object unmarshal into a `WebSocketEvent` object. This gets rid of the allocation in the read path completly and allows GC more time to do other tasks.

As we see in the alloc_space profile here:

Initial:
baseheap

After:
reusedheap

There were 2 big boxes in the original profile. Not there's just one. So effectively 10GB of allocation is gone.

---------------------

There is going to be the same PR in the server repo also. But we vendor the code for now until we cut a stable branch in master.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-24397


agnivade:
For ease of review, here is the commit: https://github.com/mattermost/mattermost-load-test-ng/commit/81a9c1e0b605d56f0b03e8ca10a05045ed88312e which contains the changes.

agnivade:
Closing this one. Will send another one.

PR #253, opened 2020-04-25
MM-24542: Tune http client and server for high performance

agnivade:
- We reduce the timeouts at the server side
- We increase the maxConnections on the client
and also reduce the timeouts.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24542


agnivade:
Sure.

PR #255, opened 2020-04-25
MM-21660: Apply advanced sysctl optimizations

agnivade:
We split out the client optimizations from server optimizations
because server configs need some more help to be able to serve
connections at a high rate.

After we improved the websocket client to maintain a semaphore
for client API calls, the number of concurrent HTTP requests
increased by 8*CPU count. On a 2 core machine, this would
result in a 16x increase in connections. Therefore, when we
would have nearly 700 users, the server would be unable
to handle 11200 connections.

The nginx logs clearly showed the problem:
```
2020/04/25 09:07:11 [crit] 3642#3642: *1315161 connect() to 172.31.1.55:8065 failed (99: Cannot assign requested address) while connecting to upstream, client: 172.31.1.148, server: _, request: "POST /api
/v4/users/ids HTTP/1.1", upstream: "http://172.31.1.55:8065/api/v4/users/ids", host: "172.31.33.154"
```

Our first task was to tune the HTTP client knobs for more
idle connections and faster timeouts. But essentially if the kernel
can't handle the load, there's not much use. We set 3 specific knobs
which are essential for the kernel to be able to serve high concurrency.
The defaults are tuned down and it is recommended to bump them during load.

```
net.ipv4.tcp_tw_reuse = 1
net.core.somaxconn = 4096
net.ipv4.tcp_max_syn_backlog = 8192
```

From the kernel documentation:

tcp_tw_reuse - INTEGER
>	Enable reuse of TIME-WAIT sockets for new connections when it is
	safe from protocol viewpoint.

This would help us recycle connections at a much faster rate.

somaxconn - INTEGER
>	Limit of socket listen() backlog, known in userspace as SOMAXCONN.
	Defaults to 4096. (Was 128 before linux-5.4)
	See also tcp_max_syn_backlog for additional tuning for TCP sockets.

Ubuntu 18.04 still uses kernel 4.15, therefore it was 128. We bump it
to 4096 which is the default in newer kernels.

tcp_max_syn_backlog - INTEGER
>	Maximal number of remembered connection requests (SYN_RECV),
	which have not received an acknowledgment from connecting client.
	This is a per-listener limit.

We set this too because `dmesg` showed warnings like this:

```
172948.627905] TCP: request_sock_TCP: Possible SYN flooding on port 80. Sending cookies.  Check SNMP counters.
```

A deeper investigation in the kernel documentation shed some
light on the matter:

	Note, that syncookies is fallback facility.
	It MUST NOT be used to help highly loaded servers to stand
	against legal connection rate. If you see SYN flood warnings
	in your logs, but investigation	shows that they occur
	because of overload with legal connections, you should tune
	another parameters until this warning disappear.
	See: tcp_max_syn_backlog, tcp_synack_retries, tcp_abort_on_overflow.

After all this, there are no more connecting errors on the client.

We also bump up the number of upstream nginx keepalive connections per core.
And make timeouts to saner values.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-21660

### DIFF
--- a/deployment/terraform/agent.go
+++ b/deployment/terraform/agent.go
@@ -86,7 +86,7 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent, output *Output
 
 		batch := []uploadInfo{
 			{srcData: strings.TrimPrefix(buf.String(), "\n"), dstPath: "/lib/systemd/system/ltagent.service", msg: "Uploading agent service file"},
-			{srcData: strings.TrimPrefix(sysctlConfig, "\n"), dstPath: "/etc/sysctl.conf"},
+			{srcData: strings.TrimPrefix(clientSysctlConfig, "\n"), dstPath: "/etc/sysctl.conf"},
 			{srcData: strings.TrimPrefix(limitsConfig, "\n"), dstPath: "/etc/security/limits.conf"},
 			{srcData: strings.TrimPrefix(coordinatorServiceFile, "\n"), dstPath: "/lib/systemd/system/ltcoordinator.service", msg: "Uploading coordinator service file"},
 		}
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -165,11 +165,13 @@ func (t *Terraform) setupAppServers(output *Output, extAgent *ssh.ExtAgent, uplo
 				return
 			}
 
-			// Upload service file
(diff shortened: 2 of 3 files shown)


agnivade:
Please run a "deployment create" once :P I didn't do that.

agnivade:
Tested and works fine.

PR #256, opened 2020-04-26
MM-24546: Bump up connection limits to DB

agnivade:
The default limits are too low for load testing. Bumping them alone brings down the response time by 1/4th.

responsetimes

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24546


streamer45:
Should these value be somehow calculated from coordinator's config `MaxActiveUser`?  
Meaning, should they scale based on how many users we are planning to simulate or should they fit most cases?

agnivade:
I think they should actually scale with the instance type. The max conn property is more of a limit of how much the instance can support and how much the DB can take. The DB is calculated automatically based on instance type. This should also be the same.

We can theoretically reduce the limit if the maxactiveuser is low, but I don't see a direct correlation. If maxactive user is low, the number of concurrent queries will anyways be low. And if it's high, it will anyways be gated by this limit.

PR #261, opened 2020-04-29
Fix nginx map block

agnivade:


agnivade:
Tested and verified this time :)

PR #264, opened 2020-05-02
[MM-24756] Fix ws client leak and handle timeout event

streamer45:
#### Summary

PR fixes a ws client leak in `userentity`. In case the ws event channel was getting closed we would exit the inner loop and re-create the client without explicitly closing the previous one which meant leaking the internal writer goroutine.
PR also adds a case to handle the timeout event.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24756

### DIFF
--- a/loadtest/user/userentity/websocket.go
+++ b/loadtest/user/userentity/websocket.go
@@ -127,7 +127,8 @@ func (ue *UserEntity) listen(errChan chan error) {
 		}
 
 		client.Listen()
-		chanClosed := false
+		var chanClosed bool
+		var timedOut bool
 		for {
 			select {
 			case ev, ok := <-client.EventChannel:
@@ -142,7 +143,6 @@ func (ue *UserEntity) listen(errChan chan error) {
 			case _, ok := <-client.ResponseChannel:
 				if !ok {
 					chanClosed = true
-					break
 				}
 			case <-ue.wsClosing:
 				client.Close()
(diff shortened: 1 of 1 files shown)


loadtest/user/userentity/websocket.go
```diff
@@ -127,7 +127,8 @@ func (ue *UserEntity) listen(errChan chan error) {
 		}
 
 		client.Listen()
-		chanClosed := false
+		var chanClosed bool
+		var timedOut bool
```
streamer45, line 131: This is pretty explicit. We could replace it with a single `var exitLoop bool` if you think the double variable is a bit excessive.

loadtest/user/userentity/websocket.go
```diff
@@ -155,8 +155,11 @@ func (ue *UserEntity) listen(errChan chan error) {
 					break
 				}
 				client.UserTyping(msg.channelId, msg.parentId)
+			case <-client.PingTimeoutChannel:
+				timedOut = true
```
streamer45, line 159: Maybe we could log something here.

loadtest/user/userentity/websocket.go
agnivade: Seems fine to me.

loadtest/user/userentity/websocket.go
agnivade: 0/5. Upto you.

streamer45:
> Looks like I introduced all the bugs. Maybe I should consider a different job

It's the law of large numbers. It just means you do a lot of work :+1:

PR #265, opened 2020-05-04
[MM-24758] Fix possible deadlock in memstore

streamer45:
#### Summary

PR fixes a possible deadlock in `memstore.RandomUsers()`.
It's practically the same problem we recently fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/254.

We assumed the current user would not be in the map of users hence allowed for picking up to `len(s.users)` which meant that in case of `len == 2` and one of them was the current user `Randomuser()` would always the return the same user and the loop in `RandomUsers()` would spin indefinitely (with the read lock acquired) trying to get a different one.

PR also improves the logic a bit by checking whether the current user is in the store and if so decreasing by one the number of *selectable* users.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24758

### DIFF
--- a/loadtest/store/memstore/random.go
+++ b/loadtest/store/memstore/random.go
@@ -22,6 +22,7 @@ var (
 	ErrChannelStoreEmpty = errors.New("memstore: channel store is empty")
 	ErrChannelNotFound   = errors.New("memstore: channel not found")
 	ErrPostNotFound      = errors.New("memstore: post not found")
+	ErrInvalidData       = errors.New("memstore: invalid data found")
 )
 
 func isSelectionType(st, t store.SelectionType) bool {
@@ -139,7 +140,11 @@ func (s *MemStore) RandomUser() (model.User, error) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 
-	if len(s.users) < 2 {
+	minLen := 1
+	if _, ok := s.users[s.user.Id]; ok {
+		minLen++
+	}
+	if len(s.users) < minLen {
 		return model.User{}, ErrLenMismatch
 	}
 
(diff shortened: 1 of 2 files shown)


agnivade:
> which meant that in case of len == 2 and one of them was the current user Randomuser() would always the return the same user

I am wondering why that happens. Is it because we are using math/rand and not crypto/rand ?

streamer45:
> > which meant that in case of len == 2 and one of them was the current user Randomuser() would always the return the same user
> 
> I am wondering why that happens. Is it because we are using math/rand and not crypto/rand ?

That's just how the logic is. `RandomUser()` will never return the current user so if you call it when there are a only 2 users in the store and one of them is the current one you are going to get the other. The problem is that `RandomUsers()` is not going to stop calling `RandomUser()` until it picked `n` users which is never going to happen if `n == 2`. We want to pick 2 out of 2 but one of those is the current one.

loadtest/store/memstore/random.go
```diff
@@ -139,7 +140,11 @@ func (s *MemStore) RandomUser() (model.User, error) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 
-	if len(s.users) < 2 {
+	minLen := 1
```
agnivade, line 143: A note here would be useful to remind us why are we doing this, given that we stumbled on this twice.

loadtest/store/memstore/random.go
```diff
@@ -162,16 +170,21 @@ func (s *MemStore) RandomUsers(n int) ([]model.User, error) {
 	s.lock.RLock()
 	defer s.lock.RUnlock()
 
-	if n > len(s.users) {
+	numUsers := len(s.users)
```
agnivade, line 173: Let's add the comment you mentioned here: https://github.com/mattermost/mattermost-load-test-ng/pull/265#issuecomment-623419534. This is a very important point.

PR #274, opened 2020-05-07
MM-24930: Pass along controller config too

agnivade:
During starting a coordinator, we were not passing the controller
config files at all. This meant that the coordinator silently used
the default configs which were not ideal. So during coordinator start
we pass the controller files too.

While here, we also fix some other things.
- Set the /etc/security/limits.conf file on app servers too.
- Tune the read/write/connect timeouts slightly on both app and nginx.
- Set the Readonly config to true which is the default and should be set.

### DIFF
--- a/deployment/terraform/coordinator.go
+++ b/deployment/terraform/coordinator.go
@@ -46,20 +46,6 @@ func (t *Terraform) StartCoordinator() error {
 		return err
 	}
 
-	agentConfig, err := t.generateLoadtestAgentConfig(output)
-	if err != nil {
-		return err
-	}
-	data, err := json.MarshalIndent(agentConfig, "", "  ")
-	if err != nil {
-		return err
-	}
-	dstPath := "/home/ubuntu/mattermost-load-test-ng/config/config.json"
-	mlog.Info("Uploading updated load-test agent config file")
-	if out, err := sshc.Upload(bytes.NewReader(data), dstPath, false); err != nil {
-		return fmt.Errorf("error uploading file, output: %q: %w", out, err)
-	}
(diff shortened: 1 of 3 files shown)


deployment/terraform/coordinator.go
```diff
@@ -77,6 +79,36 @@ func (t *Terraform) StartCoordinator() error {
...
+	}
+	data, err = json.MarshalIndent(simpleConfig, "", "  ")
+	if err != nil {
+		return err
+	}
+	mlog.Info("Uploading simplecontroller config file")
+	dstPath = "/home/ubuntu/mattermost-load-test-ng/config/simplecontroller.json"
+	if out, err := sshc.Upload(bytes.NewReader(data), dstPath, false); err != nil {
+		return fmt.Errorf("error running ssh command: output: %s, error: %w", out, err)
+	}
```
streamer45, line 110: Can this be compressed somehow? We do the same things, just the controller type changes.

agnivade: Sure, let me take look.

agnivade: better now ?

streamer45: Yes, wondering why you didn't use`uploadBatch()` though.

agnivade: Because I needed to encode the JSON too. So it would be two loops. 0/5.

streamer45: Fair enough :+1: 
We can probably pass `bytes.NewReader(data)` directly since it's pretty clear what it returns.

PR #277, opened 2020-05-09
[MM-25008] Implement collect command

streamer45:
#### Summary

PR implements a new `ltctl collect` command to help collecting files (logs and configurations) needed to aid debugging.

#### Ticket

https://mattermost.atlassian.net/browse/MM-25008

### DIFF
--- a/deployment/terraform/collect.go
+++ b/deployment/terraform/collect.go
@@ -1,0 +1,239 @@
+package terraform
+
+import (
+	"archive/tar"
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"os/user"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
+)
+
+type collectInfo struct {
+	instance string
+	src      string
+	compress bool
+}
+
+type file struct {
+	name string
+	data []byte
+}
+
+func saveCollection(files []file) error {
+	u, err := user.Current()
+	if err != nil {
+		return err
+	}
+
+	uid, err := strconv.ParseInt(u.Uid, 10, 32)
+	if err != nil {
+		return err
+	}
+
+	gid, err := strconv.ParseInt(u.Gid, 10, 32)
+	if err != nil {
+		return err
+	}
+
+	name := fmt.Sprintf("collection_%d", time.Now().Unix())
+	filename := fmt.Sprintf("./%s.tar", name)
+	f, err := os.Create(filename)
+	if err != nil {
+		return fmt.Errorf("failed to create file %w", err)
+	}
+	defer f.Close()
+
+	tw := tar.NewWriter(f)
+	hdr := &tar.Header{
+		Name:    name + "/",
+		Mode:    0755,
+		Uid:     int(uid),
+		Gid:     int(gid),
+		ModTime: time.Now(),
+	}
+	if err := tw.WriteHeader(hdr); err != nil {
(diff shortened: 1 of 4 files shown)


deployment/terraform/ssh/ssh.go
```diff
@@ -127,6 +127,33 @@ func (sshc *Client) UploadFile(src, dst string, sudo bool) ([]byte, error) {
...
+	}
+
+	sess, err := sshc.client.NewSession()
+	if err != nil {
+		return err
+	}
+	defer sess.Close()
+	sess.Stdout = dst
+
+	cmd := fmt.Sprintf("cat '%s'", src)
```
agnivade, line 145: This cat business seriously needs to go away. Maybe our first help wanted ticket ?

deployment/terraform/collect.go
```diff
@@ -0,0 +1,239 @@
...
+		sshc, err := extAgent.NewClient(agent.PublicIP)
+		if err != nil {
+			return fmt.Errorf("error in getting ssh connection %w", err)
+		}
+		clients[fmt.Sprintf("agent%d", i)] = sshc
+		if i == 0 {
+			clients["coordinator"] = sshc
+		}
+	}
+
```
agnivade, line 140: Overall looks great. But I have a slight concern, which you would have probably seen it coming :)

I think we are adding too much functionality into the `terraform` package. The part after this line (L140) is not related to terraform at all, and can be in its own separate package (we can call it debug, 0/5 on it). Same for the `collection` function.

All that is needed from the terraform package is the `clients` map which can be done in the cmd/ltctl package itself IMO. It's just some wrapper code over the output and creating ssh clients. And then it can pass it to the `debug` package and let all of this be there. In future, this can also be reused by other deployment environments.

This is kinda what we did with the ssh package too. I think initially it was inside terraform and then moved to its own subpackage. I am okay if you want to do it later, but just wanted to open it up for discussion.

Thoughts ?

deployment/terraform/collect.go
```diff
@@ -0,0 +1,239 @@
...
+	wg.Wait()
+
+	numFiles := len(filesChan)
+	if numFiles == 0 {
+		return errors.New("failed to collect any file")
+	}
+
+	files := make([]file, numFiles)
```
agnivade, line 229: mmm .. what if some files succeed and some fail. Then won't you be waiting forever ? Perhaps it might be better to just range over the channel in a separate goroutine and close it after `wg.Wait` is done to let it auto-exit. Then I believe we don't even need it to be buffered.

streamer45: If some files fail to get collected they won't get sent into the channel. We only loop on the actual content (`numFiles`) that we find in the channel.

deployment/terraform/collect.go
streamer45: That's all true. The reasoning was that the `ssh` package is actually used in many places while this is a standalone thing. Maybe for now I could move all non terraform functionality directly into a `cmd/ltctl/collect.go`.

deployment/terraform/ssh/ssh.go
streamer45: Yes. Ideally we should use scp/sftp but didn't feel like adding the dependency for a debug utility.

deployment/terraform/collect.go
agnivade: That sounds good to me. :+1:

deployment/terraform/collect.go
agnivade: Ah missed that. Thanks.

PR #278, opened 2020-05-09
[MM-24688] Refine SimulController

streamer45:
#### Summary

PR further refines actions frequencies for `SimulController` to better match those of a real user.
Data from our community servers has been used to perform the tuning.
Other notable changes included are:

- Tuned average idle time.
- Improved fetching users statuses.
- Made clients use a shared `http.Transport`. This should help improving overall connectivity performance.
- Fixed a bug with file uploads. Not sure if it was related with the errors we've been seeing server side but it's a possibility.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24688

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -6,9 +6,11 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"net/http"
 	"net/http/pprof"
 	"strconv"
+	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
@@ -102,6 +104,22 @@ func (a *API) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// http.Transport to be shared amongst all clients.
+	transport := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   1 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		MaxConnsPerHost:       500,
+		MaxIdleConns:          500,
+		MaxIdleConnsPerHost:   500,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   1 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
+
 	newControllerFn := func(id int, status chan<- control.UserStatus) (control.UserController, error) {
 		ueConfig := userentity.Config{
 			ServerURL:    ltConfig.ConnectionConfiguration.ServerURL,
@@ -119,7 +137,7 @@ func (a *API) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		if err != nil {
 			return nil, err
 		}
-		ue := userentity.New(store, ueConfig)
+		ue := userentity.New(store, transport, ueConfig)
 		switch ltConfig.UserControllerConfiguration.Type {
 		case loadtest.UserControllerSimple:
 			return simplecontroller.New(id, ue, ucConfig.(*simplecontroller.Config), status)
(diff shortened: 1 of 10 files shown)


api/server.go
```diff
@@ -102,6 +104,22 @@ func (a *API) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
...
+	transport := &http.Transport{
+		Proxy: http.ProxyFromEnvironment,
+		DialContext: (&net.Dialer{
+			Timeout:   1 * time.Second,
+			KeepAlive: 30 * time.Second,
+			DualStack: true,
+		}).DialContext,
+		MaxConnsPerHost:       500,
+		MaxIdleConns:          500,
+		MaxIdleConnsPerHost:   500,
```
streamer45, line 117: These are not completely arbitrary values. I've ran several iterations and given the default instance for a load-test agent (`t3.xlarge`) these are the values that result in the highest average throughput (requests per second) running `1500` users which was found to be more or less the maximum supported by such instance.

I'm still working on a way to calculate these values from the `MaxActiveUsers` but didn't get to a magic formula yet so for now they'll have be a hard-coded (but tested) default.

agnivade: Sharing the client is definitely great. And setting MaxConns is also awesome. I am a bit surprised though that 500 turned out to be the max limit. The linux kernel should be certainly capable of handling more than 500 connections. What sort of bottlenecks did you notice when you increase the maxconns to more than 500 ?

streamer45: I am sure there are many details affecting that, not last how the app is designed. Let's not forget those connections will be on top of `MaxActiveUsers` websocket connections receiving a fairly load of events. Also those limits only apply to the specific instance type with those configurations. 

Ideally I would have liked at least one HTTP connection per user (which is achievable but then requires lowering the number of max users with no apparent benefit). 
In the end I think it's a trade-off between high concurrency and throughput and I am probably more in favor of the latter since what we really want to stress is the application logic and not so much the connectivity part.
With that said, the main problem was a huge spike in system CPU time:

lt

These are three successive load-tests, first one with connections limited to 500 and the following ones higher values, 1000, and 1500.

I've also tested a couple of sysctl parameters with no noticeable improvement. Apparently most of them are meant for servers accepting connections and not so much for outgoing connections.
We can obviously iterate more on that front. For now I can just present what I found to be a stable enough setting.

agnivade: > Ideally I would have liked at least one HTTP connection per user 

What if we do `max(maxactiveusers, 500)` ?

> the main problem was a huge spike in system CPU time:

Interesting. Would you happen to have CPU and trace profiles for those 3 tests by any chance ?

streamer45: > What if we do `max(maxactiveusers, 500)` ?
> 
> EDIT: This is assuming that the number is appropriately set for the instance type.

That would mean lowering quite a bit the number of supported users for not much gain as far as I can see. Might need more data to confirm though.

> Interesting. Would you happen to have CPU and trace profiles for those 3 tests by any chance ?

I've already cleanup up those unfortunately but I collected new ones for the three scenarios (500/1000/1500):

CPU profiles:

[500](https://gist.github.com/streamer45/72a3b65190bd62231a28fbe6219f54a5/raw/b48a51c200170166ad587cb99cf1cc447f3351e7/profile_500) [1000](https://gist.github.com/streamer45/72a3b65190bd62231a28fbe6219f54a5/raw/b48a51c200170166ad587cb99cf1cc447f3351e7/profile_1000) [1500](https://gist.github.com/streamer45/72a3b65190bd62231a28fbe6219f54a5/raw/b48a51c200170166ad587cb99cf1cc447f3351e7/profile_1500)

Trace profiles:

[500](https://gist.github.com/streamer45/72a3b65190bd62231a28fbe6219f54a5/raw/b48a51c200170166ad587cb99cf1cc447f3351e7/trace_500) [1000](https://gist.github.com/streamer45/72a3b65190bd62231a28fbe6219f54a5/raw/b48a51c200170166ad587cb99cf1cc447f3351e7/trace_1000) [1500](https://gist.github.com/streamer45/72a3b65190bd62231a28fbe6219f54a5/raw/b48a51c200170166ad587cb99cf1cc447f3351e7/trace_1500)

agnivade: > That would mean lowering quite a bit the number of supported users for not much gain as far as I can see.

Indeed that's true. Thanks for the profiles ! I will go through them and see if there's anything interesting I can find.

PR #281, opened 2020-05-12
[MM-24880] Review default Grafana dashboard

streamer45:
#### Summary

PR reviews the default Grafana dashboard, fixing some labels and units to help keep consistency.
It also adds panels for some node-exporter metrics:

- CPU Load
- Memory Usage
- Network TX/RX Rates
- TCP Connections

#### Ticket

https://mattermost.atlassian.net/browse/MM-24880


deployment/terraform/assets/dashboard_data.json
```diff
@@ -657,18 +657,15 @@
           "format": "time_series",
           "intervalFactor": 1,
           "legendFormat": "{{instance}}",
-          "metric": "mattermost_api_time_sum",
+          "metric": "mattermost_http_request_duration_seconds_sum",
```
agnivade, line 660: This should be api_time_sum

deployment/terraform/assets/dashboard_data.json
```diff
@@ -342,16 +342,16 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "increase(mattermost_http_requests_total{instance=~\"$server\"}[1m])",
+          "expr": "rate(mattermost_http_requests_total{instance=~\"$server\"}[1m])",
```
agnivade, line 345: From what I understand, this should be "increase". Or was I wrong the first time ? I thought you wanted this changed @ethervoid ? I keep forgetting the reason.

ethervoid: Yes. The increase function is syntactic sugar and should be used in panels. Rates are usually mean for alerts or slow-moving counters as you can see in [their docs](https://prometheus.io/docs/prometheus/latest/querying/functions/#rate):

```
rate should only be used with counters. It is best suited for alerting, and for graphing of slow-moving counters.
```

In case you use `rate` your Y-axis is going to reflect the increase ration instead of the real measurement (requests, messages, etc). From their docs:

```
 It is syntactic sugar for rate(v) multiplied by the number of seconds under the specified time range window, and should be used primarily for human readability
```

streamer45: I am fine with using `increase` but then we need to divide by 60 to get the requests per second.

deployment/terraform/assets/dashboard_data.json
streamer45: Not sure how that gets updated from Grafana. The query uses the correct metric.

deployment/terraform/assets/dashboard_data.json
ethervoid: Yep given that the number returned is based on the time range you're using. But is better than using rate IMHO

deployment/terraform/assets/dashboard_data.json
agnivade: What do you mean "updated from Grafana" ? I meant that both api_time_sum and http_request_duration_seconds sum capture the same thing. I have a PR out which removes the http_request_duration metric from Prometheus. We should just be using api_time_sum.

streamer45: I mean, the query uses the right metric but that's not reflected in that JSON field. I don't know where I can set that from Grafana UI.

ethervoid: That's because that metric [is going to be removed](https://github.com/mattermost/mattermost-server/pull/14520)

deployment/terraform/assets/dashboard_data.json
agnivade: Let's use rate. It does not make much sense to use increase and again divide by 60. And this is why the RPS numbers were wrong. We weren't really pushing 200k RPS then. Which makes sense now.

deployment/terraform/assets/dashboard_data.json
streamer45: Updated anyway :P

deployment/terraform/assets/dashboard_data.json
streamer45: Yeah, not sure what's the difference if indeed `increase` is just another `form` of doing `rate`. It's not like it's more precise is it?

agnivade: So .. just to clarify is rate different than increase/60 ?

streamer45: I believe it's the same if you are applying to [1m] which we are.

agnivade: Well, okay, then I don't see any reason _not_ to use rate.

ethervoid: After investigating with Claudio looks like using rate is the way to go in this case (RPS) because using rate you get the per-second value of that instant point without aggregation

Sorry for the confusion

streamer45:
Made a few more fixes and added two panels for Duration of API and Store calls.

deployment/terraform/assets/dashboard_data.json
```diff
@@ -60,7 +60,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum(increase(mattermost_db_store_time_count{method=~\"$top_db_count\"}[1m])) by (method)",
+          "expr": "sum(increase(mattermost_db_store_time_count{method=~\"$top_db_count\"}[$__range])) by (method)",
```
ethervoid, line 63: If this is a panel instead a table or counter I think is better to use the `$__interval` variable instead of range because interval takes into account the resolution of the graph

streamer45: But they are not equivalent. That graph should show how many API calls we made in the defined time range which value is `$__range`. I am not sure how using `$__interval` there would help getting the correct results.

ethervoid: :+1:

PR #285, opened 2020-05-17
MM-25292: Fix idempotency regressions

agnivade:
We stop the service first and then copy the binary.
For load agents too, we restart it rather than starting it.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25292


deployment/terraform/create.go
```diff
@@ -179,6 +179,12 @@ func (t *Terraform) setupAppServers(output *Output, extAgent *ssh.ExtAgent, uplo
 			// Upload binary if needed.
 			if uploadBinary {
 				mlog.Info("Uploading binary", mlog.String("host", ip))
+				cmd := "sudo service mattermost stop || true"
```
streamer45, line 182: Why do we need the `|| true`? I thought the stop command would succeed whether or not the service is running.

agnivade: It will fail the first time, because the service isn't even installed.

streamer45: The service file has been uploaded at that point. Maybe we just need a `systemctl daemon-reload`?

agnivade: Yes, should be fine.

PR #286, opened 2020-05-18
[MM-25047] Distribute controllers by action rate

streamer45:
#### Summary

PR implements a simple distribution for controllers based on their action rate.
Instead of using a fixed rate value for all the controllers we can now define which percentage of them should run with a given rate.

#### Ticket

https://mattermost.atlassian.net/browse/MM-25047


loadtest/config.go
```diff
@@ -69,12 +70,15 @@ type UserControllerConfiguration struct {
...
-	// A rate multiplier that will affect the speed at which user actions are
+	// A distribution of rate multipliers that will affect the speed at which user actions are
 	// executed by the UserController.
 	// A Rate of < 1.0 will run actions at a faster pace.
 	// A Rate of 1.0 will run actions at the default pace.
 	// A Rate > 1.0 will run actions at a slower pace.
-	Rate float64
+	RatesDistribution []struct {
+		Rate       float64
+		Percentage float32
```
agnivade, line 80: Any reason to keep this float32 and the other as float64 ? I don't mind either ways. Just curious.

streamer45: I've noticed that. We can probably move everything to float32 since we clearly don't need high precision.

streamer45: Apparently `float64` is pretty much the default as far as Golang is concerned so might aswell move everything to it. Using `float32` would fill us of `float64()` casting with no real added benefit.

loadtest/config.go
```diff
@@ -69,12 +70,15 @@ type UserControllerConfiguration struct {
...
+	// A distribution of rate multipliers that will affect the speed at which user actions are
 	// executed by the UserController.
 	// A Rate of < 1.0 will run actions at a faster pace.
 	// A Rate of 1.0 will run actions at the default pace.
 	// A Rate > 1.0 will run actions at a slower pace.
-	Rate float64
+	RatesDistribution []struct {
+		Rate       float64
+		Percentage float64
+	}
```
streamer45, line 81: @isacikgoz  I hope this anon struct won't cause too much troubles with https://github.com/mattermost/mattermost-load-test-ng/pull/257. If so, let me know :)

PR #289, opened 2020-05-19
MM-24970: Fix double-locking of the store

agnivade:
We were using the read lock recursively which is not allowed.
To fix this, we use a separate internal method.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-24970


streamer45:
I believe `memstore.ChannelPostsSorted()` could be suffering from the same issue. Should be easy to fix it.

agnivade:
Oh yes indeed !

PR #293, opened 2020-05-20
[MM-24613] Update some defaults

streamer45:
#### Summary

PR sets the default controller to `simulative` and also changes the default for max active users.
These values come from several tests and they apply to default configs.  Currently I've verified the following:

```golang
map[string]int{
  "t3.xlarge": 2000,
  "m5.xlarge": 2500,
  "m5.2xlarge": 4500,
}
```

These are the maximum number of user to be ran on a single agent instance of that type.
When I have gathered more data I'll also include this info in the documentation.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24613

### DIFF
--- a/config/config.default.json
+++ b/config/config.default.json
@@ -6,7 +6,7 @@
     "AdminPassword": "Sys@dmin-sample1"
   },
   "UserControllerConfiguration": {
-    "Type": "simple",
+    "Type": "simulative",
     "RatesDistribution":[
       {
         "Rate": 1.0,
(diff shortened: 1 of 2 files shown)


streamer45:
Just want to wait for a few more tests before merging. Also there are a couple of issues that I'd like fixed before settling on these numbers.

streamer45:
After recent tests I would say these are slightly conservative numbers but close enough to be valid defaults.

PR #294, opened 2020-05-21
MM-25049: Add cluster controller

agnivade:
Just a basic smoke testing of the APIs that involve cluster communication.
It just iterates over all APIs serially, rolling over to the first API.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25049

### DIFF
--- a/cmd/ltagent/loadtest.go
+++ b/cmd/ltagent/loadtest.go
@@ -19,7 +19,6 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/user/userentity"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
-	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/spf13/cobra"
 )
 
@@ -105,15 +104,8 @@ func RunLoadTestCmdF(cmd *cobra.Command, args []string) error {
 			// For cluster controller, we only use the sysadmin
 			// because we are just testing system console APIs.
 			ueConfig.Username = ""
-			ueConfig.Email = ""
-			ueConfig.Password = ""
-			err = store.SetUser(&model.User{
-				Email:    config.ConnectionConfiguration.AdminEmail,
-				Password: config.ConnectionConfiguration.AdminPassword,
-			})
-			if err != nil {
-				return nil, err
-			}
+			ueConfig.Email = config.ConnectionConfiguration.AdminEmail
+			ueConfig.Password = config.ConnectionConfiguration.AdminPassword
 
 			admin := userentity.New(store, transport, ueConfig)
 			return clustercontroller.New(id, admin, status)
--- a/cmd/ltagent/loadtest.go
+++ b/cmd/ltagent/loadtest.go
@@ -11,12 +11,15 @@ import (
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/clustercontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/noopcontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/simplecontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/simulcontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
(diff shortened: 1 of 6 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -881,3 +884,71 @@ func (ue *UserEntity) SetCurrentChannel(channel *model.Channel) error {
...
+}
+
+func (ue *UserEntity) getPluginStatuses() (model.PluginStatuses, *model.Response) {
+	r, err := ue.client.DoApiGet(ue.client.GetPluginsRoute()+"/statuses", "")
+	if err != nil {
+		return nil, model.BuildErrorResponse(r, err)
+	}
+	defer func() {
+		_, _ = io.Copy(ioutil.Discard, r.Body)
+		_ = r.Body.Close()
```
isacikgoz, line 951: I see you are a man of culture as well :)

loadtest/user/userentity/actions.go
```diff
@@ -881,3 +884,71 @@ func (ue *UserEntity) SetCurrentChannel(channel *model.Channel) error {
...
+// GetAnalytics fetches the system analytics.
+func (ue *UserEntity) GetAnalytics() error {
+	teamIds, err := ue.GetTeamsForUser(ue.store.Id())
+	if err != nil {
+		return err
+	}
+	if len(teamIds) == 0 {
+		return fmt.Errorf("user does not have any teams")
+	}
+	_, resp := ue.client.GetAnalyticsOld("", teamIds[0])
```
isacikgoz, line 906: Is there any reason to pick first team every time?

loadtest/user/user.go
```diff
@@ -130,6 +131,18 @@ type User interface {
 	SetCurrentTeam(team *model.Team) error
 	SetCurrentChannel(channel *model.Channel) error
 
+	// System console functionalities
```
isacikgoz, line 134:
```suggestion
	// System console functionalities
	
```

loadtest/user/userentity/actions.go
agnivade: Just keeping it simple. This is not a load test in any sense. Just a basic sanity test of the cluster APIs.

loadtest/user/userentity/actions.go
```diff
@@ -881,3 +884,71 @@ func (ue *UserEntity) SetCurrentChannel(channel *model.Channel) error {
...
+}
+
+// GetAnalytics fetches the system analytics.
+func (ue *UserEntity) GetAnalytics() error {
+	teamIds, err := ue.GetTeamsForUser(ue.store.Id())
+	if err != nil {
+		return err
+	}
+	if len(teamIds) == 0 {
+		return fmt.Errorf("user does not have any teams")
```
streamer45, line 904: We agreed on using `errors.New()` for simple strings :grin:

loadtest/user/userentity/actions.go
streamer45, line 906: I still think this simplification should be made at the controller level by making `GetAnalytics()` accept a `teamId`. Alternatively just omit the team parameter which should make it return analytics for all the teams since the goal is just making the call.

loadtest/user/userentity/actions.go
agnivade: Yep. This should have been caught by golangci-lint actually. Hmm..

cmd/ltagent/loadtest.go
```diff
@@ -98,6 +101,22 @@ func RunLoadTestCmdF(cmd *cobra.Command, args []string) error {
...
+			ueConfig.Username = ""
+			ueConfig.Email = ""
+			ueConfig.Password = ""
+			err = store.SetUser(&model.User{
+				Email:    config.ConnectionConfiguration.AdminEmail,
+				Password: config.ConnectionConfiguration.AdminPassword,
+			})
+			if err != nil {
+				return nil, err
+			}
```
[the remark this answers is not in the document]
agnivade: This is how we are doing it in ltagent/init.go. I was just trying to be consistent. 0/5 on either ways.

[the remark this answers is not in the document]
streamer45: Well, in `init` we are somewhat abusing our own APIs. We could have directly used a `model.Client4` to do all that but went with the `UserEntity` for convenience. 
Also, the current init process can be probably replaced in the future by running `GenController`.

Here we are actually setting up a controller to be later run so I think it makes more sense to follow what we do with the others, setting up the entity and let it deal with the store.

PR #296, opened 2020-05-25
Add go.tools.mod to track tool dependencies

agnivade:


Makefile
```diff
@@ -17,6 +17,9 @@ COORDINATOR_ARGS=-mod=readonly -trimpath ./cmd/ltcoordinator
 AGENT=ltagent
 AGENT_ARGS=-mod=readonly -trimpath ./cmd/ltagent
 
+export GOBIN = $(PWD)/bin
+export PATH = $(shell printenv PATH):$(GOBIN)
```
streamer45, line 21: Is there a reason we are not doing just:

```sh
PATH=$PATH:$GOBIN
```

Makefile
```diff
@@ -17,6 +17,9 @@ COORDINATOR_ARGS=-mod=readonly -trimpath ./cmd/ltcoordinator
 AGENT=ltagent
 AGENT_ARGS=-mod=readonly -trimpath ./cmd/ltagent
 
+export GOBIN = $(PWD)/bin
```
streamer45, line 20: I somehow prefer the var assignment without spaces, mainly because I've been hit by this a few times in the past.

Makefile
agnivade: Yes, we need to do "shell printenv" because otherwise this happens- Recursive variable `PATH' references itself (eventually). .

And I needed to do export because without that it's just a simple variable set, and sub-processes spawned from the Makefile command do not get them. But somehow they do now. Will change.

PR #300, opened 2020-05-27
MM-25117: add inbucket instance to the cluster

agnivade:
We reuse the metrics server to run our inbucket instance.


agnivade:
Yeah, I didn't feel the need to spin up yet another instance for it. It's a small service and can fit in fine with Prom and Grafana. We can split it out later if needed.

PR #303, opened 2020-05-28
MM-25307: Add s3 support for clustered deployments

agnivade:
We add an s3 filestore support using Terraform when
the user creates a clustered setup.

This is to have a shared filestore to prevent objects
from not being found.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-25307


streamer45:
@agnivade Tests are failing on this one.

agnivade:
Woops, didn't see it.

PR #308, opened 2020-06-02
Include Elasticsearch in the load tester cluster

ethervoid:
#### Summary

This PR adds Elasticsearch as an optional part for the load tester. It will include:

- Elasticsearch instance configured
- Elasticsearch Prometheus metrics exporter
- New Elasticsearch dasboard

Screenshot from 2020-06-02 21-31-02

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25269


config/deployer.default.json
```diff
@@ -1,9 +1,17 @@
 {
   "ClusterName": "loadtest",
+  "VpcId": "",
```
ethervoid, line 3: We can probably use this id and put all the machines in the same VPC

deployment/config.go
```diff
@@ -48,6 +50,14 @@ type Config struct {
 	LogSettings         logger.Settings
 }
 
+type ElasticSearchSettings struct {
```
agnivade, line 53: godocs please :)

deployment/config.go
```diff
@@ -48,6 +50,14 @@ type Config struct {
...
+	Enable       bool   // Elasticsearch is going to be created
```
agnivade, line 54: This is a very small thing. So feel free to ignore. But all comments here end with a period. So might be good to keep that consistency.

deployment/terraform/create.go
```diff
@@ -338,6 +344,16 @@ func (t *Terraform) updateAppConfig(ip string, sshc *ssh.Client, output *Output)
 	cfg.PluginSettings.Enable = model.NewBool(true)
 	cfg.PluginSettings.EnableUploads = model.NewBool(true)
 
+	if t.config.HasElasticSearch() {
```
agnivade, line 347: We usually check the output struct for this. It helps confirm that Value[0] is actually there. We can add a similar method to the output struct also like `HasProxy`.

deployment/terraform/destroy.go
```diff
@@ -20,6 +20,7 @@ func (t *Terraform) Destroy() error {
 		"-var", fmt.Sprintf("app_instance_type=%s", t.config.AppInstanceType),
 		"-var", fmt.Sprintf("agent_instance_count=%d", t.config.AgentInstanceCount),
 		"-var", fmt.Sprintf("agent_instance_type=%s", t.config.AgentInstanceType),
+		"-var", fmt.Sprintf("es_instance=%t", t.config.HasElasticSearch()),
```
agnivade, line 23: I think you also need to pass vpc id and everything else (after you remove the default values). This is what I got when I tried to destroy

```
020-06-05T00:04:01.948+0530	info	terraform/engine.go:81	
2020-06-05T00:04:01.986+0530	info	terraform/engine.go:81	var.vpc
2020-06-05T00:04:01.986+0530	info	terraform/engine.go:81	  Enter a value: 
```

While it did go ahead and destroyed everything, it would be good to avoid this.

deployment/config.go
```diff
@@ -77,9 +87,17 @@ func (c *Config) IsValid() error {
 		return fmt.Errorf("load-test package file must be a tar.gz file")
 	}
 
+	if c.HasElasticSearch() && c.VpcID == "" {
```
agnivade, line 90: I also got this

```
Error: "domain_name" must start with a lowercase alphabet and be at least 3 and no more than 28 characters long. Valid characters are a-z (lowercase letters), 0-9, and - (hyphen)
```

when I gave "agnivaES" as the name. Might be worth improving the check on L80 to incorporate the length limit and lowercase restriction. 0/5 on whether you want to do it now or in a separate PR.

deployment/terraform/output.go
```diff
@@ -50,6 +50,14 @@ type Output struct {
 			PrivateDNS string `json:"private_dns"`
 		} `json:"value"`
 	} `json:"metricsServer"`
+	ElasticServer struct {
```
agnivade, line 53: Please also show the endpoint info at the end, as part of `(t *Terraform) displayInfo`

deployment/terraform/destroy.go
agnivade: Ah wait, will it destroy the entire VPC if I pass the vpc id ? Hmm .. we might need to create a VPC from scratch then. Something for later I guess.

agnivade: Actually, it shouldn't. Because we aren't creating the VPC. Just creating stuff inside it. So should be fine.

deployment/config.go
```diff
@@ -18,19 +18,21 @@ import (
...
-	DBInstanceEngine    string // Type of the DB instance - postgres or mysql.
-	DBUserName          string // Username to connect to the DB.
-	DBPassword          string // Password to connect to the DB.
+	ClusterName           string // Name of the cluster.
+	VpcID                 string // Id of the VPC to be used
+	AppInstanceCount      int    // Number of application instances.
+	AppInstanceType       string // Type of the EC2 instance for app.
+	AgentInstanceCount    int    // Number of agents, first agent and coordinator will share the same instance.
+	AgentInstanceType     string // Type of the EC2 instance for agent.
+	ElasticSearchSettings ElasticSearchSettings
```
isacikgoz, line 27: Interesting approach, we were using a flat order here. I think it would be better to keep it consistent. Either group `app`, `agent`, `db` etc. or remove the ES struct here and add individual configurations here. 0/5 on the both TBH.

config/deployer.default.json
```diff
@@ -1,9 +1,17 @@
...
   "AppInstanceCount": 1,
   "AppInstanceType": "c5.xlarge",
   "AgentInstanceCount": 2,
   "AgentInstanceType": "t3.xlarge",
+  "ElasticSearchSettings": {
+    "Enable": false,
+    "InstanceType": "r4.large.elasticsearch",
+    "Version": 6.5,
+    "EBSType": "gp2",
+    "EBSSize": 10
```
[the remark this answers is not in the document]
ethervoid: Hum, II agree with moving the type inside but I'd leave the EBS size outside to avoid poking around the terraform code. What do you think?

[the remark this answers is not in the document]
agnivade: I think you can just use the method followed for app servers:

```
variable "root_block_device" {
  default = [
    {
      volume_type = "gp2"
      volume_size = 50
    },
  ]
}
```

Just declare the defaults and let terraform use that.

deployment/terraform/destroy.go
ethervoid: AFAIK the destroy command only destroys the infrastructure created by terraform so passing the id should be fine and it shouldn't destroy the VPC

deployment/config.go
ethervoid: Done in this PR

deployment/terraform/output.go
ethervoid: Done

deployment/config.go
ethervoid: Hum, we have the `LoggerSettings`. I think we should have the configuration grouped as we have in the `config.json`. I'd leave it as it is because all the configuration related code is based on that :\. But if @agnivade is agree with you I'll change it

config/deployer.default.json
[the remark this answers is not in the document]
ethervoid: Done

deployment/config.go
```diff
@@ -77,9 +87,17 @@ func (c *Config) IsValid() error {
...
 	}
 
+	if c.HasElasticSearch() && c.VpcID == "" {
+		return fmt.Errorf("vpc id must be included in order to include an elastic search instance")
+	}
+
 	return nil
 }
 
+func (c *Config) HasElasticSearch() bool {
```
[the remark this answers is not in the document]
ethervoid: Removed method :)

deployment/terraform/assets/variables.tf
```diff
@@ -13,6 +17,33 @@ variable "agent_instance_count" {
...
+# Elasticsearch variables
+
+variable "es_instance" {
+  type = bool
+  default = false
+}
+
+variable "es_instance_type" {
+  type = string
+  default = "r4.large.elasticsearch"
```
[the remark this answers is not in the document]
ethervoid: Removed the default values and types

deployment/terraform/create.go
```diff
@@ -80,10 +80,16 @@ func (t *Terraform) Create() error {
 
 	err = t.runCommand(nil, "apply",
 		"-var", fmt.Sprintf("cluster_name=%s", t.config.ClusterName),
+		"-var", fmt.Sprintf("vpc=%s", t.config.VpcID),
 		"-var", fmt.Sprintf("app_instance_count=%d", t.config.AppInstanceCount),
 		"-var", fmt.Sprintf("app_instance_type=%s", t.config.AppInstanceType),
 		"-var", fmt.Sprintf("agent_instance_count=%d", t.config.AgentInstanceCount),
 		"-var", fmt.Sprintf("agent_instance_type=%s", t.config.AgentInstanceType),
+		"-var", fmt.Sprintf("es_instance=%t", t.config.HasElasticSearch()),
```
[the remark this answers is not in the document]
ethervoid: Changed

deployment/terraform/assets/cluster.tf
```diff
@@ -419,6 +481,26 @@ resource "aws_security_group" "metrics" {
...
+
+  ingress {
+    from_port       = 443
+    to_port         = 443
+    protocol        = "tcp"
+    security_groups = ["${aws_security_group.app.id}", "${aws_security_group.metrics.id}"]
+  }
+
+  egress {
+    from_port   = 443
```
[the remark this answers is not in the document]
ethervoid: We only want to expose the 443, which is the one is exposing the AWS service. Is not possible to select the port

deployment/terraform/metrics.go
```diff
@@ -102,6 +128,16 @@ func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent, output *Output) error {
 		return fmt.Errorf("error while uploading dashboard_json: output: %s, error: %w", out, err)
 	}
 
+	if t.config.HasElasticSearch() {
```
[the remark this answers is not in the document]
ethervoid: Done

deployment/terraform/assets/cluster.tf
```diff
@@ -147,6 +158,57 @@ resource "aws_instance" "proxy_server" {
...
+          }
+      ]
+  }
+  CONFIG
+
+  depends_on = [
+    aws_iam_service_linked_role.es,
+  ]
+
+  count = var.es_instance ? 1 : 0
```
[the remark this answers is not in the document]
ethervoid: Changed

deployment/config.go
ethervoid: Done!

deployment/config.go
ethervoid: Done

deployment/config.go
agnivade: I am fine with grouping. Doesn't make too much of a difference either ways. Let's go with this for now.

deployment/terraform/assets/cluster.tf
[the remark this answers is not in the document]
agnivade: Isn't exposing done with the ingress block ? We are allowing incoming traffic only to 443, which makes sense, but we are also allowing outbound traffic only to 443. I am fine with it, but just curious as to the reason for this.

agnivade:
Any thoughts on this ?

> Lastly, I did not see any uptick in the Query rate panel. Is it because we aren't making any queries at all, or is there a mistake in the panel ? It may well be the first, but just checking.

ethervoid:
Oh sorry I forgot to include the screenshot:

Screenshot from 2020-06-05 19-44-36

I'm using the simple controller and is doing searchs so the grafana panel is working. Which of the controllers are you using?

deployment/terraform/assets/cluster.tf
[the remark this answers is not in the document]
ethervoid: Without egress block set to 443 it wasn't working. Maybe is because some ec2 instances are created in a different vpc?

agnivade:
Ah I was using simulcontroller. Should be fine then. Thanks for checking

PR #312, opened 2020-06-09
[MM-22444] Implement more queries to signal performance degradation

streamer45:
#### Summary

PR adds a couple more queries to `coordinator` that seem to help in detecting performance degradation:

- Lowered the average request duration threshold to 100ms.
- Added query to check for server side errors.
- Added query to check for client request timeouts (currently hard-coded to 5s).

After extensive testing these three seem to cover most cases. I was thinking of adding something CPU related but wasn't getting conclusive results. Other queries would constantly fail before any significant CPU load could be applied.

I'd like to keep testing these for a while before merging and if anything else comes to mind integrate further.

#### Ticket

https://mattermost.atlassian.net/browse/MM-22444


config/coordinator.default.json
```diff
@@ -13,9 +13,21 @@
...
-        "Description": "Request duration",
+        "Description": "Average request duration",
         "Query": "rate(mattermost_http_request_duration_seconds_sum[1m])/rate(mattermost_http_request_duration_seconds_count[1m])",
-        "Threshold": 0.2,
+        "Threshold": 0.1,
+        "Alert": true
+      },
+      {
+        "Description": "Rate of 500 errors",
+        "Query": "sum(rate(mattermost_api_time_count{status_code=\"500\"}[1m]))",
```
agnivade, line 23: Would it be better to detect errors from nginx, which would mean checking client side errors than server side. I was of the mind that nginx can return a 502 in cases when a server is unable to establish a connection. In those cases, there won't be an API response, and the client would still get a response (albeit 502) within timeout.

streamer45: Good point. I admit that was added before I decided to integrate client metrics.  
I guess we can convert that one to something like:

```
sum(rate(loadtest_http_errors_total[1m]))
```

agnivade: Sounds perfect.

config/coordinator.default.json
```diff
@@ -13,9 +13,27 @@
...
-        "Description": "Request duration",
+        "Description": "Average request duration",
         "Query": "rate(mattermost_http_request_duration_seconds_sum[1m])/rate(mattermost_http_request_duration_seconds_count[1m])",
-        "Threshold": 0.2,
+        "Threshold": 0.1,
+        "Alert": true
+      },
+      {
+        "Description": "Average request duration",
+        "Query": "rate(mattermost_api_time_sum[1m])/rate(mattermost_api_time_count[1m])",
```
streamer45, line 23: **Note**

This is actually wrong. These updated metrics do actually return the time/count for each API handler so we need to sum them to get the value we are looking for. Otherwise we'll just consider the first we find in the resulted vector (:man_facepalming:).

PR #313, opened 2020-06-09
[MM-25884] Role for ES is being created when is not required

ethervoid:
#### Summary

AWS Elasticsearch service needs a service-linked role but
the role is unique and shared across the account so if that role
is already created you receive an error creating the cluster

With this new flag, the user is able to check for that role
and set the flag to true to create the role if missing.

Also, the security group is being created when the ES instance
is created. Now is being created always

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-25884


deployment/terraform/assets/cluster.tf
```diff
@@ -496,13 +497,7 @@ resource "aws_security_group" "elastic" {
     security_groups = ["${aws_security_group.app.id}", "${aws_security_group.metrics.id}"]
   }
 
-  egress {
```
ethervoid, line 499: I've been re-checking the need for this, taking into account the feedback received in the last PR, and is not need.

agnivade:
> Now is being created always

I guess you meant the opposite ?

docs/deployer_config.md
```diff
@@ -46,12 +46,23 @@ Number of the instances to be created. Right now only support 1 o 0 values.
...
 
+*string*
+
 Version of Elasticsearch to be deployed. See [here](https://aws.amazon.com/elasticsearch-service/faqs/?nc=sn&loc=6) the supported versions.
 
+### CreateRole
+
+*bool*
+
+Elasticsearch depends on the AWSServiceRoleForAmazonElasticsearchService service-linked role. This role is unique and shares across all the account so if is already created you can't create it again and you'll receive an error.
```
agnivade, line 63: if _it_ is

docs/deployer_config.md
```diff
@@ -46,12 +46,23 @@ Number of the instances to be created. Right now only support 1 o 0 values.
...
+
+You can check if the role is already created [here](https://console.aws.amazon.com/iam/home#roles) and if is not created set this property as true
```
agnivade, line 65: if _it_ is not created, set this property to true.

streamer45:
I am starting to questioning if it's `ltctl`'s job to create this role in the first place.
Maybe we should just explain in the documentation that if ElasticSearch is to be used, then the service linked role should be created (by other means), or already existing, as a requirement.

ethervoid:
@streamer45 I wouldn’t do that. Creating resources in AWS manually could be very intimidating. We can create a FAQ/common errors section in the documentation to give more visibility. Other projects are using the same approach like [this one](https://github.com/cloudposse/terraform-aws-elasticsearch/blob/9065ba41ac47620cf129fe03532842e2cfaf94c3/examples/complete/variables.tf#L61)

@agnivade 

> I guess you meant the opposite ?

No, the security group was being created even if the count for the ES instance was 0

agnivade:
Yes, but your last statement seemed to contradict the previous :) Nvm, I got it.

ethervoid:
@agnivade oh 🤦

streamer45:
> @streamer45 I wouldn’t do that. Creating resources in AWS manually could be very intimidating. We can create a FAQ/common errors section in the documentation to give more visibility. Other projects are using the same approach like [this one](https://github.com/cloudposse/terraform-aws-elasticsearch/blob/9065ba41ac47620cf129fe03532842e2cfaf94c3/examples/complete/variables.tf#L61)

Somehow I am not entirely comfortable with having this tool touching an account wide setting which can potentially affect other users.

A couple of questions:

- Is it possible that the user doesn't have permissions to create this role?
- Would the role get deleted on `deployment destroy` only if it was created by us or even if it existed already?

ethervoid:
> Is it possible that the user doesn't have permissions to create this role?

Yes that could be possible, but that could happen with any resource you are trying to create

> Would the role get deleted on deployment destroy only if it was created by us or even if it existed already?

I have to check that to make sure. Shouldn’t destroy resources not created by terraform. For example we’re using an already created VPC and it’s not being destroyed.

streamer45:
> I have to check that to make sure. Shouldn’t destroy resources not created by terraform. For example we’re using an already created VPC and it’s not being destroyed.

Alright, let's verify that.

Also, regarding the VPC, is it possible to make it required only if ElasticSearch is enabled? (instance count greater than 0 so to speak). That way we would be backward compatible with the previous config.

agnivade:
> Also, regarding the VPC, is it possible to make it required only if ElasticSearch is enabled?

Yes we do that right know IIUC.

streamer45:
> Yes we do that right know IIUC.

Last time I checked I was constantly getting:
```
Error: failed to validate config: vpc id must be included in order to include an elastic search instance
```

`ElasticSearchSettings.InstanceCount` being zero.

ethervoid:
@streamer45 yes the idea was to make all the resources being created in the same vpc. If you don’t want that I’ll move the vpcid inside the elastic config

streamer45:
> @streamer45 yes the idea was to make all the resources being created in the same vpc. If you don’t want that I’ll move the vpcid inside the elastic config

Yes, unless there's a strong reason to keep it for the default case (in which case the error string needs updating).

ethervoid:
The idea was to include all the load tester resources in the same vpc but if that is not needed or desired then we move it to the Es settings

streamer45:
> The idea was to include all the load tester resources in the same vpc but if that is not needed or desired then we move it to the Es settings

My understanding was that the other resources already are in a VPC? It's just that when creating ES you need that VPC's id?

ethervoid:
> My understanding was that the other resources already are in a VPC? It's just that when creating ES you need that VPC's id?

Yes but they are sometimes created in different VPCs. My idea was to leverage the inclusion of the vpcid to make all the resources belong to the same VPC

streamer45:
> Yes but they are sometimes created in different VPCs. My idea was to leverage the inclusion of the vpcid to make all the resources belong to the same VPC

Understood. Then I believe we can unify everything when needed (ES enabled) and leave the default case as is so that the user has one less manual config to set if he's not concerned with testing ES.

agnivade:
> ElasticSearchSettings.InstanceCount being zero.

Ah yes, that was a bug. Because the code checks `c.ElasticSearchSettings != ElasticSearchSettings{}` and zero is the default value :laughing: 

We need to change that to check count.

ethervoid:
What about Leaving it in the upper level and:
- If the vpcid is set is it for all the resources
- set it empty by default
- if ES is being created and the vpcid is empty return and error

streamer45:
> What about Leaving it in the upper level and:
> 
> * If the vpcid is set is it for all the resources
> * set it empty by default
> * if ES is being created and the vpcid is empty return and error

No objections :+1:

PR #314, opened 2020-06-10
[MM-24527] Add GenController documentation

streamer45:
#### Summary

PR adds documentation for `GenController`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-24527

### DIFF
--- a/docs/gencontroller_config.md
+++ b/docs/gencontroller_config.md
@@ -53,3 +53,7 @@ The percentage of direct channels to be created.
 *float64*
 
 The percentage of group channels to be created.
+
+#### Note
+
+The total sum of channels percentages must be equal to 1.
--- a/docs/gencontroller_config.md
+++ b/docs/gencontroller_config.md
@@ -1,0 +1,55 @@
+# GenController Configuration
+
+## NumTeams
+
+*int64*
+
+The number of teams to be created.
+
+## NumChannels
+
+*int64*
+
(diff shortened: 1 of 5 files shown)


docs/generating_data.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Configuration
+
+First of all, the type of `UserController` to be used should be set to
+`generative` as explained [here](loadtest_config.md#usercontrollerconfiguration).
+
+### Copy and modify needed configuration file
+
+```sh
+cp config/gencontroller.default.json config/gencontroller.json
```
agnivade, line 20: I think this will change to sample.json after @isacikgoz's change.

docs/gencontroller_config.md
```diff
@@ -0,0 +1,55 @@
...
+
+*float64*
+
+The percentage of direct channels to be created.
+
+## PercentGroupChannels
+
+*float64*
+
+The percentage of group channels to be created.
```
agnivade, line 55: We should also mention that the sum of these should be equal to 1.

streamer45: Good point. I'll add a note.

docs/generating_data.md
streamer45: Yes, I think we can replace everything in one go when we merge #257.

PR #317, opened 2020-06-16
[MM-25899] Use gencontroller to improve init process

streamer45:
#### Summary

PR improves current `init` process by leveraging `GenController`.
This will give us more control on the initial data to populate when creating a deployment.

#### Ticket

https://mattermost.atlassian.net/browse/MM-25899

### DIFF
--- a/cmd/ltagent/init.go
+++ b/cmd/ltagent/init.go
@@ -112,7 +112,17 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
 		ExpectContinueTimeout: 1 * time.Second,
 	}
 
-	genConfig := config.InstanceConfiguration
+	genConfig := gencontroller.Config{
+		NumTeams:               config.InstanceConfiguration.NumTeams,
+		NumChannels:            config.InstanceConfiguration.NumChannels,
+		NumPosts:               config.InstanceConfiguration.NumPosts,
+		NumReactions:           config.InstanceConfiguration.NumReactions,
+		PercentReplies:         config.InstanceConfiguration.PercentReplies,
+		PercentPublicChannels:  config.InstanceConfiguration.PercentPublicChannels,
+		PercentPrivateChannels: config.InstanceConfiguration.PercentPrivateChannels,
+		PercentDirectChannels:  config.InstanceConfiguration.PercentDirectChannels,
+		PercentGroupChannels:   config.InstanceConfiguration.PercentGroupChannels,
+	}
 
 	newControllerFn := func(id int, status chan<- control.UserStatus) (control.UserController, error) {
 		ueConfig := userentity.Config{
--- a/cmd/ltagent/init.go
+++ b/cmd/ltagent/init.go
@@ -50,7 +50,9 @@ func genData(lt *loadtest.LoadTester, numUsers int64) error {
 	}(time.Now())
 
 	for lt.Status().NumUsersAdded != numUsers {
-		lt.AddUsers(10)
+		if _, err := lt.AddUsers(10); err != nil {
+			return fmt.Errorf("failed to add users %w", err)
+		}
 		time.Sleep(5 * time.Second)
 	}
 
--- a/cmd/ltagent/init.go
+++ b/cmd/ltagent/init.go
@@ -5,93 +5,59 @@ package main
 
 import (
 	"fmt"
-	"math/rand"
+	"net"
+	"net/http"
 	"os"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
-	"github.com/mattermost/mattermost-load-test-ng/loadtest/store"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/user/userentity"
 	"github.com/mattermost/mattermost-load-test-ng/logger"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
-	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/spf13/cobra"
 )
 
-func createTeams(admin *userentity.UserEntity, numTeams int) error {
-	team := &model.Team{
-		AllowOpenInvite: true,
-		Type:            "O",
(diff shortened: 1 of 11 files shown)


loadtest/config.go
```diff
@@ -81,7 +75,7 @@ type UsersConfiguration struct {
 type Config struct {
 	ConnectionConfiguration     ConnectionConfiguration
 	UserControllerConfiguration UserControllerConfiguration
-	InstanceConfiguration       InstanceConfiguration
+	InstanceConfiguration       gencontroller.Config
```
streamer45, line 78: I have to admit I don't like this particularly but the alternative of having an identical structure and copy over the data wasn't perfect either.

cmd/ltagent/init.go
```diff
@@ -107,63 +74,95 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
-		return err
+	lt, err := loadtest.New(config, newControllerFn)
+	if err != nil {
+		return fmt.Errorf("error while initializing loadtest: %w", err)
 	}
 
-	mlog.Info("done", mlog.String("elapsed", time.Since(start).String()))
-
-	return nil
+	return genData(lt, 50)
```
streamer45, line 153: These values (rate and number of users) come from various benchmarks that I ran. I was able to get decent times with these on different configurations.

cmd/ltagent/loadtest.go
```diff
@@ -26,30 +25,19 @@ import (
...
-		defer wg.Done()
-		for {
-			if status := lt.Status(); status.NumUsersStopped == int64(numUsers) {
-				err = lt.Stop()
-				return
-			}
-			time.Sleep(1 * time.Second)
-		}
-	}()
-	wg.Wait()
```
streamer45, line 48: Don't ask me why I wrote this.

cmd/ltagent/init.go
```diff
@@ -5,93 +5,60 @@ package main
...
 	"github.com/mattermost/mattermost-server/v5/mlog"
-	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/spf13/cobra"
 )
 
-func createTeams(admin *userentity.UserEntity, numTeams int) error {
-	team := &model.Team{
-		AllowOpenInvite: true,
-		Type:            "O",
+func initDone(serverURL, userPrefix string) (bool, error) {
```
agnivade, line 25: `isInitDone` sounds better.

loadtest/config.go
streamer45: `GenController` config would then be coupled with a higher level construct. I tend to prefer this direction when possible (higher level struct including lower level struct).

streamer45:
> 1. I think the `int64` fix that you made probably still has some bugs. I had to manually add

I think the fix was fine but I've definitely forgot to update the default config to:

```json
  "InstanceConfiguration": {
    "NumTeams": 2,
    "NumChannels": 20,
    "NumPosts": 1000,
    "NumReactions": 200,
    "PercentReplies": 0.5,
    "PercentPublicChannels": 0.2,
    "PercentPrivateChannels": 0.1,
    "PercentDirectChannels": 0.6,
    "PercentGroupChannels": 0.1
  }
```

> 2. We are not checking the return value of AddUsers. This leaves out a very sneaky bug because it does not know there is an error and goes into an infinite loop without logging anything.

Will add a check :+1: 

> it worked. If we are always taking the value from config.json, is there a reason we have a separate gencontroller.json?
> To me, values in config.json should be top-level and usable by any controller. If we have gencontroller specific fields in config.json, maybe we should only keep that in gencontroller.json? Maybe there's a bug somewhere here.

The reason to keep it at the top-level is because of the `init` process. We could have everything only in `gencontroller.json` but that would mean an additional config file and we would also be slightly hiding that information through indirection.

I believe duplicating some config values might be the best option here. Using `GenController` during the init process is just our convenience as developers, not something our users should be concerned about. They should be just looking at what/how much data they want to be created.

That's why I kept `InstanceConfiguration` which is not something related to any controller. It's part of the load-test configuration. It currently matches `gencontroller.Config` and that's why I've used that directly but we can change it either by keeping an identical `loadtest.InstanceConfiguration` struct or through some embedding.

> 3. I don't see all of the config values used in the init help

I am not sure I understood. What values were you expecting?

loadtest/config.go
agnivade: Generally, that would be okay. But the main question I had while reading the code is why `gencontroller`? Because `gencontroller` is one such controller from a whole lot of controllers. What is so special about that it needs to be included in the loadtest package?

My reasoning was if we do the other way round, gencontroller is just taking a top-level struct which is free to be used by any low-level packages. Other controllers may or may not use that, and that's fine.

agnivade:
> The reason to keep it at the top-level is because of the init process. We could have everything only in gencontroller.json but that would mean an additional config file and we would also be slightly hiding that information through indirection.

Yeah, it seems like we want to hide the abstraction that init actually runs a controller underneath. I'm okay with things as it is now.

> I am not sure I understood. What values were you expecting?

I mean that all the flag options in the init command aren't actually used by the init command. They are just shown because init is a sub-command of loadtest and are inherited from the parent command's flags. My point was to investigate whether those flags are meant to be inherited by any other commands than the main command, and if yes, then replace PersistenFlags with Flags.

loadtest/config.go
streamer45: > I mean that all the flag options in the init command aren't actually used by the init command. They are just shown because init is a sub-command of loadtest and are inherited from the parent command's flags. My point was to investigate whether those flags are meant to be inherited by any other commands than the main command, and if yes, then replace PersistentFlags with Flags.

Understood. Yes, some flags are clearly not meant to be part of the `init` subcommand. I'll remove them.

streamer45:
> I mean that all the flag options in the init command aren't actually used by the init command. They are just shown because init is a sub-command of loadtest and are inherited from the parent command's flags. My point was to investigate whether those flags are meant to be inherited by any other commands than the main command, and if yes, then replace PersistentFlags with Flags.

Understood. Yes, some flags are clearly not meant to be part of the `init` subcommand. I'll remove them.

loadtest/config.go
streamer45: Yes, the confusion is reasonable because it looks odd indeed. Maybe it's fine to have `InstanceConfiguration` be a structure that mirrors (at least for now) `gencontroller.Config`.
It just means copy over the values in `cmd/ltagent/init.go`.

agnivade: That works. A little duplication is better than bringing odd package dependencies.

PR #321, opened 2020-06-18
[MM-25315] Implement HTTP API for coordinator

streamer45:
#### Summary

PR implements a basic HTTP API for coordinator, similarly to what we already have for the load-test agent. The goal is being able to programmatically spawn multiple coordinators from the same instance to enable a multi-instance load-test scenario.

The biggest change is that I've unified the two APIs into a single command `ltapi` that can deal with both resources. I've also slightly restructured the `api` package.

A separate to address documentation changes will follow. First I'd like to get this reviewed.

#### Ticket

https://mattermost.atlassian.net/browse/MM-25315


coordinator/agent/agent.go
```diff
@@ -27,6 +26,15 @@ type LoadAgent struct {
 	client *http.Client
 }
 
+// TODO: maybe move this into a shared model package.
+// agentResponse contains the data returned by load-test agent API.
+type agentResponse struct {
```
streamer45, line 31: This was needed to avoid a cyclic inclusion between packages. Didn't want to bring a `model` package so I opted for some minor duplication.

cmd/ltagent/loadtest.go
```diff
@@ -155,73 +149,3 @@ func MakeLoadTestCommand() *cobra.Command {
 	cmd.PersistentFlags().IntP("user-offset", "", 0, "numerical offset applied to user ids")
 	return cmd
 }
-
-func newControllerWrapper(config *loadtest.Config, controllerConfig interface{}, userOffset int, namePrefix string, metrics *performance.Metrics) loadtest.NewController {
```
streamer45, line 159: It didn't live very long there, apologizes @isacikgoz :mage:

streamer45:
> * Please add the new steps of how to run a coordinator in the docs.

Yes, that's what the "Do not merge" label was about :P

> * Just so I understand this clearly, this effectively means that now the coordinator with merge with the agent service. The loadtests will just start agents (now ltapi), and then start the coordinator through the API. There will be no more a separate ltcoordinator service. Is that correct?

Correct. We'll run the single binary `ltapi` regardless of whether we'll be running agents, coordinators or both. `ltctl` will be in charge of creating/starting/stopping a coordinator.
From the user's perspective nothing is going to change, except for the `ltagent server` case.

> * Doesn't the coordinator start/shutdown through terraform needs to change after this? O
r do you plan to do it later?

I plan to do it before merging this but in a separate PR, mainly to aid the reviewing process.

agnivade:
Ok, I will wait until we have the steps before I make a second sweep. But overall looks good to me. :+1:

coordinator/coordinator.go
```diff
@@ -133,8 +179,12 @@ func New(config *Config) (*Coordinator, error) {
 	}
 
 	return &Coordinator{
-		config:  config,
-		cluster: cluster,
-		monitor: monitor,
+		mut:      sync.RWMutex{},
```
agnivade, line 182: No need for this.

coordinator/coordinator.go
```diff
@@ -133,8 +179,12 @@ func New(config *Config) (*Coordinator, error) {
...
+		stopChan: make(chan struct{}),
+		doneChan: make(chan struct{}),
+		status:   Status{},
```
agnivade, line 185: No need for this.

streamer45:
1.

Yes, we can improve that one.

2.

Interestingly that was by design. If you notice I don't have a test trying to run it twice once stopped and I never re-initialize the `stopChan` either.
I am not sure making it runnable multiple times (like the load-test agent) is worth it. 
I believe we can either enforce this behaviour through state or simply unexport the `Stop` method and just provide `destroy` functionality at the API level. I probably prefer the first approach.

agnivade:
2. Ah I see. Ok, as long as we document it, I am fine with it. Thanks.

streamer45:
> 1. Ah I see. Ok, as long as we document it, I am fine with it. Thanks.

I actually tried to document it here:

https://github.com/mattermost/mattermost-load-test-ng/blob/1e1de11709309ab9924fe7fa5413dde59ff0fcab/coordinator/coordinator.go#L32

I can easily add a check though. I had plans to add a new `Done` state  anyway.

agnivade:
I meant like documenting it publicly in some markdown files that re-creating a new coordinator test through the HTTP API will fail.

streamer45:
> I meant like documenting it publicly in some markdown files that re-creating a new coordinator test through the HTTP API will fail.

I can add a small note :+1:

PR #327, opened 2020-06-22
[MM-25970] Enable multi-coordinator deployments

streamer45:
#### Summary

This PR is a follow-up to #321. From a user's perspective the deployment process should work as before. We are simply replacing `ltcoordinator` binary for `ltapi` and creating a coordinator using the new API.
PR also updates the documentation to reflect the new changes. 

#### Ticket

https://mattermost.atlassian.net/browse/MM-25970


deployment/terraform/coordinator.go
```diff
@@ -113,9 +115,34 @@ func (t *Terraform) StartCoordinator() error {
...
+		LoadTestConfig    loadtest.Config
+	}{
+		*coordinatorConfig,
+		*agentConfig,
+	})
+	if err != nil {
+		return err
+	}
+	// TODO: create a wrapper to simplify coordinator's HTTP API usage.
+	id := fmt.Sprintf("%s-coordinator-%d", t.config.ClusterName, 0)
```
agnivade, line 129: Doesn't this restrict one user to run only one coordinator? Or you are saying the differentation comes from the cluster name? In which case, we should just remove the "-0" part since that doesn't really mean anything.

deployment/terraform/coordinator.go
```diff
@@ -113,9 +115,34 @@ func (t *Terraform) StartCoordinator() error {
...
+	apiURL := fmt.Sprintf("http://%s:4000/coordinator/%s", ip, id)
```
agnivade, line 130: Should we also take the port from the config somewhere later? The port is actually exposed via a flag, it's just not captured in a config file.

deployment/terraform/coordinator.go
```diff
@@ -138,19 +165,23 @@ func (t *Terraform) StopCoordinator() error {
 	}
 	ip := output.Agents.Value[0].PublicIP
 
-	extAgent, err := ssh.NewAgent()
+	id := fmt.Sprintf("%s-coordinator-%d", t.config.ClusterName, 0)
+
+	mlog.Info("Stopping the coordinator", mlog.String("ip", ip))
+
+	apiURL := fmt.Sprintf("http://%s:4000/coordinator/%s", ip, id)
+	client := &http.Client{}
```
agnivade, line 173: Why not use `http.DefaultClient`? It is effectively the same thing.

streamer45: No reason. That code should be gone as soon as we find the time to create a simple wrapper so didn't spend any time and went with the first thing I thought of.

deployment/terraform/coordinator.go
streamer45: Yes, we can probably make a ticket for that but that might mean a new config file (for the API server).

deployment/terraform/coordinator.go
streamer45: The idea was to address that in [MM-26387](https://mattermost.atlassian.net/browse/MM-26387). Here really trying to get things working as they are now (before the two PRs merged) and just preparing the code for what's next.

streamer45:
> What I wanted to have were the manual steps to start/stop a coordinator via curl. Would it be possible to add those too?
> 
> Things like:
> 
> ```
> curl -d "{\"LoadTestConfig\": $(cat config/config.json), \"CoordinatorConfig\": $(cat config/coordinator.json)}" http://localhost:4000/coordinator/create?id=lt1
> curl -X POST http://localhost:4000/coordinator/lt1/run
> curl  http://localhost:4000/coordinator/lt1/status
> ```


I was mainly focused on fixing breaking changes but we can certainly add them :+1:

deployment/terraform/coordinator.go
streamer45: I was actually annoyed that `http` didn't provide a `Delete()` method which meant I had to create the request.

docs/local_loadtest.md
```diff
@@ -167,3 +167,38 @@ go run ./cmd/ltcoordinator
 ```
 
 This will start running a load-test across the configured cluster of load-test agents.
+
+### Run coordinator using the API server
+
+Similarly to what happens for the load-test agents, a coordinator (or more than
```
agnivade, line 173: Similar to what ..

docs/local_loadtest.md
```diff
@@ -167,3 +167,38 @@ go run ./cmd/ltcoordinator
...
+one) can be created and ran using the API server.
```
agnivade, line 174: and run using ...

PR #331, opened 2020-06-24
MM-26421: calculate diff after rounding off

agnivade:
Right now, we are calculating diff from the original value, but printing rounded off values.
That shows confusing output as the diff and the values don't match up.

This PR changes that to show proper time units in duration rounded off to ms.

This changes

```
ChannelStore.GetMembersForUser |  Avg| 0.002| 0.002 | 0.000 | 3.116
| |  P99| 0.008| 0.008 | 0.001 | 7.457
| ChannelStore.GetPinnedPostCount |  Avg| 0.000| 0.000 | -0.000 | -2.957
| |  P99| 0.005| 0.005 | -0.000 | -0.043
| ChannelStore.GetPublicChannelsForTeam |  Avg| 0.001| 0.001 | -0.000 | -2.145
| |  P99| 0.005| 0.005 | -0.000 | -0.246
```

to

```
| ChannelStore.GetMembersForUser |  Avg| 2ms| 14ms | 12ms | 85.714
| |  P99| 8ms| 340ms | 332ms | 97.647
| ChannelStore.GetPinnedPostCount |  Avg| 0s| 2ms | 2ms | 100.000
| |  P99| 5ms| 23ms | 18ms | 78.261
| ChannelStore.GetPublicChannelsForTeam |  Avg| 1ms| 11ms | 10ms | 90.909
| |  P99| 5ms| 216ms | 211ms | 97.685
```

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26421

### DIFF
--- a/loadtest/report/compare.go
+++ b/loadtest/report/compare.go
@@ -5,15 +5,17 @@ package report
 
 import (
 	"fmt"
+	"math"
 	"os"
+	"time"
 
 	"github.com/prometheus/common/model"
 )
 
 // diff contains the differences from a base measurement.
 type diff struct {
-	actual       float64
-	delta        float64
+	actual       time.Duration
+	delta        time.Duration
 	deltaPercent float64
 }
 
@@ -104,9 +106,12 @@ func calculateDeltas(reports ...Report) comp {
 		// XXX: This can be somewhat refactored but whether absolute metrics
 		// are useful or not needs to be seen.
 		for label, value := range base.AvgStoreTimes {
-			actual := roundTo3DecimalPlaces(float64(r.AvgStoreTimes[label]))
-			delta := actual - roundTo3DecimalPlaces(float64(value))
-			deltaP := (delta / actual) * 100
+			actual := getDuration(float64(r.AvgStoreTimes[label]))
+			delta := actual - getDuration(float64(value))
+			deltaP := (delta.Seconds() / actual.Seconds()) * 100
+			if math.IsNaN(deltaP) {
+				deltaP = 0
+			}
 
 			diffs := c.store[label]
 			diffs[0] = append(diffs[0], diff{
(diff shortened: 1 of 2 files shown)


loadtest/report/output.go
```diff
@@ -167,3 +168,7 @@ func sortKeys(m map[model.LabelValue]avgp99) []model.LabelValue {
 	})
 	return labels
 }
+
+func roundTo3DecimalPlaces(f float64) float64 {
```
streamer45, line 172: That's some explicit naming :grinning:

agnivade: It's the hardest problem in computer science for a reason. :)

agnivade:
1. Do you have any better suggestions?
2. Yes, I had thought about it a lot. Ultimately, I decided to go with seconds because there are some p99 metrics where things do take a full second. And moreover, it's a natural tendency to count in seconds. I gave a try by changing it to ms and it felt a bit weird seeing the metrics. It might also be possible to increase the precision by one more. Open to suggestions.

streamer45:
> 1. Do you have any better suggestions?

What's causing the `NaN`? Would 0 be applicable?

> 2. Yes, I had thought about it a lot. Ultimately, I decided to go with seconds because there are some p99 metrics where things do take a full second. And moreover, it's a natural tendency to count in seconds. I gave a try by changing it to ms and it felt a bit weird seeing the metrics. It might also be possible to increase the precision by one more. Open to suggestions.

One option would be to use units. We default to `ms` and if it becomes higher we go with `s` and so on.

agnivade:
> What's causing the NaN? Would 0 be applicable?

It's because while trying to calculate the percent, we are doing a 0/0. 0 is technically not applicable here. NaN is the right answer.

> One option would be to use units. We default to ms and if it becomes higher we go with s and so on.

Sounds good to me. I will make the change.

streamer45:
> It's because while trying to calculate the percent, we are doing a 0/0. 0 is technically not applicable here. NaN is the right answer.

I see but at the high level that means there's no difference between the two values right? If so, I would probably expect the percentage to be 0.

agnivade:
Sure.

loadtest/report/compare.go
```diff
@@ -104,9 +106,12 @@ func calculateDeltas(reports ...Report) comp {
...
 		for label, value := range base.AvgStoreTimes {
-			actual := roundTo3DecimalPlaces(float64(r.AvgStoreTimes[label]))
-			delta := actual - roundTo3DecimalPlaces(float64(value))
-			deltaP := (delta / actual) * 100
+			actual := getDuration(float64(r.AvgStoreTimes[label]))
+			delta := actual - getDuration(float64(value))
+			deltaP := (delta.Seconds() / actual.Seconds()) * 100
+			if math.IsNaN(deltaP) {
+				deltaP = 0
+			}
```
streamer45, line 114: The other option would be not to bother calculating `deltaP` if `delta` is zero.

agnivade: Yeah .. that was my first approach. Somehow didn't like the code arrangement. :P Plus I get to use an esoteric stdlib function. :)

PR #333, opened 2020-06-24
[MM-26459] Split logs for agent and coordinator

streamer45:
#### Summary

PR splits logging for the two main components. 
This is needed to avoid having a single `ltapi.log` file with everything inside.

#### Ticket

https://mattermost.atlassian.net/browse/MM-26459

### DIFF
--- a/api/server.go
+++ b/api/server.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/performance"
 
 	"github.com/gorilla/mux"
+	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
 // api keeps track of the load-test API server state.
@@ -19,6 +20,8 @@ type api struct {
 	agents       map[string]*loadtest.LoadTester
 	coordinators map[string]*coordinator.Coordinator
 	metrics      *performance.Metrics
+	clog         *mlog.Logger
+	alog         *mlog.Logger
 }
 
 func (a *api) pprofIndexHandler(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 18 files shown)


logger/logger.go
```diff
@@ -19,6 +19,18 @@ type Settings struct {
 	FileLocation  string `default:"loadtest.log"`
 }
 
+func New(logSettings *Settings) *mlog.Logger {
```
agnivade, line 22: godoc please :)

api/server.go
```diff
@@ -12,13 +12,16 @@ import (
...
+	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
 // api keeps track of the load-test API server state.
 type api struct {
 	agents       map[string]*loadtest.LoadTester
 	coordinators map[string]*coordinator.Coordinator
 	metrics      *performance.Metrics
+	clog         *mlog.Logger
+	alog         *mlog.Logger
```
agnivade, line 24: Would prefer them to be `coordLog` and `agentLog`. Somebody reading the codebase without any context would not understand quickly what they mean.

cmd/ltapi/main.go
```diff
@@ -18,6 +18,7 @@ import (
 func RunServerCmdF(cmd *cobra.Command, args []string) error {
 	port, _ := cmd.Flags().GetInt("port")
 
+	// TODO: add a config file for the API server.
```
agnivade, line 21: This is getting really out of hand now with so many config files :)

streamer45: Well, I am still not convinced by the monolithic config approach :)

api/server.go
streamer45: Fair enough :)

coordinator/cluster/cluster.go
```diff
@@ -29,14 +30,17 @@ type errorTrack struct {
 
 // New creates and initializes a new LoadAgentCluster for the given config.
 // An error is returned if the initialization fails.
-func New(config LoadAgentClusterConfig) (*LoadAgentCluster, error) {
+func New(config LoadAgentClusterConfig, log *mlog.Logger) (*LoadAgentCluster, error) {
+	if log == nil {
+		return nil, fmt.Errorf("logger should not be nil")
```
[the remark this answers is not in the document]
streamer45: Yes, my reasoning there was to use `fmt` if the `errors` package was not already imported. It felt weird to import one package for a single log line. Also, [MM-23914](https://mattermost.atlassian.net/browse/MM-23914) is around the corner.

PR #335, opened 2020-06-25
Remove everything with mattermost-load-test-ng in agent creation

agnivade:
If there is already a .tar.gz file present, then this throws an error
while trying to move the directory. We therefore delete everything with that name.


streamer45:
> If there is already a .tar.gz file present, then this throws an error
> while trying to move the directory. We therefore delete everything with that name.

Is this `tar.gz` file something our tool uploads or just a custom file uploaded by the user?

agnivade:
Nope, it was uploaded by me. But I think the user would expect it to be overwritten anyways. So there's no harm in just removing everything rather than throw a weird error.

PR #338, opened 2020-06-26
[MM-26387] Enable load-test only deployments

streamer45:
#### Summary

PR enables deployments where `AppInstanceCount` is equal to zero.
This lets us create only agent instances to be used to load-test an existing deployment.
PR also makes security groups creation conditional so to avoid creating unneeded resources.

#### Ticket

https://mattermost.atlassian.net/browse/MM-26387


deployment/terraform/assets/cluster.tf
```diff
@@ -501,24 +513,26 @@ resource "aws_security_group" "elastic" {
...
   }
 
-  count = var.es_instance_count
+  count = var.es_instance_count > 0 ? 1 : 0
 }
 
 # We need a separate security group rule to prevent cyclic dependency between
 # the app group and metrics group.
 resource "aws_security_group_rule" "app-to-inbucket" {
+  count                    = var.app_instance_count > 1 ? 1 : 0
```
streamer45, line 525:
```suggestion
  count                    = var.app_instance_count > 0 ? 1 : 0
```

deployment/terraform/assets/cluster.tf
```diff
@@ -34,8 +34,8 @@ resource "aws_instance" "app_server" {
   key_name      = aws_key_pair.key.id
   count         = var.app_instance_count
   vpc_security_group_ids = [
-    "${aws_security_group.app.id}",
-    "${aws_security_group.app_gossip.id}"
+    "${aws_security_group.app[0].id}",
+    "${aws_security_group.app_gossip[0].id}"
```
streamer45, line 38: This is somehow problematic and will fail with an "empty tuple error" in case of a single app server deployment. 
Not sure yet if making a bunch of `aws_security_group_rule` is the way to solve this case.

agnivade: Not sure if it's worth trying to optimize some spurious security group creations, if it's taking too much effort. Thoughts on leaving it as is?

streamer45: I tend to agree. Ideally I wanted to create only the strictly necessary but somehow I feel this is not the way to do it. Not sure if it's a limitation with terraform or just us using it not the intended way.

agnivade: Yeah, I don't think it's a limitation per-se. But security group dependencies are a bit tricky and time-consuming to get it right. If it's not a big issue, IMO there's no need to spend too much time on it and just let it be.

streamer45:
So, since most of the work was done already I am inclined to just fix the current problem which means we are going to create a security group for gossip even for a single app instance deployment. Apart from that nothing else should get created which is not needed. 

Will now test this with a few different configs (load-test only, single app instance, multiple app instances).

streamer45:
Tested the three cases and didn't notice any issue.

PR #340, opened 2020-06-29
MM-26558: Fix index for multi-report comparison

agnivade:
We were using the wrong index to append further reports.

#### Ticket link

https://mattermost.atlassian.net/browse/MM-26558

### DIFF
--- a/loadtest/report/compare.go
+++ b/loadtest/report/compare.go
@@ -36,6 +36,11 @@ type labelValues struct {
 	values []model.SamplePair // The series of values for a given metric.
 }
 
+type gplot struct {
+	name   string // the name of the metric
+	graphs []labelValues
+}
+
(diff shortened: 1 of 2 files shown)


loadtest/report/compare.go
```diff
@@ -50,41 +55,7 @@ func Compare(target *os.File, genGraph bool, reports ...Report) error {
...
-					// Graph of the same type from another report, append it.
-					gPlots[i].graphs = append(gPlots[i].graphs, labelValues{
-						label:  r.Label,
-						values: g.Values,
-					})
-				}
-			}
-		}
-
+		gPlots := getPlots(reports...)
```
streamer45, line 58: Does it make sense to pass reports without the base here? So that `getPlots` can work on a slice of reports without worrying about this implementation detail.

PR #346, opened 2020-07-03
Use weighted lines for graphs instead of points

agnivade:
Makes it look a bit more nice and clean.


agnivade:
Old: https://github.com/mattermost/mattermost-server/pull/14691#issuecomment-648017462

New: https://community-daily.mattermost.com/core/channels/developers-performance/gxxx7qufgib5tpctb758tdz6oa

The difference becomes clear when there are more points and more graphs. It's much less crowded and clear.

PR #347, opened 2020-07-03
[MM-26588] De-couple API client driver and integrate it in current code

streamer45:
#### Summary

PR refactors the API client driver to decouple `agent` and `coordinator`. This allows us to integrate the agent client in our coordinator package without creating nasty import cycles. 

This PR is based on top of #342. Most of the internal client code is the same with a few exceptions (removed the client wrapper that was causing the coupling). This will essentially replace that PR which I am going to close once this has been approved. 

The integration happened in two places:

- `deployment/terraform/coordinator.go`
- `coordinator/cluster/cluster.go`

Once approved I'll run a couple of load-tests before merging since the change is quite substantial.

#### Ticket

https://mattermost.atlassian.net/browse/MM-26588


coordinator/cluster/cluster.go
```diff
@@ -91,9 +108,9 @@ func (c *LoadAgentCluster) Shutdown() {
 	var wg sync.WaitGroup
 	wg.Add(len(c.agents))
 	for _, ag := range c.agents {
-		go func(ag *agent.LoadAgent) {
+		go func(ag *client.Agent) {
 			defer wg.Done()
-			if err := ag.Stop(); err != nil {
+			if _, err := ag.Stop(); err != nil {
```
streamer45, line 113:
```suggestion
			if _, err := ag.Destroy(); err != nil {
```

The previous behaviour was actually to destroy the agents. Not sure why it ended up being called `Stop()`.
This actually turned out to be useful and uncovered [MM-26694](https://mattermost.atlassian.net/browse/MM-26694) during testing.

api/coordinator.go
```diff
@@ -8,22 +8,15 @@ import (
...
-// CoordinatorResponse contains the data returned by load-test coordinator API.
-type CoordinatorResponse struct {
-	Id      string              `json:"id,omitempty"`      // The load-test coordinator unique identifier.
-	Message string              `json:"message,omitempty"` // Message contains information about the response.
-	Status  *coordinator.Status `json:"status,omitempty"`  // Status contains the current status of the coordinator.
-	Error   string              `json:"error,omitempty"`   // Error is set if there was an error during the operation.
-}
-
-func writeCoordinatorResponse(w http.ResponseWriter, status int, resp *CoordinatorResponse) {
+func writecoordinatorResponse(w http.ResponseWriter, status int, resp *client.CoordinatorResponse) {
```
agnivade, line 19: Any reason the C became lower-case?

coordinator/cluster/cluster.go
```diff
@@ -38,20 +40,37 @@ func New(config LoadAgentClusterConfig, ltConfig loadtest.Config, log *mlog.Logg
...
+		agent, err := client.New(config.Agents[i].Id, config.Agents[i].ApiURL, nil)
 		if err != nil {
-			return nil, fmt.Errorf("cluster: failed to create agent: %w", err)
+			return nil, fmt.Errorf("cluster: failed to create api client: %w", err)
 		}
 		agents[i] = agent
 		errMap[agent] = errorTrack{}
+
+		// We check if the agent has already been created.
+		if _, err := agent.Status(); err == nil {
```
agnivade, line 54: Do we want to have a dedicated API for this which just returns the id or something? Totally fine with this too. Just thinking out loud.

api/coordinator.go
streamer45: I am guessing an erroneous string replace when I exported `CoordinatorResponse` struct.  
Nice catch :+1:

coordinator/cluster/cluster.go
streamer45: That went through my mind. I was also thinking of having the `Create()` method return a special `ErrExists` error than could be ignored in this case (to avoid making two calls essentially) but in the end didn't settle on it and went for this.
If the resource exists we can return the status of it so I thought this was the closest thing to an `Exists()` method since it's essentially a simple `GET /loadagent/${id}`.

agnivade: Hmm .. okay. Perhaps in future, we can wrap this into an `Exists` method to make things more readable. Fine as is for now.

PR #352, opened 2020-07-07
[MM-25969] Investigate running load-test through public internet

streamer45:
#### Summary

PR adds a panel with the client request times to the default Grafana dashboard. 
This turned out to be quite useful during this investigation so I'd thought of keeping it for future needs.

#### Investigation results

Most interesting/relevant insights:

- Server-side average API call durations don't match the client-side ones as precisely as expected.
   - We should consider adding an average client request duration threshold to coordinator's queries.
   - Investigate what's causing the added latency.
- Running through a proxy (nginx) causes coordinator to lower users count sooner.
  - Further investigation is needed. I've opened [MM-26764](https://mattermost.atlassian.net/browse/MM-26764).
- Running through the public internet doesn't seem to have much of an impact, at least in the case of the agents located in the same region as the app instances.
  - As a result of this I wouldn't touch the query threshold (currently set at 100ms).

### 2 instances setup with default configs

image
Server-side averages here are extremely low. This seems to be caused by nginx (caching?).

image
Client-side is a totally different picture. Both averages and p99 are substantially higher.

image

image

image

### Agents configured to directly hit 1 instance only

image

image
Client-side averages without the proxy in between are very high. With a proper query we would never let the coordinator reach this point. It's interesting to notice that server side average was well under the limit (100ms) and no other significant error occurred.

image

image
Very smooth line clearly indicating the ceiling.

image
We can clearly see the app0 instance is used to its CPU limits. 

### Only 1 proxy upstream configured

image
Again, with the proxy in between server-side averages are kept very low. 

image
Similarly to the first test we can see the proxy is doing a good job at keeping averages low but somehow it's failing sooner on other metrics (possibly timeouts and/or 500 errors).

image

image

image

#### Ticket

https://mattermost.atlassian.net/browse/MM-25969


agnivade:
Nice work.

From what I understand, the running theory is that nginx does some caching due to which, in cases where a proxy is running, server gets lower RPS as seen from the graphs. And as a result, it has lower request duration, and clients also get the response faster.

But somehow inspite of this, the coordinator somehow does the opposite thing where if nginx is running, it lowers the user count sooner than expected and thereby we reach a lower ceiling that an equivalent setup with only one upstream configured? Is that accurate?

streamer45:
> Nice work.
> 
> From what I understand, the running theory is that nginx does some caching due to which, in cases where a proxy is running, server gets lower RPS as seen from the graphs. And as a result, it has lower request duration, and clients also get the response faster.
> 
> But somehow inspite of this, the coordinator somehow does the opposite thing where if nginx is running, it lowers the user count sooner than expected and thereby we reach a lower ceiling that an equivalent setup with only one upstream configured? Is that accurate?

Yes, that's an accurate summary. Nginx does a good job at keeping request times low but surprisingly ends up failing sooner/more often on other metrics.

For some more context, here are the graphs for client timeouts which ultimately are what made the coordinator lower the user count since the server-side averages never went over their threshold.

#### Without proxy:

image

More timeouts but well defined pattern.

#### With proxy:
image

Less timeouts but also less defined (more spread) pattern.

agnivade:
Very interesting. Might be worth experimenting with the `max_fails` value.

streamer45:
> Very interesting. Might be worth experimenting with the `max_fails` value.

Yes, that's definitely something that could be tried during MM-26764.

Also I think starting to use client request times would probably help mitigating this problem. Ideally we don't want to reach such a high number of timeouts so maybe a combination of average and p99 queries for the client-side could help to prevent that.

PR #354, opened 2020-07-08
[MM-26784] Fix data race in cluster.Status()

streamer45:
#### Summary

PR fixes a data race (and possible crash) in `cluster.Status()`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-26784


Makefile
```diff
@@ -93,7 +93,7 @@ golangci-lint:
 	golangci-lint run ./...
 
 test:
-	$(GO) test -v -mod=readonly -failfast ./...
+	$(GO) test -v -mod=readonly -failfast -race ./...
```
streamer45, line 96: This would have been spotted before.

streamer45:
Nice, looks like CI has spotted something else :eyes:

streamer45:
> Woops thanks. Finally we enable the `-race` flag :)
> 
> I am wondering why isn't the CI failing if there are other races?

I believe it's our first "flaky test". I wasn't able to reproduce it locally and re-running the CI pipeline worked.
Somehow it has to do with https://golang.org/src/testing/testing.go?s=31943:32093#L975

PR #363, opened 2020-07-10
[MM-26759] Provide sample app making use of the API driver

streamer45:
#### Summary

PR provides a sample application that makes use of the new API driver.
The app makes it easier to run a multi-instance load-test.

#### Ticket

https://mattermost.atlassian.net/browse/MM-26759


cmd/ltctl/reset.go
```diff
@@ -68,6 +68,11 @@ func RunResetCmdF(cmd *cobra.Command, args []string) error {
 			value:  fmt.Sprintf("%s reset --confirm", binaryPath),
 			client: appClient,
 		},
+		{
+			msg:    "Restarting app server",
+			value:  fmt.Sprintf("sudo systemctl restart mattermost"),
+			client: appClient,
+		},
```
streamer45, line 75: Unrelated but I spotted it and included it here.

PR #366, opened 2020-07-13
[MM-23496] Fix race conditions when creating/destroying API resources

streamer45:
#### Summary

PR fixes a couple of high level race conditions with creating/destroying API resources:
Trying to create a resource with the same id concurrently could cause more than one to be created.

This should be the last known race condition hence closing the investigation ticket with it.

#### Example

```sh
curl -d "{\"CoordinatorConfig\": $(cat config/coordinator.json),\"LoadTestConfig\": $(cat config/config.json)}" http://localhost:4000/coordinator/create\?id\=ltc0 &
curl -d "{\"CoordinatorConfig\": $(cat config/coordinator.json),\"LoadTestConfig\": $(cat config/config.json)}" http://localhost:4000/coordinator/create\?id\=ltc0 &
curl -d "{\"CoordinatorConfig\": $(cat config/coordinator.json),\"LoadTestConfig\": $(cat config/config.json)}" http://localhost:4000/coordinator/create\?id\=ltc0 &
curl -d "{\"CoordinatorConfig\": $(cat config/coordinator.json),\"LoadTestConfig\": $(cat config/config.json)}" http://localhost:4000/coordinator/create\?id\=ltc0 &


{"id":"ltc0","message":"load-test coordinator creation failed","error":"could not create coordinator: coordinator: failed to create cluster: cluster: failed to create agent: agent: load-test agent api request error: load-test agent with id lt0 already exists"}
{"id":"ltc0","message":"load-test coordinator created","status":{"State":"stopped","StartTime":"0001-01-01T00:00:00Z","StopTime":"0001-01-01T00:00:00Z","ActiveUsers":0,"NumErrors":0,"SupportedUsers":0}}
{"id":"ltc0","message":"load-test coordinator creation failed","error":"could not create coordinator: coordinator: failed to create cluster: cluster: failed to create agent: agent: load-test agent api request error: load-test agent with id lt0 already exists"}
{"id":"ltc0","message":"load-test coordinator created","status":{"State":"stopped","StartTime":"0001-01-01T00:00:00Z","StopTime":"0001-01-01T00:00:00Z","ActiveUsers":0,"NumErrors":0,"SupportedUsers":0}}
```

#### Ticket

https://mattermost.atlassian.net/browse/MM-23496

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -116,7 +116,12 @@ func (a *api) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
-	a.setResource(agentId, lt)
+	if ok := a.setResource(agentId, lt); !ok {
+		writeAgentResponse(w, http.StatusBadRequest, &client.AgentResponse{
+			Error: fmt.Sprintf("resource with id %s already exists", agentId),
+		})
+		return
+	}
 
 	writeAgentResponse(w, http.StatusCreated, &client.AgentResponse{
 		Id:      agentId,
@@ -192,7 +197,13 @@ func (a *api) destroyLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 
 	_ = lt.Stop() // we are ignoring the error here in case the load test was previously stopped
 
-	a.deleteResource(mux.Vars(r)["id"])
+	id := mux.Vars(r)["id"]
+	if ok := a.deleteResource(id); !ok {
+		writeAgentResponse(w, http.StatusNotFound, &client.AgentResponse{
+			Error: fmt.Sprintf("load-test agent with id %s not found", id),
+		})
+		return
+	}
 	writeAgentResponse(w, http.StatusOK, &client.AgentResponse{
 		Message: "load-test agent destroyed",
 		Status:  lt.Status(),
--- a/api/agent.go
+++ b/api/agent.go
@@ -96,11 +96,11 @@ func (a *api) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 	agentId := r.FormValue("id")
 	if val, ok := a.getResource(agentId); ok && val != nil {
 		if _, ok := val.(*loadtest.LoadTester); ok {
-			writeAgentResponse(w, http.StatusBadRequest, &client.AgentResponse{
+			writeAgentResponse(w, http.StatusConflict, &client.AgentResponse{
 				Error: fmt.Sprintf("load-test agent with id %s already exists", agentId),
 			})
 		} else {
-			writeAgentResponse(w, http.StatusBadRequest, &client.AgentResponse{
+			writeAgentResponse(w, http.StatusConflict, &client.AgentResponse{
 				Error: fmt.Sprintf("resource with id %s already exists", agentId),
 			})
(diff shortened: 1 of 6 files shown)


api/client_test.go
```diff
@@ -0,0 +1,274 @@
...
+	handler := SetupAPIRouter(logger.New(&logger.Settings{}), logger.New(&logger.Settings{}))
+
+	// run server using httptest
+	server := httptest.NewServer(handler)
+	defer server.Close()
+
+	var wg sync.WaitGroup
+
+	t.Run("create", func(t *testing.T) {
+		wg.Add(n)
```
agnivade, line 38: Would prefer this to be added at the start of the loop

api/client_test.go
```diff
@@ -0,0 +1,274 @@
...
+		st, err := agent.Run()
+		require.NoError(t, err)
+		require.Equal(t, loadtest.Running, st.State)
+		st, err = agent.AddUsers(n)
+		require.NoError(t, err)
+		require.Equal(t, int64(n), st.NumUsers)
+		wg.Add(n)
+		for i := 0; i < n; i++ {
+			if i%2 == 0 {
+				go func() {
```
agnivade, line 152: Wondering if we can move the goroutine a level up and do the `i%2` calculation inside it?

api/client_test.go
```diff
@@ -0,0 +1,274 @@
...
+		var ucConfig simulcontroller.Config
+		defaults.Set(&ltConfig)
+		defaults.Set(&ucConfig)
+		agent, err := agentClient.New(id, server.URL, nil)
+		require.NoError(t, err)
+		require.NotNil(t, agent)
+		for i := 0; i < n; i++ {
+			go func() {
+				if _, err := agent.Create(&ltConfig, &ucConfig); err == nil {
+					success += 1
```
agnivade, line 51: Are we relying on the correctness of `agent.Create` that multiple goroutines won't write to the success variable?

api/agent.go
```diff
@@ -116,7 +116,12 @@ func (a *api) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
-	a.setResource(agentId, lt)
+	if ok := a.setResource(agentId, lt); !ok {
+		writeAgentResponse(w, http.StatusBadRequest, &client.AgentResponse{
```
agnivade, line 120: Can we return a 409 conflict to make us look more classy? :)

streamer45: I am always in for some class :+1:

api/client_test.go
streamer45: Yes. I could have used atomics but thought that the race would happen only if the test failed so I kept it simple.
We can add a comment if it looks too sketchy :P

api/client_test.go
streamer45: If the end goal is using a single goroutine, yes, we can do that.

api/client_test.go
agnivade: Comment would be good. :+1:

api/client_test.go
agnivade: 0/5. This is fine too. Up to you.

PR #374, opened 2020-08-03
[MM-27469] Fix cases where gencontroller could get stuck

streamer45:
#### Summary

PR fixes a couple of cases where `gencontroller` (and the init process as a consequence) could get stuck without never exiting.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27469

### DIFF
--- a/loadtest/control/gencontroller/controller.go
+++ b/loadtest/control/gencontroller/controller.go
@@ -67,6 +67,13 @@ func (c *GenController) Run() {
 		c.sendStopStatus()
 	}()
 
+	done := func() bool {
+		return st.get("teams") >= c.config.NumTeams &&
+			st.get("channels") >= c.config.NumChannels &&
+			st.get("posts") >= c.config.NumPosts &&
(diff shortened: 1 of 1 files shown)


loadtest/control/gencontroller/controller.go
```diff
@@ -167,19 +172,19 @@ func (c *GenController) Run() {
 			return
 		}
 
-		if st.get("channels") == c.config.NumChannels {
+		if st.get("channels") >= c.config.NumChannels {
```
agnivade, line 175: Was it possible before to create more than one channel in a single action? Or this is just a defensive measure?

streamer45: We can guarantee to create at least the configured number of resources but it can happen to create more because we don't limit the concurrency of the API requests, we only check the state before and after, without locking for the whole thing.

agnivade: Oh my bad, state is shared globally amongst controllers. Yeah makes sense.

PR #375, opened 2020-08-10
[MM-27486] Add license validation check before deploying

streamer45:
#### Summary

PR adds a MM license validation check before creating a new deployment.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27486

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -47,10 +47,12 @@ func New(cfg *deployment.Config) *Terraform {
 
 // Create creates a new load test environment.
 func (t *Terraform) Create() error {
-	err := t.preFlightCheck()
-
-	if err != nil {
+	if err := t.preFlightCheck(); err != nil {
 		return err
+	}
+
+	if err := validateLicense(t.config.MattermostLicenseFile); err != nil {
+		return fmt.Errorf("license validation failed: %w", err)
 	}
 
 	extAgent, err := ssh.NewAgent()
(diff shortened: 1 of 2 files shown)


deployment/terraform/create.go
```diff
@@ -47,12 +47,14 @@ func New(cfg *deployment.Config) *Terraform {
 
 // Create creates a new load test environment.
 func (t *Terraform) Create() error {
-	err := t.preFlightCheck()
-
-	if err != nil {
+	if err := t.preFlightCheck(); err != nil {
```
streamer45, line 50: :eyeglasses:

deployment/terraform/utils.go
```diff
@@ -114,3 +117,30 @@ func openBrowser(url string) (err error) {
 	}
 	return
 }
+
+func validateLicense(filename string) error {
```
agnivade, line 121: This sounds like a good candidate for a unit test?

streamer45: I believe most of the functionality is already covered by the tests in the server repository.
We could add a unit test but it would require including sample license files.

agnivade: Yeah, I think I would be fine with that. I mainly wanted to test the `!IsStarted` and `IsExpired` part, and verify that we are indeed returning errors for them. They are individually tested in the server repo, but this additional logic isn't tested.

PR #376, opened 2020-08-11
[MM-27330] Add some missing search related actions

streamer45:
#### Summary

PR adds some missing search related actions to `SimulController`.

#### Ticket 

https://mattermost.atlassian.net/browse/MM-27330

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -902,8 +902,14 @@ func searchPosts(u user.User) control.UserActionResponse {
 		opts.In = channel.Name
 	}
 
+	timeOpts := []*time.Time{
+		&opts.On,
+		&opts.Before,
+		&opts.After,
+	}
 	if rand.Float64() < 0.2 {
-		opts.On = time.Now()
+		to := timeOpts[rand.Intn(len(timeOpts))]
+		*to = time.Now().Add(-time.Duration(rand.Intn(7)) * time.Hour * 24)
 	}
 
 	if rand.Float64() < 0.2 {
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -870,3 +870,95 @@ func searchChannels(u user.User) control.UserActionResponse {
 		return control.UserActionResponse{Info: fmt.Sprintf("found %d channels", len(channels))}
 	})
 }
+
+func searchPosts(u user.User) control.UserActionResponse {
+	team, err := u.Store().CurrentTeam()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	} else if team == nil {
+		return control.UserActionResponse{Err: control.NewUserError(errors.New("current team should be set"))}
+	}
+
+	var words []string
+	var opts control.PostsSearchOpts
+	// This is an arbitrary limit on the number of words to search for.
+	// TODO: possibly use user analytics data to improve this.
+	count := 1 + rand.Intn(4)
+
+	// TODO: back the probability of these choices with real data.
+	if rand.Float64() < 0.2 {
+		user, err := u.Store().RandomUser()
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+		opts.From = user.Username
+	}
+
+	if rand.Float64() < 0.2 {
+		channel, err := u.Store().RandomChannel(team.Id, store.SelectMemberOf|store.SelectNotDirect|store.SelectNotGroup)
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+		opts.In = channel.Name
+	}
+
+	if rand.Float64() < 0.2 {
+		opts.On = time.Now()
+	}
+
+	if rand.Float64() < 0.2 {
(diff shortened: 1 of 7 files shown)


loadtest/control/simulcontroller/actions.go
```diff
@@ -870,3 +870,95 @@ func searchChannels(u user.User) control.UserActionResponse {
...
+	if rand.Float64() < 0.2 {
+		channel, err := u.Store().RandomChannel(team.Id, store.SelectMemberOf|store.SelectNotDirect|store.SelectNotGroup)
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+		opts.In = channel.Name
+	}
+
+	if rand.Float64() < 0.2 {
+		opts.On = time.Now()
```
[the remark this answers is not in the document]
streamer45: Yes. Let me update.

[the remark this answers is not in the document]
streamer45: Regarding the older date. The resolution of the search appears to be a day so unless the load-test has been running for a long time it shouldn't make much of a difference.

loadtest/control/simulcontroller/actions.go
```diff
@@ -870,3 +870,101 @@ func searchChannels(u user.User) control.UserActionResponse {
...
+	})
+}
+
+func searchGroupChannels(u user.User) control.UserActionResponse {
+	user, err := u.Store().RandomUser()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.EmulateUserTyping(user.Username[:1+rand.Intn(4)], func(term string) control.UserActionResponse {
```
agnivade, line 961: Let's add the comment about 4 here also for clarity

loadtest/control/simulcontroller/actions.go
```diff
@@ -870,3 +870,101 @@ func searchChannels(u user.User) control.UserActionResponse {
...
+func searchUsers(u user.User) control.UserActionResponse {
+	user, err := u.Store().RandomUser()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.EmulateUserTyping(user.Username, func(term string) control.UserActionResponse {
+		users, err := u.SearchUsers(&model.UserSearch{
+			Term:  term,
+			Limit: 100,
```
agnivade, line 946: Is this the limit used by the webapp?

loadtest/control/simulcontroller/actions.go
```diff
@@ -870,3 +870,101 @@ func searchChannels(u user.User) control.UserActionResponse {
...
+		return control.UserActionResponse{Err: control.NewUserError(errors.New("current team should be set"))}
+	}
+
+	var words []string
+	var opts control.PostsSearchOpts
+	// This is an arbitrary limit on the number of words to search for.
+	// TODO: possibly use user analytics data to improve this.
+	count := 1 + rand.Intn(4)
+
+	// TODO: back the probability of these choices with real data.
```
agnivade, line 888: Let's file a ticket for this? Now that you have access, it shouldn't be very hard to get this.

loadtest/control/simulcontroller/actions.go
```diff
@@ -870,3 +870,101 @@ func searchChannels(u user.User) control.UserActionResponse {
...
+	}
+
+	timeOpts := []*time.Time{
+		&opts.On,
+		&opts.Before,
+		&opts.After,
+	}
+	if rand.Float64() < 0.2 {
+		to := timeOpts[rand.Intn(len(timeOpts))]
+		*to = time.Now().Add(-time.Duration(rand.Intn(7)) * time.Hour * 24)
```
agnivade, line 912: Why 7?

And also, this seems like a roundabout way to do things. Is there a reason we cannot just get a random number between 0 to 3 and update the right variable accordingly?

loadtest/control/simulcontroller/actions.go
streamer45: I am not sure we collect this kind of data. To improve this we would need to store the actual text content of what the users are searching.
Were you thinking of something else?

loadtest/control/simulcontroller/actions.go
streamer45: > Why 7?

It's an arbitrary value like everything else in this kind of methods. I chose to go back not more than a week. hence the 7. This was to address @isacikgoz's review.

> And also, this seems like a roundabout way to do things. Is there a reason we cannot just get a random number between 0 to 3 and update the right variable accordingly?

It's the first thing I thought of. Maybe it was the [C](https://en.wikipedia.org/wiki/C_(programming_language)) coder inside me. I *simplified*.

loadtest/control/simulcontroller/actions.go
streamer45: Yes.

streamer45:
> One concern I had is that we are not emulating the autocomplete of users and channels when the user is actually typing the "in:" or "from:" string. That also generates some autocomplete queries which we should do. Perhaps a separate ticket is appropriate for this?

That's a good point. I added some logic to make it a bit more realistic.

loadtest/control/simulcontroller/actions.go
agnivade: Oh yeah you're right. Nvm then.

PR #377, opened 2020-08-14
[MM-27510] Disable server metrics during init process

streamer45:
#### Summary

PR fixes a problem with the init process causing false positives on first coordinator's run.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27510

### DIFF
--- a/cmd/ltagent/init.go
+++ b/cmd/ltagent/init.go
@@ -18,6 +18,7 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/logger"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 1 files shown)


agnivade:
I am trying to understand this better. How is the coordinator running during the init process? Doesn't the coordinator start only when we start a load test? Or you mean the init process already affects the past data for the coordinator when it actually runs because it queries a given interval and thereby feeds it incorrect data?

I am wondering if we can leverage the coordinator start time for this and then tweak the time interval queries not to go past the start time. Disabling and enabling metrics seems like a roundabout way to achieve this. Our objective is to not query for data beyond the time which we are interested, so rather than not generating data, can we not query for data in the incorrect time range?

streamer45:
> Or you mean the init process already affects the past data for the coordinator when it actually runs because it queries a given interval and thereby feeds it incorrect data?

What happens is that the init process can really raise the average API times since it essentially bombards the server with requests which in turn can affect the results of the queries the coordinator runs against Prometheus causing it to decrease users sooner.

> I am wondering if we can leverage the coordinator start time for this and then tweak the time interval queries not to go past the start time. Disabling and enabling metrics seems like a roundabout way to achieve this. Our objective is to not query for data beyond the time which we are interested, so rather than not generating data, can we not query for data in the incorrect time range?

I am not sure. I think that collecting metrics during the init process is not something we should be doing anyway so this seems to me like the most straightforward solution to the problem.

agnivade:
> What happens is that the init process can really raise the average API times since it essentially bombards the server with requests which in turn can affect the results of the queries the coordinator runs against Prometheus causing it to decrease users sooner.

To clarify, are we running the coordinator during the init process or after (which means that historical data in the range query is what is causing the problem)?

streamer45:
> > What happens is that the init process can really raise the average API times since it essentially bombards the server with requests which in turn can affect the results of the queries the coordinator runs against Prometheus causing it to decrease users sooner.
> 
> To clarify, are we running the coordinator during the init process or after (which means that historical data in the range query is what is causing the problem)?

The latter. The init process does not involve the coordinator. It's a `GenController` with a fixed number of users.

agnivade:
Correct, so can we tweak the query interval not to go past the coordinator start time? Or do you think it's too much effort?

streamer45:
> Correct, so can we tweak the query interval not to go past the coordinator start time? Or do you think it's too much effort?

I can take a look but I have a feeling it's more complex than just fixing the query interval. I remember cases where the coordinator was launched several minutes after the init process had finished and still suffered from this cause of the way Prometheus was averaging the request times.

agnivade:
Hmm.. alright then.

streamer45:
I am going to close this for now. Hopefully https://github.com/mattermost/mattermost-load-test-ng/pull/378 is enough to fix the issue.

PR #378, opened 2020-08-17
[MM-27906] Fix inconsistencies when querying metrics

streamer45:
#### Summary

PR is an alternative approach to fix the issue that https://github.com/mattermost/mattermost-load-test-ng/pull/377 was addressing.

We introduce a new query specific config value `MinIntervalSec` that defines the minimum amount of time (in seconds) that needs to have passed before attempting to run the query.

#### Ticket

https://mattermost.atlassian.net/browse/MM-27906


isacikgoz (approved):
LGTM 🏖️

PR #381, opened 2020-09-09
[MM-28497] Avoid failing if user mention cannot be autocompleted

streamer45:
#### Summary

PR fixes `SimulController.createMessage()` so that it won't fail if user mention could not be constructed.
We should still support posts with mentions of users not in the channel.

#### Ticket

https://mattermost.atlassian.net/browse/MM-28497


isacikgoz (approved):
Looks good 👍

PR #382, opened 2020-09-14
Releasing v1.1.0

streamer45:


agnivade (approved):
Wondering if we should create a script for this.

streamer45:
> Wondering if we should create a script for this.

Yes, I think it makes sense to do so.

isacikgoz (approved):
🎉

PR #383, opened 2020-09-16
Fix incorrect link

agnivade:
Not sure how this worked earlier. Probably GitHub changed their URL scheme.


streamer45:
I think this was goreleaser not adding the `v`. Let me fix the actual file so we can keep the URL as is.

streamer45:
It should work now.

agnivade:
Cool, thanks.

PR #386, opened 2020-09-28
[MM-29006] Allow starting load-test on agents only deployments

streamer45:
#### Summary

PR fixes starting a load-test on agents only deployments.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29006

### DIFF
--- a/deployment/terraform/coordinator.go
+++ b/deployment/terraform/coordinator.go
@@ -10,6 +10,7 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/coordinator"
 	"github.com/mattermost/mattermost-load-test-ng/coordinator/cluster"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest"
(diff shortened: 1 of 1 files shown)


deployment/terraform/coordinator.go
```diff
@@ -76,9 +73,14 @@ func (t *Terraform) StartCoordinator() error {
 
 	mlog.Info("Uploading other load-test config files")
 
-	agentConfig, err := t.generateLoadtestAgentConfig(output)
+	var agentConfig *loadtest.Config
+	if len(output.Instances) > 0 {
+		agentConfig, err = t.generateLoadtestAgentConfig(output)
```
agnivade, line 78: Do we just want to pass a boolean to the method to just return after doing `loadtest.ReadConfig("")`? It also does the same thing.

streamer45: I think both approaches are valid but I feel that being explicit at the cost of some minor repetition makes the code easier to follow. We either generate the config or not. I considered both options and thought that adding more state to the helper function just to avoid a couple of lines was not worth it.

agnivade (approved):

isacikgoz (approved):
Nice 👍

PR #387, opened 2020-10-01
Update deployer_config.md

reflog:
#### Summary

Added some FAQ entries

### DIFF
--- a/docs/deployer_config.md
+++ b/docs/deployer_config.md
@@ -242,3 +242,19 @@ A friendly name for the graph.
 *string*
 
 The Prometheus query to run.
+
+### FAQ
+
+##### Providing custom build of mattermost binary doesn't work.
+If you are specifying a custom Mattermost build, you might get the following message:
+```
+2020-10-01T11:23:42.484+0300    info    terraform/create.go:441 Waiting for the server...
+Error: error whiling pinging server: timeout after 30 seconds, server is not responding
+```
+If this happens, disable CGO and rebuild `mattermost-server` like so:
+```
+CGO_ENABLED=0 make build
+```
+
+##### Load test doesn't stop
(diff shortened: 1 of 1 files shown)


agnivade (commented):
I think a separate FAQ page with a link from an existing page might be better.

And as discussed, we can remove the note about glibc. The error message of failure to ping server can arise from other reasons too. The user can always ssh into the server and look into the logs.

docs/deployer_config.md
```diff
@@ -242,3 +242,19 @@ A friendly name for the graph.
...
+2020-10-01T11:23:42.484+0300    info    terraform/create.go:441 Waiting for the server...
+Error: error whiling pinging server: timeout after 30 seconds, server is not responding
+```
+If this happens, disable CGO and rebuild `mattermost-server` like so:
+```
+CGO_ENABLED=0 make build
+```
+
+##### Load test doesn't stop
+Make sure that the number of active users is set up correctly, it should look like: `UsersConfiguration.MaxActiveUsers = AgentInstanceCount * 1000`
```
streamer45, line 260: We discussed this briefly in our conversation. In your case I think the problem was that the `coordinator` was not converging to a number because it was limited at `2000` users which wasn't enough to trigger a performance degradation event.

The rule of thumb here is that when starting a load-test we should always shoot for more users than what we think an installation can support. 

`UsersConfiguration.MaxActiveUsers` should be `AgentInstanceCount * 2000`. I think we should update `coordinator.sample.json` to reflect this do that at least we can comfortably cover the default case.

agnivade:
Hey @reflog - just checking on this one :)

reflog:
@agnivade - I don't feel like I'm the right person to update this document, so I'm going to close this PR.

streamer45:
No worries Eli, I can take care of this. Created [MM-30788](https://mattermost.atlassian.net/browse/MM-30788).

PR #388, opened 2020-10-05
[MM-29276] Set correct error count when done

streamer45:
#### Summary

PR fixes an issue with the coordinator failing to set the total errors count once done.
This would result in `ltctl loadtest status` showing zero errors at the end of a load-test.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29276

### DIFF
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -87,12 +87,19 @@ func (c *Coordinator) Run() (<-chan struct{}, error) {
 
 		defer func() {
 			c.monitor.Stop()
+			clusterStatus, err := c.cluster.Status()
+			if err != nil {
+				c.log.Error("coordinator: cluster status error:", mlog.Err(err))
+			}
 			c.cluster.Shutdown()
 			close(c.doneChan)
 			c.mut.Lock()
 			c.status.State = Done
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
LGTM 👍

coordinator/coordinator.go
```diff
@@ -87,12 +87,19 @@ func (c *Coordinator) Run() (<-chan struct{}, error) {
...
+			if err != nil {
+				c.log.Error("coordinator: cluster status error:", mlog.Err(err))
+			}
 			c.cluster.Shutdown()
 			close(c.doneChan)
 			c.mut.Lock()
 			c.status.State = Done
 			c.status.SupportedUsers = supported
 			c.status.StopTime = time.Now()
+			if clusterStatus.NumErrors > 0 {
```
agnivade, line 100: Is this check even needed?

agnivade (approved):

coordinator/coordinator.go
streamer45: I've added it because `cluster.Status()` doesn't really return an error but in case of one it can return a zero status value. I wanted to avoid a case where we might be overwriting an existing value for `c.status.NumErrors`.

PR #397, opened 2020-11-11
Remove final logout on SimulController stop

streamer45:
#### Summary

We remove the final `logout` action when stopping a `SimulController`. 

This was causing an enormous spike in server response times which was skewing metrics and making them less readable overall.

It was originally introduced to avoid keeping many old sessions in the DB but that's not really an issue anymore since we end up resetting the DB during comparisons.


isacikgoz (approved):
Looks good 👍

agnivade (approved):

PR #400, opened 2020-11-13
[MM-30437] Add Grafana panel for store times

streamer45:
#### Summary

PR adds the important (but missing) panel for store call times (avg and p99).

#### Ticket

https://mattermost.atlassian.net/browse/MM-30437


agnivade (approved):
IIRC, we had this originally, but then removed it later?

streamer45:
> IIRC, we had this originally, but then removed it later?

Yes, I recall having store times (no p99) in the original dashboard but it was way before we made real use of it in comparisons and such.

PR #403, opened 2020-11-18
[MM-30332] Add connected users count to coordinator status output

streamer45:
#### Summary

PR adds a connected users count to the output of `ltctl loadtest status`. This can further help to identify possible issues with the load-test.

#### Ticket

https://mattermost.atlassian.net/browse/MM-30332

### DIFF
--- a/cmd/ltctl/loadtest.go
+++ b/cmd/ltctl/loadtest.go
@@ -15,6 +15,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
+func getUsersCount(helper *prometheus.Helper) (int, error) {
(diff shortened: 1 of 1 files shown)


agnivade (approved):
Passing a separate parameter feels a bit weird. But at the same time, I do see it doesn't make sense to make it part of coordinator.Status.

streamer45:
> Passing a separate parameter feels a bit weird. But at the same time, I do see it doesn't make sense to make it part of coordinator.Status.

Yes, this is quickly becoming more like "show load-test stats" rather than coordinator's status. I think it's an acceptable patch for now but will think about ways to making this look (and feel) better.

PR #405, opened 2020-11-23
[MM-29243] Add some info on commonly tested setups

streamer45:
#### Summary

PR adds a document with some information about commonly tested setups, results and some important notes and gotchas.
As far as results, I've only listed the unbounded ones as they are easily expressible with a number.

In parallel to this my plan is to work on storing comparison results in a machine friendly format (JSON or similar) and start building an archive from which we could derive, generate and automatically keep up to date more documents like this.

As usual, let me know of anything you'd like to be added to this.

#### Ticket

https://mattermost.atlassian.net/browse/MM-29243

### DIFF
--- a/docs/setups.md
+++ b/docs/setups.md
@@ -1,0 +1,42 @@
+## Commonly tested setups
+
+### Single App cluster
+
+| Name    | Type        | Specs                | Amount |
+|---------|-------------|----------------------|--------|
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
Very nice! 💪

agnivade (approved):

PR #412, opened 2020-12-11
Add message export on cluster controller

nronas:
#### Summary
As per https://mattermost.atlassian.net/browse/MM-10669 adds message export on loadtest agent.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-10669


loadtest/user/userentity/actions.go
```diff
@@ -1091,3 +1091,16 @@ func (ue *UserEntity) UpdateConfig(cfg *model.Config) error {
 	ue.store.SetConfig(cfg)
 	return nil
 }
+
+// MessageExport creates a job for a compliance message export
+func (ue *UserEntity) MessageExport() error {
+	messageExportJob := &model.Job{
+		Type: "message_export",
```
agnivade, line 1098: Do we want to set some data attributes? Or just go with the defaults?

agnivade (commented):
Thanks for adding this!

I think we should add it to the `control` package and make it a common method, invokable from simulcontroller and simplecontroller as well. We can remove it from clustercontroller. It was a one-time thing written by me to only test HA related functionalities. Message export is not an HA functionality and needs to be available on other controllers too.

loadtest/user/userentity/actions.go
nronas: This is basically mimicking the actual API call that the webapp is doing when triggering a message export. So it only creates a new message export job with the `Data: nil`

nronas:
> I think we should add it to the control package and make it a common method, invokable from simulcontroller and simplecontroller as well. We can remove it from clustercontroller. It was a one-time thing written by me to only test HA related functionalities. Message export is not an HA functionality and needs to be available on other controllers too.

That was my initial thinking as well, but semantically the following things lead me to not do that:

1) Message Export is only available to users that are admins, the simulation controller is doing a probabilistic user action selection for each user. The profiling of message export is only meaningful if there is an existing load in the system and then an admin triggers a message export since it relies on the amount of non previously exported posts.

2) Ideal load testing scenario for message export is IMO, generate a bunch of messages in many channels and DMs, then triggers an one of message export.

Happy to make the `MessageExport` available to control but I am not sure how valuable that is from a load testing perspective.

Let me know what you think.

loadtest/user/userentity/actions.go
agnivade: Perfect. :+1:

agnivade:
Makes sense to me. In that case, I would suggest to make it part of simplecontroller. That way, we can specifically create a controller to trigger the message export however frequently we want it to, concurrently with other controllers like simul/cluster running with it. Then we can wait for the system to generate a considerable amount of messages, and then add a simple controller run, trigger the export, and stop it.

Basically, it seems like it has a better fit towards a standalone action that can be triggered as and when necessary, rather than part of a user action selection, which is what simulcontroller does.

nronas:
> Makes sense to me. In that case, I would suggest to make it part of simplecontroller. That way, we can specifically create a controller to trigger the message export however frequently we want it to, concurrently with other controllers like simul/cluster running with it. Then we can wait for the system to generate a considerable amount of messages, and then add a simple controller run, trigger the export, and stop it.

Totally agree. I will move it to the simple controller, my understanding is that still, a simple controller cannot guarantee that the user is admin right?

nronas:
I will try to allow access to the admin within the action, with this we will be able to invoke admin allowed actions.

agnivade:
Yeah it cannot guarantee that. I think we have to work around that with some approach like you mentioned above.

streamer45:
We already have a case where we run something only if the user is admin. There's a `user.IsSysAdmin()` method for that.
What is missing is logging the sysadmin. That should be easy to do from `NewControllerWrapper()`.

nronas:
@streamer45 yeap currently the `Reload` check if the user is a sysadmin, so what I am going to do is I will create a method called `addAdmin` and use it right after the `addUsers` method. Then we can guarantee that at least one user(the sysadmin) is logged in and can perform actions that need admin privileges. How that sounds?

streamer45:
> @streamer45 yeap currently the `Reload` check if the user is a sysadmin, so what I am thinking of doing is I will create a method called `addAdmin` and use it right after the `addUsers` in the `loadtest.Run()`. Then we can guarantee that at least one user(the sysadmin) is logged in and can perform actions that need admin privileges. How that sound?

Not sure that's going to work. The `loadtest` package doesn't directly deal with users. It's completely decoupled. I think it would be easier to do it at the higher level where we define how those users really look like.

nronas:
aha i see what you are saying, I think :D. Let me see what I can do.

nronas:
@agnivade @streamer45 let me know what you think https://github.com/mattermost/mattermost-load-test-ng/pull/412/commits/eb37a791822409700c330f1767a5a6a54160b2df

loadtest/control/simplecontroller/controller.go
```diff
@@ -86,6 +88,16 @@ func (c *SimpleController) Run() {
...
+		{
+			run: func(user.User) control.UserActionResponse {
+				resp := control.Login(c.admin)
+				if resp.Err != nil {
+					return resp
+				}
+				c.connect()
+				return resp
+			},
+		},
```
streamer45, line 100: This is going to run for each user, so we essentially end up logging in the admin N times. I don't think we should change the controller since any user is equal from its point of view. What I would suggest is to do something even easier like making the first created user an admin in `NewControllerWrapper()`.

nronas: That will work as well, my thinking is that each user basically can perform admin actions, so we have to log admin in N times, but you are right essentially that admin is a single user. I will update the controller to basically make the first user to be an admin

streamer45: Yes, the problem is that the controller should really be responsible for one user session only otherwise we end up making things more complex and potentially less realistic. 

I apologize for the confusion, it's totally okay to simply add a check on creation and make the first controlled user an admin.

nronas: i think that should do it 😄 https://github.com/mattermost/mattermost-load-test-ng/pull/412/commits/b3acb8507b91a5ee66402bc4c21e23f536257f64

loadtest/loadtest.go
```diff
@@ -72,7 +72,7 @@ func (lt *LoadTester) AddUsers(numUsers int) (int, error) {
 		return 0, ErrNotRunning
 	}
 	for i := 0; i < numUsers; i++ {
-		if err := lt.addUser(); err != nil {
+		if err := lt.addUser(i); err != nil {
```
streamer45, line 75: This would result in starting an admin every time we add new users to the load-test.

api/agent.go
```diff
@@ -353,7 +353,13 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
 			ueSetup.Metrics = metrics.UserEntityMetrics()
 		}
-		ue := userentity.New(ueSetup, ueConfig)
+		var ue *userentity.UserEntity
+		if index == 0 {
+			// We want a user to be admin in order to be able to perform admin allowed actions
+			ue = admin(ueSetup, ueConfig, config.ConnectionConfiguration.AdminEmail, config.ConnectionConfiguration.AdminPassword)
+		} else {
+			ue = userentity.New(ueSetup, ueConfig)
+		}
```
streamer45, line 362: Why the additional parameter? I thought adding a check on `id == 0` would be sufficient.

This problem should be really dealt with during data initialization by making some (configurable) percentage of the users to be admins. We can do this at a later time of course. I am okay with the quick fix here.

nronas: I thought the userId is not guaranteed to start from 0, since we randomize but I see that we do that only after the first user has been initialized: https://github.com/mattermost/mattermost-load-test-ng/blob/master/loadtest/loadtest.go#L99.

> This problem should be really dealt with during data initialization by making some (configurable) percentage of the users to be admins. We can do this at a later time of course. I am okay with the quick fix here.

Completely agree with this, i am ok with proceeding with that approach here.

streamer45: > I thought the userId is not guaranteed to start from 0, since we randomize but I see that we do that only after the first user has been initialized: https://github.com/mattermost/mattermost-load-test-ng/blob/master/loadtest/loadtest.go#L99.

Yes, also that's only done if we allow multiple sessions per user, which is not usually the case.

> Completely agree with this, i am ok with proceeding with that approach here.

That would be even better :+1:

nronas: Sorry i focused yesterday on the message export work, have a look on https://github.com/mattermost/mattermost-load-test-ng/pull/412/commits/f78df399e6b31e0c4d3e30f03f68a91878ef8c41 , the `loadtest` module itself kknows about the index of the user that is adding, hence having all the `addUser` and `addUser(s)` methods init. So I figured a deterministic way to identify if we need an admin is to pass bool argument from loadtest to agent. 

The addUsers now is asymmetric with the removeUsers happy to make removeUsers to remove the right % of admin vs non-admin users

loadtest/control/actions.go
```diff
@@ -729,3 +729,33 @@ func Reload(u user.User) UserActionResponse {
...
+
+	if err := u.GetConfig(); err != nil {
+		return UserActionResponse{Err: NewUserError(err)}
+	}
+
+	cfg := u.Store().Config()
+
+	if !*cfg.MessageExportSettings.EnableExport {
+		return UserActionResponse{Info: "message export is not enabled"}
+	}
```
streamer45, line 753: We should probably make sure the pointer is not nil. I've seen too many crashes due to missing config settings.

cmd/ltagent/init.go
```diff
@@ -110,23 +119,43 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
 
 	config.UserControllerConfiguration.Type = loadtest.UserControllerGenerative
 	config.UsersConfiguration.InitialActiveUsers = 0
+	config.UsersConfiguration.PercentageOfAdminUsers = adminPercent
```
streamer45, line 122: I think this setting should be in `InstanceConfiguration` since that's what the init process deals with.
Also, for our use case, we might want to drop the percentage and go for a specific number since it might get confusing otherwise (percentage of what value? Initial users, max users, some hard-coded constant).

We also might want to overwrite the config value only if `adminPercent` was actually passed through the command line.

cmd/ltagent/init.go
```diff
@@ -110,23 +119,43 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+		lt, err := loadtest.New(config, newC, logger)
+		if err != nil {
+			return fmt.Errorf("error while initializing loadtest: %w", err)
+		}
+
+		mlog.Info(fmt.Sprintf("generating %d users with prefix %s", numUsers, userPrefix))
+		err = genData(lt, numUsers)
+		if err != nil {
+			return err
+		}
```
streamer45, line 155: I am not sure this is going to work as we expect it. We shouldn't be running the generative load-test more than once.
I think for this case we might want to keep it very simple and before even running this make sure the configured number of admins gets created without the need to going through the load-test. Just creating some user entities and promoting them in a loop will do.

loadtest/control/actions.go
nronas: Yeap, good point

cmd/ltagent/init.go
nronas: Makes sense, will amend

cmd/ltagent/init.go
nronas: I've tried to keep things as homogenous as possible. Happy to have the admins being generated up front. Personally, I do not see a problem with running generative loadTest instances since the init phase is a one-off operation, no strong pref though.

streamer45: The main issue I see doing this is that the admin users will end up generating all the data. The way the gencontroller works is that it's not going to stop until all data has been generated.

nronas: Yeah true, which in a sense is ok right, since essentially what are we doing above is splitting the genData onto two segments, one with N users and one with M admins but our intention still remains to generate N+M users.

streamer45: The point is generating the data (as in teams, channels, posts, reactions), not the users. So the second pass would not really be needed if the admins already created everything. But also there was a reason those numbers were chosen (50 at 0.2 rate). Those values gave the highest throughput.

nronas: Right, I didn't know about the second reason. Cool, that makes sense. Will amend.

loadtest/control/actions.go
```diff
@@ -729,3 +729,33 @@ func Reload(u user.User) UserActionResponse {
...
+// a compliance message export
+func MessageExport(u user.User) UserActionResponse {
+	isAdmin, err := u.IsSysAdmin()
+	if err != nil {
+		return UserActionResponse{Err: NewUserError(err)}
+	}
+
+	if !isAdmin {
+		return UserActionResponse{Info: "user is not a sysadmin and cannot perform a message export"}
+	}
```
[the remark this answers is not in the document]
nronas: Every user is calling this action but only users that are admins will proceed with actually triggering a message export. We have a similar check in the `Reload` action. Happy to bubble this up to the actual action invocation.

loadtest/user/userentity/user.go
```diff
@@ -199,6 +201,56 @@ func (ue *UserEntity) IsTeamAdmin() (bool, error) {
...
+func (ue *UserEntity) PromoteToAdmin(u *UserEntity) error {
+	if err := ue.Login(); err != nil {
+		return err
+	}
+
+	isAdmin, err := ue.IsSysAdmin()
+	if err != nil {
+		return err
+	}
+	if !isAdmin {
```
nronas, line 214: This is here for a sanity check, in the case that this func is called from a non-admin user.

loadtest/control/actions.go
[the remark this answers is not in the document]
streamer45: The difference is that the Reload case is somewhat optional, here the whole action needs admin permission. I am mainly trying to avoid a spam of logs otherwise it's the same really.

loadtest/control/actions.go
```diff
@@ -729,3 +729,33 @@ func Reload(u user.User) UserActionResponse {
...
+
+	if err := u.GetConfig(); err != nil {
+		return UserActionResponse{Err: NewUserError(err)}
+	}
+
+	cfg := u.Store().Config()
+
+	if &cfg != nil && !*cfg.MessageExportSettings.EnableExport {
```
streamer45, line 751: I actually meant `cfg.MessageExportSettings.EnableExport != nil`. We use pointers for the config values (not ideal but that comes from the server).
`cfg` itself is safe since we return a value.

cmd/ltagent/init.go
```diff
@@ -126,9 +129,98 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+			KeepAlive: 30 * time.Second,
+		}).DialContext,
+		MaxConnsPerHost:       500,
+		MaxIdleConns:          500,
+		MaxIdleConnsPerHost:   500,
+		ResponseHeaderTimeout: 5 * time.Second,
+		IdleConnTimeout:       90 * time.Second,
+		TLSHandshakeTimeout:   1 * time.Second,
+		ExpectContinueTimeout: 1 * time.Second,
+	}
```
streamer45, line 157: I am thinking we could use the default transport for this. We are doing sequential requests anyway so it should work just fine.

cmd/ltagent/init.go
```diff
@@ -126,9 +129,98 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+	adminStore, err := memstore.New(&memstore.Config{
+		MaxStoredPosts:          1,
+		MaxStoredUsers:          1,
+		MaxStoredChannelMembers: 1,
+		MaxStoredStatuses:       1,
+	})
```
streamer45, line 163: We can pass `nil` to `memstore.New()` and it will use some internal defaults.

cmd/ltagent/init.go
```diff
@@ -126,9 +129,98 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+	}
+	sysadmin := userentity.New(adminUeSetup, adminUeConfig)
+
+	for i := 0; i < int(config.InstanceConfiguration.NumAdmins); i++ {
+		userStore, err := memstore.New(&memstore.Config{
+			MaxStoredPosts:          1,
+			MaxStoredUsers:          1,
+			MaxStoredChannelMembers: 1,
+			MaxStoredStatuses:       1,
+		})
```
streamer45, line 186: Same thing.

loadtest/config.go
```diff
@@ -88,6 +88,8 @@ type InstanceConfiguration struct {
 	NumPosts int64 `default:"0" validate:"range:[0,]"`
 	// The target number of reactions to be created.
 	NumReactions int64 `default:"0" validate:"range:[0,]"`
+	// The target number of admin users to be created.
+	NumAdmins int64 `default:"10" validate:"range:[0,]"`
```
streamer45, line 92: This should probably be the same value we set in the `sample.json` file. So I'd say 0 should be the default.

config/config.sample.json
```diff
@@ -31,6 +31,7 @@
     "NumChannels": 10,
     "NumPosts": 0,
     "NumReactions": 0,
+    "NumAdmins": 0,
```
streamer45, line 34: Let's also update `docs/loadtest_config.md` with the added setting.

streamer45 (commented):
Thank you for the hard work on this. I've left a few minor comments.

loadtest/user/userentity/user.go
```diff
@@ -199,6 +201,56 @@ func (ue *UserEntity) IsTeamAdmin() (bool, error) {
 	return user.IsInRole(model.TEAM_ADMIN_ROLE_ID), nil
 }
 
+// PromoteToAdmin promotes the given user to a sysadmin role
+func (ue *UserEntity) PromoteToAdmin(u *UserEntity) error {
```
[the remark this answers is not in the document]
nronas: Sounds good to me!

cmd/ltagent/init.go
nronas: Awesome, TIL

nronas:
Everything is amended on https://github.com/mattermost/mattermost-load-test-ng/pull/412/commits/313b9c91b38289a4597d33be754d851effbf10d4

docs/loadtest_config.md
```diff
@@ -74,6 +74,34 @@ This value overrides the actual server version. If left empty, the one returned
 The number of teams the target Mattermost instance should have.  
 These will be created during the `init` process.
 
+### NumChannels
```
streamer45, line 77: Thank you :+1: I didn't realize it was missing all that.

cmd/ltagent/init.go
```diff
@@ -126,9 +129,95 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+		err = loadtest.PromoteToAdmin(sysadmin, userentity.New(userSetup, ueConfig))
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func createUser(sysadmin *userentity.UserEntity, user *model.User) (string, error) {
```
streamer45, line 200: Is this needed? The init should exit in the presence of conflicts and otherwise be run on an empty state. Were you having some issues on re-runs?

cmd/ltagent/init.go
```diff
@@ -126,9 +129,95 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+	userId, err := sysadmin.CreateUser(user)
+	if err != nil {
```
streamer45, line 202: I'd probably return early in case of no error to avoid one indentation level.

streamer45 (approved):
Looks great :+1: , left non-blocking comments.

cmd/ltagent/init.go
nronas: oh i see, i thought that there can be a scenario that we run it on a dirty state. Will defo remove to reduce complexityl

streamer45: The only way to do that would be to change user prefixes but it's only done for local testing. We can work on the assumption that it will always run on a fresh state.

agnivade (commented):
Wow this PR became way bigger in scope than originally thought. Don't want to hold it back any longer. Just a small query.

cmd/ltagent/init.go
```diff
@@ -126,9 +128,74 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
...
+
+		user.Id = userId
+		err = userStore.SetUser(user)
+		if err != nil {
+			return err
+		}
+		userSetup := userentity.Setup{
+			Store: userStore,
+		}
+		err = loadtest.PromoteToAdmin(sysadmin, userentity.New(userSetup, ueConfig))
```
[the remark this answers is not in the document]
nronas: Defo a good point, i can pass a "logged in" admin for sure. The hypothesis here is that admin promotions will be reasonably fewer than normal user creation.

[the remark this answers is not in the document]
agnivade: Sure, I will leave it upto you. Not a big deal either way. Can even be a HW ticket.

agnivade (approved):

PR #413, opened 2020-12-15
MM-31367: Support custom credential file

agnivade:
This extends the load-test functionality to read
user credentials from an external file and then
use them in the load-test.

This is helpful to simulate LDAP users which weren't possible
in the existing load-test framework

https://mattermost.atlassian.net/browse/MM-31367

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -333,23 +333,9 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 		}
 	}
 
-	var creds []string
-	f, err := os.Open(config.UsersConfiguration.UsersFilePath)
-	if err == nil {
-		// The file is an optional parameter which may not be present.
-		// So we read it only if it opens successfully.
-		defer f.Close()
-		scanner := bufio.NewScanner(f)
-		for scanner.Scan() {
-			creds = append(creds, scanner.Text())
-		}
-		if err := scanner.Err(); err != nil {
-			return nil, fmt.Errorf("failed to read from %s: %w", f.Name(), err)
-		}
-	} else if !os.IsNotExist(err) {
-		// If the file exists, but failed to open, then probably it is worthwhile
-		// logging it.
-		mlog.Warn("Failed to open UsersFilePath. Continuing with default signup", mlog.Err(err))
+	creds, err := getUserCredentials(config.UsersConfiguration.UsersFilePath, config)
+	if err != nil {
+		return nil, err
 	}
 
 	return func(id int, status chan<- control.UserStatus) (control.UserController, error) {
@@ -358,12 +344,6 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 		var username, password, email string
 
 		if len(creds) > 0 {
-			// We do not allow to mix user-defined and auto signup modes.
-			// If the user specifies a custom login file, then it should contain
-			// all the users that is expected to login during a load-test.
-			if id >= len(creds) {
-				return nil, errors.New("exceeded user list limit")
-			}
 			// Emails and passwords are separated by space.
 			split := strings.Split(creds[id], " ")
 			if len(split) < 2 {
@@ -433,4 +413,30 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 			panic("controller type must be valid")
 		}
 	}, nil
+}
+
+func getUserCredentials(usersFilePath string, config *loadtest.Config) ([]string, error) {
+	var lines []string
+	if usersFilePath == "" {
+		return lines, nil
+	}
+	f, err := os.Open(usersFilePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open %q: %w", usersFilePath, err)
+	}
+	defer f.Close()
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("failed to read from %s: %w", f.Name(), err)
+	}
+	// We do not allow to mix user-defined and auto signup modes.
+	// If the user specifies a custom login file, then it should contain
+	// all the users that is expected to login during a load-test.
+	if len(lines) < config.UsersConfiguration.MaxActiveUsers+1 {
+		return nil, fmt.Errorf("number of lines in %q is %d, which is less than MaxActiveUsers+1(%d)", usersFilePath, len(lines), config.UsersConfiguration.MaxActiveUsers+1)
+	}
+	return lines, nil
 }
--- a/api/agent.go
+++ b/api/agent.go
@@ -4,6 +4,7 @@
 package api
 
 import (
+	"bufio"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -11,7 +12,9 @@ import (
 	"io/ioutil"
 	"net"
 	"net/http"
+	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	client "github.com/mattermost/mattermost-load-test-ng/api/client/agent"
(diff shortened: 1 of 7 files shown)


docs/loadtest_config.md
```diff
@@ -12,7 +12,7 @@ The URL to direct the load. Should be the public facing URL of the target Matter
 
 *string*
 
-The URL to the WebSocket endpoint the users will connect to.  
+The URL to the WebSocket endpoint the users will connect to.
```
streamer45, line 15: The double spaces were actually meant to break the line. I guess it might have been an automatic editor thing on save :P

docs/loadtest_config.md
```diff
@@ -52,9 +52,9 @@ The distribution of action rates for running controllers.
 
 Rate is a multiplier that will affect the speed at which user actions are executed by the `UserController`.
 
-A rate < 1.0 will run actions at a faster pace.   
-A rate == 1.0 will run actions at the default pace.    
-A rate > 1.0 will run actions at a slower pace.  
+A rate < 1.0 will run actions at a faster pace.
+A rate == 1.0 will run actions at the default pace.
+A rate > 1.0 will run actions at a slower pace.
```
streamer45, line 57: Same here.

docs/loadtest_config.md
```diff
@@ -71,7 +71,7 @@ This value overrides the actual server version. If left empty, the one returned
 
 *int*
 
-The number of teams the target Mattermost instance should have.  
+The number of teams the target Mattermost instance should have.
```
streamer45, line 74: Same here.

api/agent.go
```diff
@@ -322,15 +333,60 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 		}
 	}
 
+	var creds []string
+	f, err := os.Open(config.UsersConfiguration.UsersFilePath)
```
streamer45, line 337: Why don't we just return an error if `UsersFilePath` is not empty and `os.Open()` fails?

api/agent.go
```diff
@@ -322,15 +333,60 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
 
+		var username, password, email string
+
+		if len(creds) > 0 {
+			// We do not allow to mix user-defined and auto signup modes.
+			// If the user specifies a custom login file, then it should contain
+			// all the users that is expected to login during a load-test.
+			if id >= len(creds) {
+				return nil, errors.New("exceeded user list limit")
+			}
```
streamer45, line 366: I think we should prevent failing in the middle of a load-test when possible. We have access to the config so we know how many users we can potentially add (`MaxActiveUsers`) and if that value doesn't match our expectations we should probably fail before starting anything.

api/agent.go
```diff
@@ -322,15 +333,60 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
+			password = split[1]
+			// Quick and dirty hack to extract username from email.
+			// This is not terribly important to be correct.
+			username = strings.Split(email, "@")[0]
+			username = strings.Replace(username, "+", "-", -1)
+		} else {
+			username = fmt.Sprintf("%s-%d", namePrefix, id)
+			email = fmt.Sprintf("%s-%d@example.com", namePrefix, id)
+			password = "testPass123$"
+		}
```
streamer45, line 382: I think all this logic would be better inside a dedicated function that takes the `UsersFilePath` and returns a list of either `model.User` or a smaller ad-hoc structure with the three values we need (username, email and password).

streamer45 (commented):
Thanks for adding this :+1:  I've left a couple of suggestions.

docs/loadtest_config.md
agnivade: Ah I see :)

api/agent.go
agnivade: We can if you want to. I thought of falling back to the default signup if the file open fails, with a warning message.

streamer45: This is a common problem. Personally, given the load-test is by nature a time consuming process, I'd prefer being precise and apply a "fail fast" approach. If a path was given in the config I'd expect it's needed. The user could miss the warning and think everything went okay since the load-test will run anyway, only to find out the wrong users logged in. It would be the user's fault of course but we can prevent it by being extra explicit.

docs/loadtest_config.md
```diff
@@ -82,6 +82,12 @@ These will be created during the `init` process.
 
 The amount of active users to run when the load-test starts.
 
+### UsersFilePath
+
+*string*
+
+The path to the file which contains a list of user email and passwords that will be used by the tool if set. Each line should be for a user containing an email and password separated by space. It should have enough lines to contain the maximum number of users a load-test is expected to reach.
+
```
isacikgoz, line 90: Maybe we should also state that it shouldn't exceed the `MaxActiveUsers` limit.

isacikgoz (commented):
Looks good, just a small thing besides @streamer45's comments.

docs/loadtest_config.md
agnivade: You mean it should be _at least_ `MaxActiveUsers` ;) It can exceed, that's not a problem.

api/agent.go
agnivade: I tried to put the entire logic in a separate function, but for the normal case, it would unnecessarily allocate a big slice. And there is the case of the custom offset also. Which might increase the slice further. So kept the user creation part as-is. But open to refactoring this too. Let me know.

docs/loadtest_config.md
isacikgoz: I see, yeah. Thanks for clarification 👍

docs/loadtest_config.md
```diff
@@ -52,9 +52,9 @@ The distribution of action rates for running controllers.
 
 Rate is a multiplier that will affect the speed at which user actions are executed by the `UserController`.
 
-A rate < 1.0 will run actions at a faster pace.
-A rate == 1.0 will run actions at the default pace.
-A rate > 1.0 will run actions at a slower pace.
+A rate < 1.0 will run actions at a faster pace.   
+A rate == 1.0 will run actions at the default pace.    
+A rate > 1.0 will run actions at a slower pace.
```
isacikgoz: These trailing whitespaces strikes back! 😅

isacikgoz (approved):
LGTM, nothing blocking. 👍

api/agent.go
```diff
@@ -434,3 +414,29 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 		}
 	}, nil
 }
+
+func getUserCredentials(usersFilePath string, config *loadtest.Config) ([]string, error) {
+	var lines []string
+	if usersFilePath == "" {
+		return lines, nil
+	}
```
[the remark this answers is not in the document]
agnivade: No reason. Doesn't make any difference. Nil slice is same as `nil`. Length will be 0 in both cases.

api/agent.go
streamer45: Are we talking about the `creds` slice? If the path is not defined/empty we shouldn't be allocating anything.

agnivade: But you wanted a slice of users for both cases right? And then the inner function would just pick up the user from that index.

api/agent.go
```diff
@@ -376,5 +412,31 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
+	for scanner.Scan() {
+		lines = append(lines, scanner.Text())
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("failed to read from %s: %w", f.Name(), err)
+	}
+	// We do not allow to mix user-defined and auto signup modes.
+	// If the user specifies a custom login file, then it should contain
+	// all the users that is expected to login during a load-test.
+	if len(lines) < config.UsersConfiguration.MaxActiveUsers+1 {
```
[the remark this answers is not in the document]
agnivade: The offset always starts from 1. So first user is never used.

api/agent.go
streamer45: I was actually only thinking about this case. Do the whole parsing before we run the load-test and return a slice of [username,email,password] tuples/structs.

agnivade: Ok, understood.

api/agent.go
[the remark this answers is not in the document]
streamer45: Interesting. I wonder if this is something we should fix. I can't remember a reason why we would be starting at 1.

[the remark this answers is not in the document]
agnivade: Yeah, we just need to get rid of the `+1` here `userId := activeUsers + 1`. Should not be an issue.

api/agent.go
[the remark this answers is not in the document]
isacikgoz: Fair enough, was curious if there was a convention that I was not aware of 😅

api/agent.go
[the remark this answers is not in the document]
streamer45: Alright, I'll write this down and take care of it in a future PR since it's unrelated.

streamer45 (approved):
Nicely done :+1:

PR #415, opened 2020-12-17
Adding cockroachDB support to the load tests, based on external installation

jespino:


agnivade:
@jespino - Can you clarify the objective of the PR a bit more in the PR commit message section? I am finding it a bit hard to understand with just the PR title. Thanks!

streamer45:
> @jespino - Can you clarify the objective of the PR a bit more in the PR commit message section? I am finding it a bit hard to understand with just the PR title. Thanks!

My understanding is that these are the changes he made to make the load-test work on a deployed instance connected to an external CockroachDB cluster.

jespino:
Yes, exactly, as Claudio explained, this is the way that I found to run load tests on CockroachDB, that also can work pretty well for testing external mysql or postgres instances (for example if we decide to go with postgresql in kubernetes, of if we want to try other databases like TiDB).

deployment/terraform/assets/cluster.tf
```diff
@@ -271,7 +271,7 @@ EOF
 
 
 resource "aws_rds_cluster_instance" "cluster_instances" {
-  count                      = var.app_instance_count > 0 ? var.db_instance_count : 0
```
jespino, line 274: I think I should check here if we have ExternalDBHost, and then, not run the RDS instances at all, instead of doing it by driver.

deployment/terraform/assets/cluster.tf
```diff
@@ -281,7 +281,7 @@ resource "aws_rds_cluster_instance" "cluster_instances" {
 }
 
 resource "aws_rds_cluster" "db_cluster" {
-  count               = var.app_instance_count > 0 ? 1 : 0
```
jespino, line 284: same here.

deployment/terraform/assets/variables.tf
```diff
@@ -62,6 +62,7 @@ variable "db_engine_version" {
   default = {
     "aurora-mysql"      = "5.7.mysql_aurora.2.09.0"
     "aurora-postgresql" = "11.7"
+    "cockroachdb"       = "20.1.2"
```
jespino, line 65: This version is not used anywhere, is to avoid failing to fail on terraform template filling, because is needed a version for each database (because RDS). I think we can solve that, or maybe put here something like `"cockroachdb" = "*"`.

agnivade:
Thanks guys :+1:

config/deployer.sample.json
```diff
@@ -21,6 +21,9 @@
   "DBInstanceType": "db.r4.large",
   "DBUserName": "mmuser",
   "DBPassword": "mostest80098bigpass_",
+  "ExternalDBHost": "",
```
jespino, line 24: Maybe we should us directly here something like ExternalDB only that get the DSN instead of forcing that, and if there is an external DSN defined use that instead of build that from the output config.

deployment/terraform/assets/cluster.tf
```diff
@@ -271,7 +271,7 @@ EOF
 
 
 resource "aws_rds_cluster_instance" "cluster_instances" {
-  count                      = var.app_instance_count > 0 ? var.db_instance_count : 0
+  count                      = var.app_instance_count > 0 && var.db_instance_engine != "cockroachdb" ? var.db_instance_count : 0
```
agnivade, line 274: Not sure if we need this, setting `db_instance_count` of 0 is all we need here.

deployment/terraform/assets/cluster.tf
```diff
@@ -281,7 +281,7 @@ resource "aws_rds_cluster_instance" "cluster_instances" {
 }
 
 resource "aws_rds_cluster" "db_cluster" {
-  count               = var.app_instance_count > 0 ? 1 : 0
+  count               = var.app_instance_count > 0 && var.db_instance_engine != "cockroachdb" ? 1 : 0
```
agnivade, line 284: Same here.

deployment/config.go
```diff
@@ -44,11 +44,17 @@ type Config struct {
 	// Type of the DB instance.
 	DBInstanceType string `default:"db.r4.large" validate:"notempty"`
 	// Type of the DB instance - postgres or mysql.
-	DBInstanceEngine string `default:"aurora-postgresql" validate:"oneof:{aurora-mysql, aurora-postgresql}"`
+	DBInstanceEngine string `default:"aurora-postgresql" validate:"oneof:{aurora-mysql, aurora-postgresql, cockroachdb}"`
```
agnivade, line 47: The database settings are getting a bit complicated. I would suggest splitting the entire thing out to a separate struct like `ElasticSearchSettings`. In that, I would mainly want 2 sections:
1. Just the terraform related settings, which are needed to deploy a DB.
2. Everything else that will be needed, if terraform doesn't deploy a DB.

With that in mind, I think we can remove `cockroachdb` from the `DBInstanceEngine` and keep the values only to what Terraform can use. This will also have the advantage of removing cockroach version from variables.tf which wasn't really needed and causes more confusion as to what's really happening.

Then I would suggest to remove separate settings for host, port and dbname, and replace them with exactly what we have in our config.json - DriverName, DataSource, DataSourceSearchReplicas. This makes the entire thing unaware of any driver specific logic and just blindly pastes whatever is given to the system.

deployment/terraform/create.go
```diff
@@ -317,6 +317,10 @@ func (t *Terraform) updateAppConfig(ip string, sshc *ssh.Client, output *Output)
 		clusterDSN = "postgres://" + t.config.DBUserName + ":" + t.config.DBPassword + "@" + output.DBCluster.ClusterEndpoint + "/" + t.config.ClusterName + "db?sslmode=disable"
 		readerDSN = []string{"postgres://" + t.config.DBUserName + ":" + t.config.DBPassword + "@" + output.DBCluster.ReaderEndpoint + "/" + t.config.ClusterName + "db?sslmode=disable"}
 		driverName = "postgres"
+	case "cockroachdb":
```
agnivade, line 320: We don't really need this case. We can just check if the terraform output has DBs, and if not straightaway go for copy-pasting whatever values are given in the datasource, drivername and replicas. This makes things more readable IMO.

You might need to add an extra method - HasDB.

agnivade (commented):
Great work putting this across @jespino! I think we can improve this a bit with some more effort. Please take a look.

deployment/terraform/assets/cluster.tf
jespino: Actually, this is about `app_instance_count`, so I guess that is a typo or we can't solve that this way because we are going to run a cluster anyway.

jespino:
@agnivade I reorganized everything using other approach, I have to test it, but I think now is more clean.

deployment/terraform/create.go
```diff
@@ -312,15 +312,22 @@ func (t *Terraform) createAdminUser(extAgent *ssh.ExtAgent, output *Output) erro
...
-	case "aurora-mysql":
-		clusterDSN = t.config.DBUserName + ":" + t.config.DBPassword + "@tcp(" + output.DBCluster.ClusterEndpoint + ")/" + t.config.ClusterName + "db?charset=utf8mb4,utf8\u0026readTimeout=30s\u0026writeTimeout=30s"
-		readerDSN = []string{t.config.DBUserName + ":" + t.config.DBPassword + "@tcp(" + output.DBCluster.ReaderEndpoint + ")/" + t.config.ClusterName + "db?charset=utf8mb4,utf8\u0026readTimeout=30s\u0026writeTimeout=30s"}
-		driverName = "mysql"
+
+	clusterDSN = t.config.ExternalDBSettings.DataSource
+	readerDSN = t.config.ExternalDBSettings.DataSourceReplicas
+	driverName = t.config.ExternalDBSettings.DriverName
+
+	if t.config.TerraformDBSettings.InstanceCount > 0 {
```
agnivade, line 320: Let's refactor this to a helper method to check the terraform output struct instead like we do below. Something like `if output.HasDB() {`

agnivade (commented):
Looks really amazing now. Just one small comment to improve on. Good to go after that.

deployment/terraform/create.go
jespino: I think now is done, but I'm not 100% sure if the `DBCluster.ClusterEndpoint` is going to be an empty string on the output.

agnivade: Can we verify that? You don't need to go through the entire rigmarole of setting up app servers and everything. Just comment out everything after `output, err := t.Output()` in (*Terraform).Create, create a deployment, and inspect the tfstate file.

jespino: Verified, it works well :)

deployment/terraform/output.go
```diff
@@ -129,7 +129,7 @@ func (o *Output) HasProxy() bool {
 	return o.Proxy.PrivateIP != ""
 }
 
-// HasProxy returns whether a deployment has proxy installed in it or not.
+// HasDB returns whether a deployment has database installed in it or not.
```
jespino, line 132: :man_facepalming:

agnivade (approved):

deployment/terraform/assets/cluster.tf
```diff
@@ -271,7 +271,7 @@ EOF
 
 
 resource "aws_rds_cluster_instance" "cluster_instances" {
-  count                      = var.app_instance_count > 0 ? var.db_instance_count : 0
+  count                      = var.db_instance_count
```
streamer45, line 274: There was a reason we were checking on `var.app_instance_count`. It's because when we set `AppInstanceCount` to zero (agents only deployment) I'd expect no DB instances to be created. We were defining a dependency relationship between the two values.

deployment/terraform/assets/cluster.tf
```diff
@@ -281,7 +281,7 @@ resource "aws_rds_cluster_instance" "cluster_instances" {
 }
 
 resource "aws_rds_cluster" "db_cluster" {
-  count               = var.app_instance_count > 0 ? 1 : 0
+  count               = var.db_instance_count > 0 ? 1 : 0
```
streamer45, line 284: Same here. This wasn't a typo. The idea was that we only create the DB cluster if we are also deploying the app servers.

streamer45 (commented):
Thanks @jespino, PR looks good :+1: I've left a couple of comments.

Also, I just want to point out this is somewhat of a breaking change but being limited to the deployer I am hoping it won't cause too many troubles.

deployment/terraform/assets/cluster.tf
jespino: I think now is working as expected.

streamer45: Perfect :+1:

streamer45 (approved):

PR #416, opened 2020-12-17
[MM-30334] Add file descriptors limit Sanity Check to the Agent

CandyZack:
#### Summary
Added a sanity check on `MaxActiveUsers` with `Rlimit`
Fixed a `defer` bug in testing - inside `TestRemoveUsers`
Changed the static `max connections in HTTP pool` to be dynamic

#### Ticket Link
JIRA - https://mattermost.atlassian.net/browse/MM-30334
Closes https://github.com/mattermost/mattermost-server/issues/16284

### DIFF
--- a/loadtest/loadtest.go
+++ b/loadtest/loadtest.go
@@ -18,6 +18,10 @@ import (
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
+const (
+	connFactor = 4
+)
+
 // LoadTester is a structure holding all the state needed to run a load-test.
 type LoadTester struct {
 	mut           sync.RWMutex
@@ -241,6 +245,11 @@ func (lt *LoadTester) Status() *Status {
 	}
 }
 
+// MaxHTTPConns returns the maximum number of HTTP connections to be used for the given number of users.
+func MaxHTTPConns(maxUsers int) int {
+	return maxUsers / connFactor
+}
+
 // New creates and initializes a new LoadTester with given config. A factory
 // function is also given to enable the creation of UserController values from within the
 // loadtest package.
@@ -259,7 +268,7 @@ func New(config *Config, nc NewController, log *mlog.Logger) (*LoadTester, error
 	}
 
 	if maxActiveUsers := config.UsersConfiguration.MaxActiveUsers; uint64(
-		maxActiveUsers+(maxActiveUsers/4)) > rlimit.Max {
+		maxActiveUsers+MaxHTTPConns(maxActiveUsers)) > rlimit.Max {
 		return nil, fmt.Errorf("MaxActiveUsers is not compatible with max Rlimit value. "+
 			"MaxActiveUsers = %d, max_Rlimit = %d",
 			maxActiveUsers, rlimit.Max)
--- a/loadtest/loadtest.go
+++ b/loadtest/loadtest.go
@@ -9,6 +9,7 @@ import (
 	"math/rand"
 	"sync"
 	"sync/atomic"
+	"syscall"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/defaults"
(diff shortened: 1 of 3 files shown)


CandyZack:
@agnivade @streamer45 Help Required on Testcases
Can you please tell me how to adjust the Unit Testcases for this changes.
How the RLIMIT_NOFILE is set for testcases ?

streamer45 (commented):
@CandyZack

Thank you for this :) I've left a couple of suggestions.

> @agnivade @streamer45 Help Required on Testcases
> Can you please tell me how to adjust the Unit Testcases for this changes.
> How the RLIMIT_NOFILE is set for testcases ?

Tests failing are actually a good sign, it means it's working. Probably the CI environment has very strict limits and the hard-coded `500` value is hitting those. Maybe we can fix this by making that value a function of `MaxActiveUsers`. That means that instead of 500 we could use `MaxActiveUsers/4`.

CandyZack:
@streamer45  I have made the changes. Can I get info on failing testcases ?

loadtest/loadtest.go
```diff
@@ -252,6 +253,12 @@ func New(config *Config, nc NewController, log *mlog.Logger) (*LoadTester, error
 		return nil, fmt.Errorf("could not validate configuration: %w", err)
 	}
 
+	if 5*config.UsersConfiguration.MaxActiveUsers/4 > syscall.RLIMIT_NOFILE {
```
streamer45, line 256: Why the 5? I am thinking the value should be `MaxActiveUsers + (MaxActiveUsers/4)`.
Also, since we are here, let's update the original values where the `500` came from:

https://github.com/mattermost/mattermost-load-test-ng/blob/fe31d56e2024672fc5c5d4400fa0045a4937a366/api/agent.go#L318-L320 

Let's make those 500 values to be `MaxActiveUsers/4` instead.

> @streamer45 I have made the changes. Can I get info on failing testcases ?

The reason the tests are failing is because `syscall.RLIMIT_NOFILE` is just a constant (its value is 7) that is used to denote what type of call to make to the underlying operating system. It's not the number we are looking for. 

To get that we need to use `syscall.GetRlimit()`, e.g.:

```go
var rlimit syscall.Rlimit
if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
  return nil, err
}
```

After that call `rlimit.Max` will hold the value we are interested in.

CandyZack: Thanks! Now I understood the stuff

loadtest/loadtest.go
```diff
@@ -252,6 +253,18 @@ func New(config *Config, nc NewController, log *mlog.Logger) (*LoadTester, error
 		return nil, fmt.Errorf("could not validate configuration: %w", err)
 	}
 
+	var rlimit syscall.Rlimit
+	if err := syscall.Getrlimit(syscall.RLIMIT_NOFILE, &rlimit); err != nil {
+		return nil, fmt.Errorf("could not get the Rlimit value: %w", err)
+	}
+
+	if maxActiveUsers := config.UsersConfiguration.MaxActiveUsers; uint64(
+		maxActiveUsers+(maxActiveUsers/4)) > rlimit.Max {
```
agnivade, line 262: Let's refactor 4 to be a constant. And use that here and in the api/agent.go file. Perhaps `loadFactor` is an apt name, although it's a bit hazy. @streamer45 - any thoughts on this?

streamer45: Maybe `connFactor` since the result is about connections. We might also want to either export it so we can use it in `NewControllerWrapper` or create a helper function (e.g. `MaxHTTPConns(maxUsers int) int`)  that given `MaxActiveUsers` returns the max connections value. This way we could also take care of some edge cases (e.g. input value too low).

CandyZack: so is a helper function required for now? or just a constant will do?

agnivade: A helper function sounds like a good idea. :+1:

loadtest/loadtest.go
```diff
@@ -240,6 +245,11 @@ func (lt *LoadTester) Status() *Status {
 	}
 }
 
+// MaxHTTPConns returns the maximum connections in HTTP pool
```
streamer45, line 248:
```suggestion
// MaxHTTPConns returns the maximum number of HTTP connections to be used for the given number of users.
```

agnivade (approved):
Nice :+1:

streamer45 (approved):
Thank you @CandyZack Looks very polished, great work on this :+1: 
Left only a minor suggestion.

PR #417, opened 2020-12-18
[MM-30397] Enable loading DB dumps during comparison

streamer45:
#### Summary

PR adds support for loading raw SQL dumps into the instances before performing a load-test comparison.

#### Ticket

https://mattermost.atlassian.net/browse/MM-30397


agnivade:
Any reason why this is only during comparison? Is the expectation that for a normal load-test, the user can upload the dump themselves?

streamer45:
> Any reason why this is only during comparison? Is the expectation that for a normal load-test, the user can upload the dump themselves?

None other than my primary focus was on this. I think it makes sense to port this functionality to common terraform deployments as well but maybe as a separate piece of work. We can open a ticket for that.

agnivade (approved):

isacikgoz (approved):
Nice one! 🚀

streamer45:
> > Any reason why this is only during comparison? Is the expectation that for a normal load-test, the user can upload the dump themselves?
> 
> None other than my primary focus was on this. I think it makes sense to port this functionality to common terraform deployments as well but maybe as a separate piece of work. We can open a ticket for that.

Created [MM-31541](https://mattermost.atlassian.net/browse/MM-31541).

PR #418, opened 2020-12-21
[MM-31481] Fix race condition during terraform init

streamer45:
#### Summary

This fixes an annoying error happening sporadically when deploying resources during comparisons which would led to an unrecoverable error with significant waste of time. 

After some investigation, it became clear it was a race condition caused by the concurrent nature of the comparison process. 
In fact, calling `terraform init` in parallel was not a great idea since it can touch some common files during plugin initialization (`.terraform/plugins/selections.json`). This would cause the process to read bad JSON at times and failing.

To fix this I've added a global mutex to the `terraform` package since the init call was buried inside unexported methods and it was indirectly called in several places (create/destroy and coordinator specific methods). Doing the lock at the higher level (e.g. in the `comparison` package) would have meant either substantial changes to the internal functioning or locking on the whole create/destroy processes which is a no go. 

So I went for the quick (possibly dirty) fix. Let me know if you find that acceptable or if you see better options.

#### Ticket

https://mattermost.atlassian.net/browse/MM-31481


agnivade:
Is it very hard to make it an exported method and move it out of preFlightCheck and call it as a separate method from everywhere?

streamer45:
> Is it very hard to make it an exported method and move it out of preFlightCheck and call it as a separate method from everywhere?

I don't think it's hard from an implementation perspective but it probably makes the resulting code harder to use.
If we decouple `t.init()` and export it we would need to have the caller do it before any of the methods that requires it. Which is fine but it also creates smaller issues like checking terraform version which is currently done before calling the init and I'd prefer not to lock on that as well. We could invert the calls, so we do init first and then we check the version but I am not sure that makes sense.

agnivade:
Hmm I see. Okay let's add a comment to the mutex explaining the reason for it to be a global.

isacikgoz (approved):
🐎

agnivade (approved):

PR #419, opened 2020-12-22
[MM-31544] Fix server sporadically failing to restart

streamer45:
#### Summary

PR fixes an issue happening sporadically when attempting to re-deploy. The server would fail to restart and eventually cause the server status check to timeout.

This is because on subsequent deploy runs we are uploading the config file without making sure the server is stopped. This was causing a sort of race condition where at the same time of the upload the server was writing to the config file making the result garbled. 

#### Ticket

https://mattermost.atlassian.net/browse/MM-31544


agnivade (approved):
Great, that was it!

isacikgoz (approved):
Nice catch!

PR #421, opened 2021-01-07
MM-31746: Handle new error strings due user creation

agnivade:
https://mattermost.atlassian.net/browse/MM-31746

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -309,7 +309,8 @@ func (t *Terraform) createAdminUser(extAgent *ssh.ExtAgent, output *Output) erro
 		return err
 	}
 	if out, err := sshc.RunCommand(cmd); err != nil {
-		if strings.Contains(string(out), "account with that username already exists") {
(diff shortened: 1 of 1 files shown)


streamer45:
I might have missed it during the error discussion but is this correct behaviour? To return the error id instead of the translation string.

If so, we have a couple more places where we had to use a similar hack:

https://github.com/mattermost/mattermost-load-test-ng/blob/fc6b711dfb3e74ef8fdd2f6d997d01b2bd129913/loadtest/control/actions.go#L44
https://github.com/mattermost/mattermost-load-test-ng/blob/fc6b711dfb3e74ef8fdd2f6d997d01b2bd129913/deployment/terraform/create.go#L312

We might want to update those as well if they are affected.

agnivade:
It's definitely not right, but the translation string just doesn't appear. I became a bit lazy and didn't dig deep enough to figure out why. Let me do that first.

agnivade:
Somehow I am not able to reproduce it now. I'll look into this when I encounter it the next time.

PR #423, opened 2021-01-15
[MM-31921] Improve error and documentation when file descriptor limit is reached

streamer45:
#### Summary

PR improves the error in case the limit of file descriptors is reached by proposing a working value.

#### Ticket 

https://mattermost.atlassian.net/browse/MM-31921


isacikgoz (approved):
Awesome 🚀

PR #425, opened 2021-01-19
ltctl ssh: extra targets

lieut-data:
#### Summary
Support the following extra `ltctl ssh` targets: `loadtest-proxy` (or just `proxy`), `loadtest-metrics` (or just `metrics`, `grafana`, or `prometheus`).

#### Ticket Link
None.

### DIFF
--- a/deployment/terraform/utils.go
+++ b/deployment/terraform/utils.go
@@ -68,6 +68,14 @@ func (t *Terraform) OpenSSHFor(resource string) error {
 		}
 	}
 	if cmd == nil {
+		switch resource {
+		case "proxy", output.Proxy.Tags.Name:
+			cmd = exec.Command("ssh", fmt.Sprintf("ubuntu@%s", output.Proxy.PublicIP))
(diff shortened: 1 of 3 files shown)


deployment/terraform/utils.go
```diff
@@ -67,6 +67,14 @@ func (t *Terraform) OpenSSHFor(resource string) error {
 			}
 		}
 	}
+	if cmd == nil {
+		switch resource {
```
agnivade, line 71: Can we move this to the upper block?

agnivade (commented):
We should also document the fact that proxy, metrics, prometheus, grafana are secret magic words that also lets you ssh to those boxes. :)

isacikgoz (approved):
Looks good, agree with @agnivade's suggestion.

deployment/terraform/utils.go
lieut-data: Good suggestion! Changes pushed, along with new documentation for using `ltctl ssh`.

agnivade (approved):
Looks amazing!

PR #426, opened 2021-01-26
[MM-31130] Review SimulController action and rates

streamer45:
#### Summary

- Reviewed action rates and frequencies to better match peak load on community.
- Added actions to create public/private channels to `SimulController`.
- Added scroll channel action to `SimulController`.
- Converted action frequencies from int to float to facilitate adding new actions without the need to re-calculate everything.
- Improved fetching users info from posts, added basic mention parsing.

Adding the `Do Not Merge` label since I am still running a few load-tests to quantify the impact of the change. Also I might tweak a few numbers here and there if it makes sense to do so. Other than that the code is ready for review.

#### Ticket

https://mattermost.atlassian.net/browse/MM-31130


loadtest/control/simulcontroller/actions.go
```diff
@@ -1017,3 +1056,92 @@ func searchGroupChannels(u user.User) control.UserActionResponse {
...
+	}
+
+	ids, err := u.GetUsersNotInChannel(team.Id, channelId, 0, 100)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	// we pick up to 4 users to add to the channel.
+	for _, id := range pickIds(ids, 1+rand.Intn(4)) {
+		if err := u.AddChannelMember(channelId, id); err != nil {
```
agnivade, line 1092: We are not doing this for creating public channels. Is that because users will join the channel by themselves?

agnivade (commented):
> Converted action frequencies from int to float to facilitate adding new actions without the need to re-calculate everything.

I might have forgotten some details about the algo, but how is changing from int to float helping?

loadtest/control/simulcontroller/actions.go
streamer45: Yes. Ideally we should have an "invite action" to invite users to channel we created and also a "leave channel" one so that they don't grow indefinitely.

> I might have forgotten some details about the algo, but how is changing from int to float helping?

The algorithm is more or less unchanged. The advantage of using float is that if I add an action I can always find a small enough number to fit that in. Before, if the action rate was less than 1 I had to multiply everything else by some constant to make sure the frequencies remained proportional to each other.

agnivade (approved):

isacikgoz (approved):
LGTM 👍

streamer45:
This is finally ready to be merged. I've added a couple of fixes and additions since last time so I am asking for a re-review of the new commits (starting from ff769d6). 

It's all pretty self-explanatory with the exception of a small tweak to the feedback loop. The `stopThreshold` variable was tuned (doubled) to account for the slower rate the controller works at now. This was done to to keep the time it takes to converge close to 1h while still maintaining stable results. I've tested it for a while and everything looked good.

coordinator/coordinator.go
```diff
@@ -132,6 +132,10 @@ func (c *Coordinator) Run() (<-chan struct{}, error) {
 					y: status.ActiveUsers,
 				})
 				latest := getLatestSamples(samples, samplesTimeRange)
+				if len(latest) > 0 {
+					c.log.Debug(fmt.Sprintf("feeback loop info: %d/%d samples, %0.3f slope",
```
agnivade, line 136: Should we use structured logging here?

agnivade (approved):

coordinator/coordinator.go
streamer45: I had it structured before and moved to this because it was getting a bit verbose and harder to read.

isacikgoz (approved):
LGTM 👍

PR #427, opened 2021-02-03
[MM-32270] Limit terraform re-initialization during comparisons

streamer45:
#### Summary

PR addresses the unhelpful spam of logs returned by the comparison process.
Mostly was due to terraform initialization when performing coordinator's commands (start/stop/getstatus).
To fix this I've decoupled the logic a bit and cached the terraform outputs from the deployments.

#### Ticket

https://mattermost.atlassian.net/browse/MM-32270

### DIFF
--- a/deployment/terraform/agent.go
+++ b/deployment/terraform/agent.go
@@ -9,27 +9,26 @@ import (
 	"strings"
 	"text/template"
 
-	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func generateLoadtestAgentConfig(dpConfig *deployment.Config, output *Output) (*loadtest.Config, error) {
+func (t *Terraform) generateLoadtestAgentConfig() (*loadtest.Config, error) {
 	cfg, err := loadtest.ReadConfig("")
 	if err != nil {
 		return nil, err
 	}
-	url := output.Instances[0].PrivateIP + ":8065"
-	if output.HasProxy() {
-		url = output.Proxy.PrivateIP
+	url := t.output.Instances[0].PrivateIP + ":8065"
+	if t.output.HasProxy() {
+		url = t.output.Proxy.PrivateIP
 	}
 
 	cfg.ConnectionConfiguration.ServerURL = "http://" + url
 	cfg.ConnectionConfiguration.WebSocketURL = "ws://" + url
-	cfg.ConnectionConfiguration.AdminEmail = dpConfig.AdminEmail
-	cfg.ConnectionConfiguration.AdminPassword = dpConfig.AdminPassword
+	cfg.ConnectionConfiguration.AdminEmail = t.config.AdminEmail
+	cfg.ConnectionConfiguration.AdminPassword = t.config.AdminPassword
 
 	return cfg, nil
 }
@@ -118,7 +117,7 @@ func (t *Terraform) initLoadtest(extAgent *ssh.ExtAgent, initData bool) error {
 		return err
 	}
 	mlog.Info("Generating load-test config")
-	cfg, err := generateLoadtestAgentConfig(t.config, t.output)
+	cfg, err := t.generateLoadtestAgentConfig()
 	if err != nil {
 		return err
 	}
--- a/deployment/terraform/agent.go
+++ b/deployment/terraform/agent.go
@@ -9,13 +9,14 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func (t *Terraform) generateLoadtestAgentConfig(output *Output) (*loadtest.Config, error) {
+func generateLoadtestAgentConfig(dpConfig *deployment.Config, output *Output) (*loadtest.Config, error) {
 	cfg, err := loadtest.ReadConfig("")
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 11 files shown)


deployment/terraform/coordinator.go
```diff
@@ -185,23 +169,15 @@ func (t *Terraform) StopCoordinator() (coordinator.Status, error) {
...
 // GetCoordinatorStatus returns information about the status of the
 // coordinator in the current load-test deployment.
-func (t *Terraform) GetCoordinatorStatus() (coordinator.Status, error) {
+func GetCoordinatorStatus(dpConfig *deployment.Config, output *Output) (coordinator.Status, error) {
 	var status coordinator.Status
-	if err := t.preFlightCheck(); err != nil {
-		return status, err
-	}
-
-	output, err := t.Output()
```
agnivade, line 194: The refactoring out of `PreFlightCheck` makes total sense to me.

I am wondering whether it's possible to maintain the output caching logic inside the terraform struct itself? It seems a bit odd to move this responsibility to consumer packages. I am considering alternatives like storing the output struct within the Terraform struct itself, and then either have a separate method called `CachedOutput` or pass a flag or something to the `Output` method to return the cached or a new response.

That way we still keep the logic of caching/not caching within the terraform package itself.

agnivade (changes requested):
Excited for this!

I just have some comments on the package design choices. Let me know what you think.

deployment/terraform/coordinator.go
streamer45: I think that's a great idea. It would mean reverting the coordinator functions to be methods of `terraform.Terraform` but at least we'd avoid the slightly awkward parameters. I like it :+1:

deployment/terraform/create.go
```diff
@@ -406,7 +406,7 @@ func (t *Terraform) updateAppConfig(ip string, sshc *ssh.Client, output *Output)
 	return nil
 }
 
-func (t *Terraform) preFlightCheck() error {
+func (t *Terraform) PreFlightCheck() error {
```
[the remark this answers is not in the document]
streamer45: Yes Sir :grin:

streamer45:
In the end I took a slightly different approach. Extracting `PreFlightCheck` was more troubles than gains. I am not sure it makes sense to trust the caller to run that on some subset of methods. Instead, I opted to keep a state variable internally to the `Terraform` object to avoid re-doing initialization and validation, which are what is eventually causing the logs. This way we still get the protection (ssh and version check) without moving the burden on the caller.

As far as the output I've mostly followed your (great) suggestion. I think we should be okay with either returning the value from disk if missing or otherwise trust whatever was last set internally. The output should only change on create/destroy as far as I understand.

Please let me know if that makes sense.

agnivade:
> The output should only change on create/destroy as far as I understand.

Hmm .. actually, I am thinking of adding a separate command for `refresh` to help ease this task. So the output can indeed change within a a create/destroy cycle. But looking at the code, it looks like it is cached only during the runtime of a single process. On the next run, it picks up from the .tfstate file again. So I think it should be good.

I will take a deeper look tomorrow :+1:

deployment/terraform/agent.go
```diff
@@ -9,13 +9,14 @@ import (
...
 
+	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
 
 	"github.com/mattermost/mattermost-server/v5/mlog"
 )
 
-func (t *Terraform) generateLoadtestAgentConfig(output *Output) (*loadtest.Config, error) {
+func generateLoadtestAgentConfig(dpConfig *deployment.Config, output *Output) (*loadtest.Config, error) {
```
agnivade, line 19: Why can't we use t.Config here?

deployment/terraform/output.go
```diff
@@ -121,7 +120,19 @@ func (t *Terraform) Output() (*Output, error) {
...
 		outputv2.S3Key = o.S3Key.Value[0]
 	}
-	return outputv2, nil
+
+	t.output = outputv2
+
+	return nil
+}
+
+// Output reads the current terraform output
```
agnivade, line 129: Maybe better to mention that after the first call, it will cache the output.

deployment/terraform/coordinator.go
```diff
@@ -187,19 +182,15 @@ func (t *Terraform) StopCoordinator() (coordinator.Status, error) {
 // coordinator in the current load-test deployment.
 func (t *Terraform) GetCoordinatorStatus() (coordinator.Status, error) {
 	var status coordinator.Status
-	if err := t.preFlightCheck(); err != nil {
-		return status, err
-	}
 
-	output, err := t.Output()
-	if err != nil {
+	if _, err := t.Output(); err != nil {
```
agnivade, line 186: Since we are in the package itself, I am wondering if we can use an internal method which does not return the config? Because it looks like all we are doing is populating the `t.output` variable. 0/5. It seems a bit odd to return the output, but not use it.

agnivade (commented):
I like this one! Just some small clarifications.

deployment/terraform/agent.go
streamer45: We can, I think I just forgot to revert from the previous implementation. Will amend :+1:

deployment/terraform/coordinator.go
streamer45: Sure we can add an internal method that only sets it if not set already.

deployment/terraform/output.go
```diff
@@ -126,13 +126,17 @@ func (t *Terraform) loadOutput() error {
...
+		return t.loadOutput()
 	}
-	return t.output, err
+	return nil
+}
+
+// Output reads the current terraform output and caches it internally for future use.
+// The output is guaranteed to be up to date after calls to Create and Destroy.
+func (t *Terraform) Output() (*Output, error) {
+	return t.output, t.setOutput()
```
streamer45, line 139: This seems to be working although I am not sure if it's defined behaviour. Please let me know if it's better/safer to store the result in a variable first.

agnivade: Functionally it's the same. But it does look a bit odd to read at first.

Something like this has more repetition, but looks more readable. 0/5, up to you

```
var err error
if t.output == nil {
  err = t.loadOutput()
}
return t.output, err
```

agnivade (approved):
Looks great!

isacikgoz (approved):
Awesome 🚀

PR #429, opened 2021-02-04
Fix collect ltagent.log files

jespino:
#### Ticket
[MM-32592](https://mattermost.atlassian.net/browse/MM-32592)


streamer45 (approved):
Thanks :+1:

PR #430, opened 2021-02-08
[MM-32670] Replace reset CLI command during automated comparisons

streamer45:
#### Summary

PR replaces the `mattermost reset` CLI command during automated comparisons.
The main reason behind the change is that the `reset` command can cause unnecessary delays during store initialization.

#### Ticket

https://mattermost.atlassian.net/browse/MM-32670

### DIFF
--- a/comparison/loadtest.go
+++ b/comparison/loadtest.go
@@ -157,11 +157,19 @@ func initLoadTest(t *terraform.Terraform, buildCfg BuildConfig, dumpFilename str
 		clients: appClients,
 	}
 
-	binaryPath := "/opt/mattermost/bin/mattermost"
 	resetCmd := cmd{
 		msg:     "Resetting database",
-		value:   fmt.Sprintf("%s reset --confirm", binaryPath),
 		clients: []*ssh.Client{appClients[0]},
+	}
+	switch dpConfig.TerraformDBSettings.InstanceEngine {
+	case "aurora-postgresql":
+		subCmd := fmt.Sprintf("-U %s -h %s %sdb", dpConfig.TerraformDBSettings.UserName, tfOutput.DBCluster.ClusterEndpoint, dpConfig.ClusterName)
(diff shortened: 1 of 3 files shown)


comparison/loadtest.go
```diff
@@ -157,12 +157,20 @@ func initLoadTest(t *terraform.Terraform, buildCfg BuildConfig, dumpFilename str
...
 
-	binaryPath := "/opt/mattermost/bin/mattermost"
 	resetCmd := cmd{
 		msg:     "Resetting database",
-		value:   fmt.Sprintf("%s reset --confirm", binaryPath),
 		clients: []*ssh.Client{appClients[0]},
 	}
+	switch dpConfig.TerraformDBSettings.InstanceEngine {
+	case "aurora-postgresql":
+		subCmd := fmt.Sprintf("-U %s -h %s %sdb", dpConfig.TerraformDBSettings.UserName, tfOutput.DBCluster.ClusterEndpoint, dpConfig.ClusterName)
```
agnivade, line 166: Should we refactor this logic of "ClusterName"+db out?

agnivade (approved):
The user doesn't need to be granted permissions to the new DB again right?

streamer45:
> The user doesn't need to be granted permissions to the new DB again right?

The db is created on behalf of the user so I'd say no. I've quickly tested it and didn't spot any issue.

comparison/loadtest.go
streamer45: Seems reasonable :+1:

agnivade:
Oh, the reason I asked that to be refactored was because it's also used in `(t *Terraform) updateAppConfig`. :P So we should use the method there too.

streamer45:
> Oh, the reason I asked that to be refactored was because it's also used in `(t *Terraform) updateAppConfig`. :P So we should use the method there too.

Alright, you meant a slightly higher level refactor than what I thought, no worries :)

PR #431, opened 2021-02-10
MM-32747: Display command output from ssh commands during ltctl reset

agnivade:
We were doing this elsewhere but missed out on this one. Without this,
the only error message we get is "exit code 1", and the actual command error
is missed out.

https://mattermost.atlassian.net/browse/MM-32747


streamer45:
Thanks :+1:

streamer45 (approved):

PR #432, opened 2021-02-16
[MM-26764] Improve proxy performance

streamer45:
#### Summary

PR introduces some changes to proxy related config settings that showed good performance improvement in terms of supported user counts.

#### Ticket

https://mattermost.atlassian.net/browse/MM-26764

### DIFF
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -70,7 +70,8 @@ pid /run/nginx.pid;
 include /etc/nginx/modules-enabled/*.conf;
 
 events {
-	worker_connections 200000;
+	worker_connections 16384;
+	use epoll;
 }
 
 
@@ -83,7 +84,8 @@ http {
 	sendfile on;
 	tcp_nopush on;
 	tcp_nodelay on;
-	keepalive_timeout 65;
+	keepalive_timeout 75s;
+	keepalive_requests 16384;
 	types_hash_max_size 2048;
 	include /etc/nginx/mime.types;
 	default_type application/octet-stream;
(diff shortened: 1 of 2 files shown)


deployment/terraform/strings.go
```diff
@@ -70,7 +70,8 @@ pid /run/nginx.pid;
 include /etc/nginx/modules-enabled/*.conf;
 
 events {
-	worker_connections 200000;
+	worker_connections 16384;
+	use epoll;
```
streamer45, line 74: This is just to be explicit. It should be on by default on linux.

deployment/terraform/strings.go
```diff
@@ -83,7 +84,8 @@ http {
 	sendfile on;
 	tcp_nopush on;
 	tcp_nodelay on;
-	keepalive_timeout 65;
+	keepalive_timeout 75s;
```
streamer45, line 87: Same, this is a default.

streamer45:
# LT Proxy Performance Analysis

## Results

image
image
image
image

## Proposed Changes

Suggested changes are in descending order of improvement. (first those that showed the highest gain)

### AWS

#### EC2 type

`m4.xlarge` -> `c5.xlarge`

`c5.xlarge` was shown to significantly outperform `m4.xlarge` on CPU heavy tasks. The proxy hardly consumes more than 500MB of RAM on average so the added memory on `m4` is not needed. The boost in CPU from `c5` instead really helps with scaling up users.

### NGNIX

#### `upstream keepalive`

`keepalive 64;` -> `keepalive 256;`

Increasing the number of upstream keepalive connections per worker. Raising this value helps with the overall throughput.

#### `keepalive_timeout`

`keepalive_timeout 65s` -> `keepalive_timeout 75s;`

Going back to defaults on this.

#### `keepalive_requests`

`keepalive_requests 100` -> `keepalive_requests 16394;`

Increasing the maximum number of requests that can be served through one keep-alive connection.
This can be especially helpful in load-testing scenarios.

#### `worker_connections`

`worker_connections 200000` -> `worker_connections 16384;`

This is a per CPU worker setting so briging it down to a sane value.

### Other evaluated settings

#### `tcp_nopush`

`tcp_nopush on;` -> `tcp_nopush off;`

The idea behing this change was that disabling `TCP_CORK` should help reducing latency especially for burst of small requests. Results did not show any significant change.

#### `listen reuseport`

`listen 80;` -> `listen 80 reuseport;`

Enables socket sharding, allowing multiple sockets to listen on the same IP address and port combination. 
The kernel then load balances incoming connections across the sockets.
This has shown good results on peak performance but not that great on longer tests. 

#### `events multi_accept`

`multiaccept off;` -> `multiaccept on;`

Lets worker processes accept new connections concurrently. Similarly to the above setting, this has shown a good increase in peak users but did eventually lower the average on longer tests.

isacikgoz (approved):
LGTM 👍

agnivade:
Terrific analysis! What was the setups that you compared with in the proxy vs no proxy case?

deployment/terraform/strings.go
```diff
@@ -83,7 +84,8 @@ http {
...
+	keepalive_requests 16384;
```
agnivade, line 88: Oh wow, this was a good one. :+1:

deployment/terraform/strings.go
```diff
@@ -99,55 +101,55 @@ http {
 const nginxSiteConfig = `
 upstream backend {
 %s
-        keepalive 64;
+    keepalive 256;
```
agnivade, line 104: This is interesting. This is actually not the number of upstream connections per worker, but rather the maximum number of _idle_ keepalive connections that are kept in the worker cache.

From nginx docs:

> It should be particularly noted that the keepalive directive does not limit the total number of connections to upstream servers that an nginx worker process can open. The connections parameter should be set to a number small enough to let upstream servers process new incoming connections as well.

I deliberately set it to 64 after considering these aspects. So it's very interesting to see that bumping this actually has an effect.

agnivade (approved):
Very methodical work as usual. Though I am a bit surprised with the keepalive setting :P

deployment/terraform/strings.go
streamer45: I think that particularly helped with bursts of requests. `64` was a good value and I've tested with several settings from disabled up to `16384` and the sweet spot seemed to lay somewhere between 200 and 300. We are not talking about huge improvements so it's okay to keep monitoring and possibly iterating. Maybe this will change with the reviewed controller. I can give this a second pass when I am merging that.

agnivade: Yep, so only for a burst of requests, this will be helpful. I guess I was surprised that we _do_ have a "burst" in our load test. Because I was thinking like a smooth ramp up, and then keep the same load.

Totally fine with merging it as is.

streamer45: I am pretty sure it's due to those get users and statuses calls we do in response to ws events.

streamer45:
> Terrific analysis! What was the setups that you compared with in the proxy vs no proxy case?

Totally missed this. I've switched between several (changing instance type and so on ) but finally settled on this:

- 3 App servers cluster (`c5.xlarge`)
- 1 DB node (`db.r5.xlarge`)
- 1 Proxy instance (`c5.xlarge`)
- 4 Agents (`t3.xlarge`), not using the first one (coordinator) so I could do 1 agent per app server in the no-proxy case.

PR #439, opened 2021-03-08
[MM-33605] Update agent binary when coming from HTTP URL

streamer45:
#### Summary

PR adds some logic to update load-test agent binary on deployment even when configured as an HTTP URL.

This really bit me. I was expecting same behaviour regardless of the file location (local vs remote). I guess I knew how it worked but after working on comparison stuff for so long I started assuming the same was happening on common deployments.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33605


isacikgoz (approved):
Looks good 🤓

deployment/terraform/assets/cluster.tf
```diff
@@ -339,10 +339,6 @@ resource "aws_instance" "loadtest_agent" {
       "while [ ! -f /var/lib/cloud/instance/boot-finished ]; do echo 'Waiting for cloud-init...'; sleep 1; done",
       "sudo apt-get -y update",
       "sudo apt-get install -y prometheus-node-exporter",
-      "wget -O tmp.tar.gz ${var.load_test_download_url}",
```
agnivade, line 342: Do we still need to pass `"-var", fmt.Sprintf("load_test_download_url=%s", loadTestDownloadURL),` now?

streamer45: Possibly not. Forgot about that, thanks :+1:

PR #442, opened 2021-03-16
[MM-33871] Provision MM builds on re-deploys

streamer45:
#### Summary

This is the companion of #439 but for MM builds. We move the provisioning logic into the app so that we can update builds coming from a remote URL in case of a re-deploy.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33871


agnivade (approved):

isacikgoz (approved):
Looks good 👍

PR #448, opened 2021-03-29
[MM-34268] Respect server config when sending typing event

streamer45:
#### Summary

Similarly to what happens on the web client, we want to send the ws typing event only if enabled on server.

#### Ticket

https://mattermost.atlassian.net/browse/MM-34268

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -531,13 +531,7 @@ func (c *SimulController) createPostReply(u user.User) control.UserActionRespons
 		rootId = post.Id
 	}
 
-	if ok, err := shouldSendTypingEvent(u, channel.Id); ok && err == nil {
-		// TODO: possibly add some additional idle time here to simulate the
-		// user actually taking time to type a post message.
-		if err := u.SendTypingEvent(channel.Id, ""); err != nil {
-			return control.UserActionResponse{Err: control.NewUserError(err)}
-		}
-	} else if err != nil {
+	if err := sendTypingEventIfEnabled(u, channel.Id); err != nil {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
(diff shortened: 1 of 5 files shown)


isacikgoz (approved):
LGTM 👍

loadtest/control/simulcontroller/actions.go
```diff
@@ -530,9 +531,13 @@ func (c *SimulController) createPostReply(u user.User) control.UserActionRespons
 		rootId = post.Id
 	}
 
-	// TODO: possibly add some additional idle time here to simulate the
-	// user actually taking time to type a post message.
-	if err := u.SendTypingEvent(channel.Id, ""); err != nil {
+	if ok, err := shouldSendTypingEvent(u, channel.Id); ok && err == nil {
+		// TODO: possibly add some additional idle time here to simulate the
+		// user actually taking time to type a post message.
+		if err := u.SendTypingEvent(channel.Id, ""); err != nil {
```
agnivade, line 537: Wondering if we should move the "should send" logic inside the main method, and rename it to `SendTypingEventIfEnabled`?

streamer45: Works for me. I've actually attempted to minimize repetition but didn't want to hide the logic inside a function. If we name it like you suggest though I guess it works since it's clear that the event gets sent based on some conditions.

streamer45: The main downside with having this as part of the main method though is that we end up adding implementation specific logic to `UserEntity`. That might not be ideal especially since we don't simply check if the event is enabled but also have some additional logic around number of channel members. It's not a big deal but I feel it might be a bit out-placed there. 

What if we just avoid the repetition by having an internal method named as you suggested but inside `simulcontroller`?

agnivade (approved):

PR #450, opened 2021-03-31
[MM-34404] Allow passing a custom config patch to deployer

streamer45:
#### Summary

PR provides a way to deploy custom MM server configs through a simple json file containing a partial config, e.g:

```json
{
  "SqlSettings": {
    "QueryTimeout": 45
  }
}
```

This will be especially useful for running automated comparisons on different server configs ([MM-33315](https://mattermost.atlassian.net/browse/MM-33315)).

#### Ticket

https://mattermost.atlassian.net/browse/MM-34404


agnivade (commented):
This is great!

The question I have is - is it possible to have different patches being applied when comparing two builds? Doesn't appear to be so if we share the same deployer config.

Would it be possible to do that?

streamer45:
> The question I have is - is it possible to have different patches being applied when comparing two builds? Doesn't appear to be so if we share the same deployer config.
> 
> Would it be possible to do that?

That's a great observation. The answer is that we'll make it possible. Ideally I'd like to decouple (and unify) all the provisioning logic into a separate package/layer to be used in both cases (single deployment and comparison). The config patch file is read on the fly so any change to the config will be reflected on a second call to `setupAppServers`.

agnivade (approved):

isacikgoz (approved):
Nice utility 👍

PR #451, opened 2021-04-07
[MM-34673] Fix nil pointer dereference crash in SimulController

streamer45:
#### Summary

PR fixes a possible crash due to `nil` pointer dereference. 
It also addresses the underlying issue that was causing the returned channel stats to be `nil`.

#### Ticket

https://mattermost.atlassian.net/browse/MM-34673


isacikgoz (approved):
Looks legit 👍

agnivade (approved):

PR #454, opened 2021-04-12
[MM-34719] Filter out system messages when randomly picking posts

streamer45:
#### Summary

At times the controllers would be trying to run actions on system messages which would fail. PR implements the necessary changes to filter out posts not directly created by users (where `Posts.Type` is set). This was already correctly handled by `memstore.RandomPostForChannelByUser()` but not by other post picking methods, hence this PR.

This is also the last fix I was planning before releasing 1.3.0 

#### Ticket

https://mattermost.atlassian.net/browse/MM-34719


isacikgoz (approved):
LGTM 👍

agnivade (approved):

PR #456, opened 2021-04-14
Update dashboard

streamer45:
#### Summary

With the bumping of the AMI, a new `node_exporter` version got installed which changed the naming of some metrics we used so some panels stopped working.  PR fixes these.


isacikgoz (approved):
Looks good 👍

PR #457, opened 2021-05-18
[MM-33179] Upgrade to terraform v0.14

streamer45:
#### Summary

PR makes the necessary changes to work with terraform v0.14
I've also removed the `minSupportedVersion` check as it was doing more harm than good. Major versions can introduce subtle changes that can cause resources not to get properly created or configured. I've received user feedback on this and decideded it makes more sense to error out if the version doesn't match the one the tool is made to work with to avoid any considerable time waste.

#### Ticket

https://mattermost.atlassian.net/browse/MM-33179
https://mattermost.atlassian.net/browse/MM-34135


.terraform.lock.hcl
```diff
@@ -0,0 +1,21 @@
+# This file is maintained automatically by "terraform init".
```
agnivade, line 1: Is this file meant to be checked in?

agnivade (approved):

.terraform.lock.hcl
streamer45: Looks that way, going by the [official documentation](https://www.terraform.io/upgrade-guides/0-14.html#provider-dependency-lock-file):

> We strongly recommend that you commit this file to your version control system

isacikgoz (approved):
👍

PR #459, opened 2021-07-15
Bump MySQL version to 2.10.0

agnivade:
No particular reason. The inner version is still 5.7.12.
Just to keep things up to date.


streamer45 (approved):
Thanks :+1:

PR #460, opened 2021-07-19
Add setting to enable performance insights on created db instances

streamer45:
#### Summary

This turned out to be pretty helpful so adding a config setting to easily use it in the future.


config/deployer.sample.json
```diff
@@ -25,7 +25,8 @@
     "InstanceEngine": "aurora-postgresql",
     "InstanceType": "db.r4.large",
     "UserName": "mmuser",
-    "Password": "mostest80098bigpass_"
+    "Password": "mostest80098bigpass_",
+    "EnablePerformanceInsights": false
```
agnivade, line 29: Hmm .. I'd actually go with default true, unless you want to make it optional? I don't see any issues with always keeping it enabled.

streamer45: I guess I wasn't sure whether enabling it caused some extra billing and thought of leaving it as an *advanced* option. I doubt most load-test users would bother with this. But it works either way really, my config will have it to `true` most of the times.

agnivade (approved):

PR #461, opened 2021-08-10
Improve error check on failing login

streamer45:
#### Summary

The returned login failure error changes from `api.user.login.invalid_credentials_email_username` to `api.user.login.invalid_credentials_sso` when an SSO method is enabled on server, leading to the users not being able to ever sign up after an initial login failure.

@sadohert 
This should fix what you've been experiencing. Unfortunately we rarely run load-test on SSO setups so this got unnoticed for a long time.

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math/rand"
 	"strconv"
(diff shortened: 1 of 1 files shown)


sadohert:
Awesome... @streamer45 !  I can confirm this has fixed my issue in my load test setup.

agnivade (approved):
Nice!

PR #462, opened 2021-08-20
Add support for custom DB parameters

streamer45:
#### Summary

Adding support for custom DB parameters. This should make it easier to run load-tests on various db configurations.


deployment/config.go
```diff
@@ -127,6 +129,18 @@ type JobServerSettings struct {
 	InstanceType string `default:"c5.xlarge"`
 }
 
+type DBParameters map[string]string
+
+func (p DBParameters) String() string {
+	var b strings.Builder
+	b.WriteString("[")
+	for k, v := range p {
+		fmt.Fprintf(&b, `{name = "%s", value = "%s"}`, k, v)
```
agnivade, line 138: Might be better written as `%q` to handle quote escaping.

agnivade (approved):

isacikgoz (approved):
LGTM 👍

agarciamontoro:
Hey, @streamer45, should we request a review again? Or is this good to go after fixing the conflicts?

streamer45:
> Hey, @streamer45, should we request a review again? Or is this good to go after fixing the conflicts?

The reason this was never merged is that I was getting some inconsistent behaviour on the Aurora/RDS side like parameters getting overridden. May need another round of checks just to verify it works as expected.

agarciamontoro:
I see, thanks for the context. Also, do you remember what the use case for this was?

streamer45:
> I see, thanks for the context. Also, do you remember what the use case for this was?

The use-case was automatically compare different DB settings. E.g. https://mattermost.atlassian.net/browse/MM-36330

agnivade:
@streamer45 - Tested this and seems to work fine. I added a parameter group, and it attached to the DB instance. I also tested manually by logging in to the DB console and running a `SHOW <param>` and it gave the expected value.

Maybe something changed with Terraform 1.0 onwards? I know we tested this long time back with 0.14 likely.

streamer45:
Do the applied settings survive db restarts? Maybe that was the issue. Anyway, no objections with moving this forward.

agnivade:
Let me test that out real quick.

agnivade:
Yep it does.

PR #463, opened 2021-09-02
Fix comparison output

streamer45:
#### Summary

We were saving only the comparison results and missing the deployment info which are quite useful for historical purposes.
Also added a sync operation on destroy which helps in certain cases where the deployment has been modified by hand.


isacikgoz (approved):
Nice, I remember I had a trouble once when I changed something on AWS. Great addition!

PR #468, opened 2021-10-15
Add ability to post permalinks

agnivade:
Also enabled the feature in deployments

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -882,14 +882,17 @@ func createMessage(u user.User, channel *model.Channel, isReply bool) (string, e
 		if err != nil && !errors.Is(err, memstore.ErrPostNotFound) {
 			return "", err
 		}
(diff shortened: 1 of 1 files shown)


loadtest/control/simulcontroller/actions.go
```diff
@@ -876,6 +876,22 @@ func createMessage(u user.User, channel *model.Channel, isReply bool) (string, e
 		message = control.AddLink(message)
 	}
 
+	// 10% of messages will contain a permalink
+	if rand.Float64() < 0.10 {
+		post, err := u.Store().RandomPostForChannel(channel.Id)
+		if err != nil && !errors.Is(err, memstore.ErrPostNotFound) {
+			return "", err
+		}
```
streamer45, line 884: Okay with not returning if the post is not found but we can't go ahead with the rest of this case or it would panic.

agnivade: :facepalm:

PR #472, opened 2021-10-29
Add cross team autoComplete APIs

agnivade:
- Load all channels at page load.
- Change the search APIs to work cross-team.


streamer45 (commented):
Thanks :+1:  What would happen if this were to run on an older MM server version? If it's not backward compatible we may want to support both ways through some basic version check. We already have everything to do this:

```go
serverVersion, _ := c.user.Store().ServerVersion()
supported, err := control.IsVersionSupported("6.2.0", serverVersion)
```

agnivade:
@streamer45 - Ready for final review now.

loadtest/control/simulcontroller/actions.go
```diff
@@ -950,11 +950,9 @@ func unreadCheck(u user.User) control.UserActionResponse {
 }
 
 func searchChannels(u user.User) control.UserActionResponse {
-	team, err := u.Store().CurrentTeam()
+	team, err := u.Store().RandomTeam(store.SelectMemberOf)
```
streamer45, line 953: Isn't this a slightly breaking change in logic for non supported versions? May be worth keeping both calls under the version check.

loadtest/control/actions.go
```diff
@@ -726,6 +726,22 @@ func Reload(u user.User) UserActionResponse {
 		}
 	}
 
+	ver, err := u.Store().ServerVersion()
+	if err != nil {
+		return UserActionResponse{Err: NewUserError(err)}
+	}
```
streamer45, line 732: Given we can assume that version won't change at runtime I am thinking we could avoid some repetition by storing it in the controller struct when we read it the first time:

https://github.com/mattermost/mattermost-load-test-ng/blob/4c6f90ce00fbec2d54b800d2c51f13443d8878d0/loadtest/control/simulcontroller/controller.go#L77

streamer45 (commented):
Great work :100: Left a couple of minor comments. Also, there seems to be some failure with tests.

agnivade:
@streamer45 - I added a new commit to move to the go install method of installing tools because the new library upgrade introduced some changes in generated code. Let me know if you prefer to have this in a separate PR.

streamer45:
> @streamer45 - I added a new commit to move to the go install method of installing tools because the new library upgrade introduced some changes in generated code. Let me know if you prefer to have this in a separate PR.

They are well separated file wise so it's not a problem to include it. Thanks :+1:

streamer45 (approved):
Looks great :+1:

streamer45 (approved):
Looks great :+1:

PR #473, opened 2021-11-30
Add support for reliable websockets

agnivade:


loadtest/user/userentity/websocket.go
```diff
@@ -98,6 +99,32 @@ func (ue *UserEntity) handlePostEvent(ev *model.WebSocketEvent) error {
...
+	}
+
+	// Now we check for sequence number, and if it does not match,
+	// we just disconnect and reconnect.
+	if ev.GetSequence() != ue.wsServerSeq {
+		mlog.Warn("Missed websocket event", mlog.Int64("got", ev.GetSequence()), mlog.Int64("expected", ue.wsServerSeq))
+		if err := ue.Disconnect(); err != nil {
+			return err
+		}
+		// TODO: reconnect?
```
agnivade, line 123: Need some guidance here. It feels a bit cumbersome to disconnect/reconnect from the user level because all the controlling happens at the controller level. Maybe we can return a custom error to the controller and try to reconnect, but even that feels complicated. The question is - should we try to handle this case?

streamer45: I'd be opting for reconnecting without the controller even noticing. Later on we could implement a reconnect callback or some other system to notify the caller but that's not currently required. I am thinking it would be sufficient for `wsEventHandler()` to return a special error and then upon receiving that in `listen()` we reinitialize the client. We can probably even factor out and reuse the existing code which already handles reconnecting:

https://github.com/mattermost/mattermost-load-test-ng/blob/be76149215d0d018c5e2ec5814c74d49ef1db937/loadtest/user/userentity/websocket.go#L117-L132

agnivade: Thanks, will take a look.

agnivade: PTAL.

streamer45: Labels :open_mouth:

agnivade: Lol yeah, there are 2 nested loops here. I can work on an improvement PR after this if you want.

streamer45: It looks clean enough actually. Just need to test it out a bit but code changes look good.

streamer45 (approved):
Thanks for the work here, highly appreciated. Tested a bit locally and looks good. Will give it a proper spin soon but unblocking things for now.

agnivade:
Sounds good. I also tested locally with both the seq number mismatch case and the reconnect case, and both were working.

PR #474, opened 2021-12-16
Implement `deletePost` action

streamer45:
#### Summary

Not the most frequent of actions but, as we are reminded as of late, anything dealing with posts can be important from a performance perspective.


agnivade (approved):
Nice!

PR #475, opened 2022-02-07
MM-41499: Use CRT switch in GetTeamsUnread

agnivade:
Previously, this was just hardcoded to false which
made it skip the CRT code paths. Now we actually
get the param from the store and pass that onwards.

https://mattermost.atlassian.net/browse/MM-41499


streamer45 (approved):
Nice catch :+1:

ashishbhate (approved):

PR #476, opened 2022-02-08
Fix channel switcher bug

agnivade:
We get channels for the user only after the curren team is set.
Without that a newly signed up user won't be part of any channels
unless they join a team first.


streamer45 (approved):
Nice :+1:

PR #477, opened 2022-02-08
MM-36045: Add support for CRT features

ashishbhate:
#### Summary

- Adds support for the following CRT features in simulcontroller
  - View Global Threads Screen - All threads and Unreads
  - Follow and unfollow a thread
  - View a thread and posts in that thread - paginated version
  - Mark all threads in team as read (the little icon near the top right of the global threads screen)
  - Update read state of a thread
- Long running threads in the simulcontroller - threads with many replies and a channel-wide mention in the root post

  
#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-36045


agnivade:
The go.tools.* files are all deleted. Feel free to remove them.

agnivade:
Great work! Before I start a review, I am curious to see a load-test with this on master with TAF on vs off.

streamer45 (commented):
Thanks @ashishbhate for the thorough work :+1: 
I gave this a first pass with some comments and suggestions.

go.mod
```diff
@@ -10,19 +10,19 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.4.2
 	github.com/mattermost/ldap v3.0.4+incompatible // indirect
-	github.com/mattermost/mattermost-server/v6 v6.0.0-20211029033710-ec56cac42da7
+	github.com/mattermost/mattermost-server/v6 v6.0.0-20220204061906-8223968576c7
```
agnivade, line 13: Is this the hash to my PR branch (https://github.com/mattermost/mattermost-server/pull/19485/files)? Doesn't seem to match though :thinking: 

In any case, this is tricky to handle. We need to test along with my branch, and then merge the server PR, update the hash here, and merge the load-test PR.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1282,3 +1300,357 @@ func sendTypingEventIfEnabled(u user.User, channelId string) error {
...
+	}
+
+	return control.UserActionResponse{Info: "Visited Global Threads Screen"}
+}
+
+func (c *SimulController) followThread(u user.User) control.UserActionResponse {
+	collapsedThreads, resp := control.CollapsedThreadsEnabled(u)
+	if resp.Err != nil || !collapsedThreads {
+		return resp
+	}
```
agnivade, line 1411: Wondering if we can create a middleware function to gate all CRT related calls? That way we don't have to repeat this check everywhere.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1282,3 +1300,357 @@ func sendTypingEventIfEnabled(u user.User, channelId string) error {
...
+	post, err := u.Store().RandomPostForChannel(channel.Id)
+	if err != nil {
+		if errors.Is(err, memstore.ErrPostNotFound) {
+			return control.UserActionResponse{Info: "followThread: can't find post to follow in channel"}
+		}
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	try := 0
+	for try < 100 && post.RootId != "" {
```
agnivade, line 1429: I don't understand why we are repeating this 100 times. Can you expand a bit?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1282,3 +1300,357 @@ func sendTypingEventIfEnabled(u user.User, channelId string) error {
...
+	channel, err := u.Store().Channel(thread.Post.ChannelId)
+	if err != nil || channel == nil {
+		err = u.GetChannel(thread.Post.ChannelId)
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+		channel, err := u.Store().Channel(thread.Post.ChannelId)
+		if err != nil || channel == nil {
+			return control.UserActionResponse{Err: control.NewUserError(errors.New("unfollow thread: can't get channel for thread"))}
+		}
```
agnivade, line 1473: Let's separate these 2 cases. In the case of `err != nil`, we need to bubble up the error.

agnivade (commented):
Did an initial first pass.

ashishbhate:
>Great work! Before I start a review, I am curious to see a load-test with this on master with TAF on vs off.

Yep, agreed! Any metric you're interested in seeing in particular? I've been focussed on confirming that the CRT api calls are actually being made when CRT is turned on vs off.

go.mod
ashishbhate: Yep, its pointing to your PR branch so I could get the new methods imported. I will update once your PR is merged.

loadtest/control/simulcontroller/actions.go
```diff
@@ -661,6 +672,10 @@ func (c *SimulController) createPost(u user.User) control.UserActionResponse {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
+	if shouldLongThread {
+		st.setLongRunningThread(postId, channel.Id, channel.TeamId)
+	}
```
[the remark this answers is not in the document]
ashishbhate: Its a bit tricky because I don't have the final `postID` returned from `CreatePost`. I'll need the post id in `createMessage` to set it as a long running thread.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1282,3 +1300,357 @@ func sendTypingEventIfEnabled(u user.User, channelId string) error {
...
+			PageSize:   25,
+			Extended:   false,
+			Deleted:    false,
+			Unread:     false,
+			Since:      0,
+			TotalsOnly: false,
+			Before:     oldestThreadId,
+		})
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
```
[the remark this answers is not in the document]
ashishbhate: Yep! I forgot to push some recent changes, including this one.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1282,3 +1300,357 @@ func sendTypingEventIfEnabled(u user.User, channelId string) error {
...
+		idleTime := time.Duration(1+rand.Intn(10)) * time.Second
+		select {
+		case <-c.stopChan:
+			return control.UserActionResponse{Info: "action canceled"}
+		case <-time.After(idleTime):
+		}
+	}
+
+	// Switch to "Unread" tabs
+	unreadThreads, err := u.Store().ThreadsSorted(true, false)
```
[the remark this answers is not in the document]
ashishbhate: The API end-point is the same, the parameter `unread` option is the only thing that changes. I couldn't figure out how to filter the metric by parameter in grafana to correctly weight the API calls. 
Since the "unreads" tab is only accessible through the Global Threads Screen, I added both calls in the same action.

loadtest/control/simulcontroller/actions.go
ashishbhate: We're only looking for a reply post (i.e. an actual thread) so that we can follow that thread. We can theoretically follow a root post that is not yet a thread, but I wanted to replicate user actions more accurately. 
100 times because I didn't want to spin infinitely looking for a random post that is also a reply. 

As @streamer45 said, its better if I made a better method in the store.

loadtest/control/simulcontroller/state.go
```diff
@@ -0,0 +1,58 @@
...
+	"sync"
+)
+
+type ThreadInfo struct {
+	Id        string
+	ChannelId string
+	TeamId    string
+}
+
+type state struct {
```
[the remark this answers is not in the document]
ashishbhate: Each user has a separate store. I need a way for all users to share some data - in this case, common threads that they should all post replies to, so that we can get truly big threads.  

I noticed this pattern in the generative controller so I replicated here. Do you have something else in mind to share data across users?

loadtest/control/simulcontroller/controller.go
```diff
@@ -186,6 +186,30 @@ func (c *SimulController) Run() {
...
+			frequency: 4.8,
+		},
+		{
+			run:       c.markAllThreadsInTeamAsRead,
+			frequency: 0.013,
+		},
+		{
+			run:       c.updateThreadRead,
+			frequency: 1.17,
+		},
```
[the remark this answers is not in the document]
ashishbhate: I found the daily average number of invocations for the calls on community. The frequency here is calculated by weighing those numbers in proportion to frequency of existing calls here.

The frequency for actions with a larger number of invocations were not in proportion to the frequency for actions with a smaller number of invocations (e.g. switchChannel vs addReaction)

So the frequency of calls for `viewThread` and `viewGlobalThreads` is weighted with respect to `switchChannel` and the rest with respect to `createPost`.

ashishbhate (commented):
Thanks for the first pass @streamer45 and @agnivade. I have responded to some comments inline asking for feedback. For comments not replied to, I am making the suggested changes locally.

loadtest/control/simulcontroller/actions.go
agnivade: Can we get a random post, and if it's a root post, reply to it and follow it. Trying something 100 times feels a bit inelegant.

agnivade:
> Yep, agreed! Any metric you're interested in seeing in particular? I've been focussed on confirming that the CRT api calls are actually being made when CRT is turned on vs off.

Nothing in particular. Just want a general pass and see the behavior of the server in CRT on vs off, now that we have proper CRT coverage. I'd expect CRT off case to perform slightly better.

loadtest/store/memstore/store.go
```diff
@@ -974,3 +986,113 @@ func (s *MemStore) SetServerVersion(version string) error {
...
+	cloneThreadResponse(thread, t)
+	s.threads[thread.PostId] = t
+
+	return nil
+}
+
+// SetThreads stores the given thread response as a thread.
+func (s *MemStore) SetThreads(trs []*model.ThreadResponse) error {
+	if len(trs) == 0 {
+		mlog.Info("No threads to add to store")
```
agnivade, line 1014: Let's avoid logging in the store. It's fine to skip this log line.

loadtest/control/simulcontroller/state.go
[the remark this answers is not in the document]
streamer45: I see, we are trying to force a common behaviour. From my experience load-testing I've never thought that creating huge threads was ever an issue, I think it happened naturally as the result of replying to recent posts/replies. 

Either way, I am thinking we could still write some heuristic based on what we have in the store (maybe leveraging `ReplyCount`?) to achieve a similar behaviour without having direct communication between users. Overall I feel it's a bit of an anti-pattern which in the case of the generative controller is accepted cause its purpose is to purely create some data but here it doesn't look great as it further complicates an already elaborate logic.

loadtest/control/simulcontroller/actions.go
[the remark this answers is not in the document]
agnivade: Fair enough, let's leave it.

loadtest/store/memstore/random.go
```diff
@@ -331,3 +332,21 @@ func pickRandomKeyFromMap(m interface{}) (interface{}, error) {
...
+	defer s.lock.RUnlock()
+
+	if len(s.threads) == 0 {
+		return model.ThreadResponse{}, ErrThreadNotFound
+	}
+
+	threadIds := make([]string, 0, len(s.threads))
+	for _, t := range s.threads {
+		threadIds = append(threadIds, t.PostId)
+	}
```
agnivade, line 348: We don't really need to do this business here. `pickRandomKeyFromMap` is there for you :)

loadtest/store/memstore/store.go
```diff
@@ -974,3 +986,113 @@ func (s *MemStore) SetServerVersion(version string) error {
...
+		return errors.New("memstore: thread should not be nil")
+	}
+
+	// We get an element from the queue and check if we have it in the map and
+	// if it points to the same memory location. If so, we delete it since it means the queue is full.
+	// This is done to keep the data pointed by the map consistent with the data stored in the queue.
+	t := s.threadsQueue.Get().(*model.ThreadResponse)
+	if tt, ok := s.threads[t.PostId]; ok && tt == t {
+		delete(s.threads, t.PostId)
+	}
```
agnivade, line 1004: Some boilerplate which I am looking forward to remove with generics :)

loadtest/store/memstore/store.go
```diff
@@ -974,3 +986,113 @@ func (s *MemStore) SetServerVersion(version string) error {
...
+	return threads, nil
+}
+
+func cloneThreadResponse(src *model.ThreadResponse, dst *model.ThreadResponse) *model.ThreadResponse {
+	dst.PostId = src.PostId
+	dst.ReplyCount = src.ReplyCount
+	dst.LastReplyAt = src.LastReplyAt
+	dst.LastViewedAt = src.LastViewedAt
+	dst.Participants = src.Participants
+	dst.Post = src.Post
```
agnivade, line 1059: These are a slice and a post respectively. Are we fine doing a shallow copy? If so, we can just derefence the struct and assign it to the target.

loadtest/store/memstore/store.go
```diff
@@ -974,3 +986,113 @@ func (s *MemStore) SetServerVersion(version string) error {
...
+	sort.Slice(threads, func(i, j int) bool {
+		if asc {
+			return threads[i].LastReplyAt < threads[j].LastReplyAt
+		}
+		return threads[i].LastReplyAt > threads[j].LastReplyAt
+	})
+	return threads, nil
+}
+
+func cloneThreadResponse(src *model.ThreadResponse, dst *model.ThreadResponse) *model.ThreadResponse {
```
agnivade, line 1053: We usually take source and return a cloned struct. Taking dst in the parameter and returning it again seems slightly unidiomatic.

loadtest/store/memstore/store.go
```diff
@@ -974,3 +985,112 @@ func (s *MemStore) SetServerVersion(version string) error {
...
+// MarkAllThreadsInTeamAsRead marks all threads in the given team as read
+func (s *MemStore) MarkAllThreadsInTeamAsRead(teamId string) error {
+	s.lock.RLock()
+	threads, err := s.getThreads(false)
+	if err != nil {
+		return err
+	}
+	now := model.GetMillis()
+	for _, thread := range threads {
+		ch, err := s.Channel(thread.Post.ChannelId)
```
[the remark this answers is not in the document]
agnivade: Yep, beware of nested read locks. Nasty ones.

agnivade (commented):
Made the final pass. Great work!

loadtest/control/simulcontroller/state.go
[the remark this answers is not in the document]
ashishbhate: @streamer45 
How about I move this logic to the generative controller? That way people wanting to test loads for long running threads can first run the generative controller to create long threads, and the logic in the simulative controller remains mostly unchanged.

loadtest/store/memstore/store.go
ashishbhate: Yeah, this is a bit iffy. I did it this way so that I could:

1. directly call this function in the parent's return statement, i.e. `return cloneThreadResponse(...)`
2. use inplace, e.g. `append(threads, cloneThreadResponse(thread, &model.ThreadResponse{}))`
3. pass the threadResponse object in the case where I need to update an existing object, i.e. in `Memstore.SetThread`

loadtest/control/simulcontroller/state.go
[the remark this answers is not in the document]
streamer45: That works as well. We can always come back to this later on if needed.

lieut-data (changes requested):
Thanks @agnivade & @streamer45 for the detailed first pass! @ashishbhate,  I'm going to mark my own review as blocked while we iterate, but feel free to re-request my review when needed :)

agnivade:
@ashishbhate - feel free to ping me when it's ready to review.

ashishbhate:
This is ready for another round of reviews. I've addressed most comments, others are still WIP but they are code style issues and not functional issues. I'll address them soon. 

So far I've only been running the coordinator load test on my separate build machine. The difference between this PR and master isn't as large as I expected.

These are the results from a quick run last night:

This PR:
```
info  [2022-03-10 22:42:40.060 +05:30] estimated number of supported users is 105    caller="coordinator/coordinator.go:142"
```

Master:
```
info  [2022-03-10 23:27:33.487 +05:30] estimated number of supported users is 116    caller="coordinator/coordinator.go:142"
```

The server was running with `"ThreadAutoFollow": true` and `"CollapsedThreads": "default_on"`


I'm going run a full load test through terraform over the weekend to get a better comparison.

agnivade (approved):
Nicely done! Would be good to see a demo of this somewhere :wink:

loadtest/store/memstore/random.go
```diff
@@ -331,3 +351,15 @@ func pickRandomKeyFromMap(m interface{}) (interface{}, error) {
 	idx := rand.Intn(len(keys))
 	return keys[idx].Interface(), nil
 }
+
+// RandomThread returns a random post.
+func (s *MemStore) RandomThread() (model.ThreadResponse, error) {
+	threads, err := s.getThreads(false)
```
streamer45, line 357: Should we read lock before getting the threads?

loadtest/store/memstore/store.go
```diff
@@ -974,3 +985,123 @@ func (s *MemStore) SetServerVersion(version string) error {
...
+		threads = append(threads, cloneThreadResponse(thread, &model.ThreadResponse{}))
+	}
+	return threads, nil
+}
+
+// ThreadsSorted returns all threads, sorted by LastReplyAt
+func (s *MemStore) ThreadsSorted(unreadOnly, asc bool) ([]*model.ThreadResponse, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+	threads, err := s.getThreads(unreadOnly)
```
streamer45, line 1038: Unless necessary I think we should unlock as soon as we get the threads and do the sorting outside.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1282,3 +1282,365 @@ func sendTypingEventIfEnabled(u user.User, channelId string) error {
 	}
 	return nil
 }
+
+func (c *SimulController) viewGlobalThreads(u user.User) control.UserActionResponse {
+	collapsedThreads, resp := control.CollapsedThreadsEnabled(u)
```
lieut-data, line 1287: I know this logic already exists, but do we have a configurable ratio for users with CRT enabled vs. disabled? It would be neat to test at 100%, 50%, etc.

ashishbhate: Currently we don't have a way to to configure the ratio. Every user is set to use CRT through the `default_on` value for the `ServiceSettings.CollapsedThreads` server setting.

This is definitely an improvement I'd like to make in the future. Currently, even with CRT off, the global frequency of actions takes into account the frequency of CRT actions when choosing which action to execute. This check makes the CRT action a no-op.
Ideally, if CRT is off, the action chooser should not consider CRT actions as valid actions in the list of possible actions.

loadtest/control/gencontroller/actions.go
```diff
@@ -179,20 +179,28 @@ func (c *GenController) createPost(u user.User) control.UserActionResponse {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
+	channelMention := ""
+	shouldLongThread := shouldMakeLongRunningThread((channel.Id))
+	if shouldLongThread {
+		channelMention = control.PickRandomString([]string{"@all ", "@here ", "@channel "})
```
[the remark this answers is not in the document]
ashishbhate: This is leftover from the time this was in the simulcontroller but I've opted to leave it in while moving to the gencontroller.

Channel mentions in the root post of a thread cause all channel members to become followers of that thread (when it becomes a thread). For large channels (e.g. "Town Square" or any announcement channel), this put pressure on our server notification code, potentially causing OOM errors. We have fixed our code to be more careful by processing changes in batches, and not spawning an uncontrolled number of goroutines. 

I left this in to organically replicate heavy load on the server when users reply to threads with a lot of followers.

loadtest/control/gencontroller/utils.go
```diff
@@ -31,3 +31,18 @@ func pickAction(actions map[string]userAction) (*userAction, error) {
 	}
 	return nil, errors.New("should not be able to reach this point")
 }
+
+// shouldMakeLongRunningThreads returns if a long thread should be created
+// TODO: The rates and logic in this function should be made configurable
+func shouldMakeLongRunningThread(channelId string) bool {
+	// 2% of the the time we check if we should make a long running thread
+	// this way we don't make all long running threads near the start
```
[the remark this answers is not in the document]
ashishbhate: This is also left over from the time this code was in the simulcontroller. I didn't want all long running threads to be created near the start. This wouldn't be an accurate representation of normal server usage. I wanted the long running threads to be created over time so that the interactions with these threads would also be spread out over time.

I suppose this doesn't matter in the gencontroller as much as it did in the simulcontroller, though I still think its better for the creation to be spread out over a period of time. Of course, the 2% check isn't very realistic, thus the comment above the function that the rate and logic in this function should be configurable. 
We also only create one long running thread per channel, which isn't very realistic either, but I think it works. I didn't want to overload the config with too many settings, but we will have to eventually.

streamer45 (approved):
Brilliant work :+1:

PR #479, opened 2022-03-05
add config validation in coordinator for MaxActiveUsers

97amarnathk:
#### Summary
Validate UsersConfig.MaxActiveUsers is not less than clusterConfig.MaxActiveUsers

#### Ticket Link
Github: https://github.com/mattermost/mattermost-server/issues/19676
JIRA: https://mattermost.atlassian.net/browse/MM-42253

### DIFF
--- a/api/coordinator_test.go
+++ b/api/coordinator_test.go
@@ -97,6 +97,7 @@ func TestCoordinatorAPI(t *testing.T) {
 		}
 
 		data.CoordinatorConfig.ClusterConfig.Agents[0].ApiURL = server.URL
+		data.CoordinatorConfig.ClusterConfig.MaxActiveUsers = 1000
 
 		obj := e.POST("/create").WithQuery("id", id).WithJSON(data).
 			Expect().Status(http.StatusCreated).
--- a/api/coordinator_test.go
+++ b/api/coordinator_test.go
@@ -40,6 +40,7 @@ func TestCoordinatorAPI(t *testing.T) {
 	config.MonitorConfig.Queries[0].Description = "Query"
 	config.MonitorConfig.Queries[0].Query = "query"
 	config.ClusterConfig.Agents[0].ApiURL = server.URL
+	config.ClusterConfig.MaxActiveUsers = 100
 
 	t.Run("create/destroy", func(t *testing.T) {
 		data := struct {
@@ -97,7 +98,6 @@ func TestCoordinatorAPI(t *testing.T) {
 		}
 
 		data.CoordinatorConfig.ClusterConfig.Agents[0].ApiURL = server.URL
-		data.CoordinatorConfig.ClusterConfig.MaxActiveUsers = 1000
 
 		obj := e.POST("/create").WithQuery("id", id).WithJSON(data).
 			Expect().Status(http.StatusCreated).
--- a/coordinator/coordinator.go
+++ b/coordinator/coordinator.go
@@ -254,7 +254,6 @@ func New(config *Config, ltConfig loadtest.Config, log *mlog.Logger) (*Coordinat
 
 	clusterConfig := config.ClusterConfig
 
-	clusterConfig.IsValid(ltConfig)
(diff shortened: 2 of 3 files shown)


streamer45 (changes requested):
Thanks @97amarnathk, a couple of suggestions:

This wasn't specified in the ticket (sorry about that) but given the coordinator can run multiple agents the check should be against the product of the value and the number of agents available. In this case:

` ltConfig.UsersConfiguration.MaxActiveUsers*len(clusterConfig.Agents)`

Now, given the check is constrained to `LoadAgentClusterConfig` I am thinking we could follow our convention and add a new `IsValid() error` method in `coordinator/cluster/config.go` to be called by the coordinator. 

Let me know if that makes sense.

97amarnathk:
@streamer45 I have added the change in validation for max active users. 
Since the validation is not constrained only on LoadAgentClusterConfig (also depends on ltConfig), I am passing ltConfig as a param to clusterConfig.IsValid

coordinator/coordinator.go
```diff
@@ -252,7 +252,14 @@ func New(config *Config, ltConfig loadtest.Config, log *mlog.Logger) (*Coordinat
 		return nil, fmt.Errorf("could not validate configuration: %w", err)
 	}
 
-	cluster, err := cluster.New(config.ClusterConfig, ltConfig, log)
+	clusterConfig := config.ClusterConfig
+
+	clusterConfig.IsValid(ltConfig)
```
streamer45, line 257: This line can go away I assume.

streamer45 (commented):
Thanks @97amarnathk , one more thing but overall looks great.

We should also fix the failing test, just need to set `MaxActiveUsers` in:

https://github.com/mattermost/mattermost-load-test-ng/blob/7ba3f8a72de583e743fedcf1bf6c4e0358d25d44/api/coordinator_test.go#L42

97amarnathk:
@streamer45 In the coordinator test, I am setting MaxActiveUsers with the default value (1000)

api/coordinator_test.go
```diff
@@ -97,6 +97,7 @@ func TestCoordinatorAPI(t *testing.T) {
 		}
 
 		data.CoordinatorConfig.ClusterConfig.Agents[0].ApiURL = server.URL
+		data.CoordinatorConfig.ClusterConfig.MaxActiveUsers = 1000
```
streamer45, line 100: We need to set this above and the value should be 100 since that's what we are using for the agent:

```patch
diff --git a/api/coordinator_test.go b/api/coordinator_test.go
index 187227d..c67f69d 100644
--- a/api/coordinator_test.go
+++ b/api/coordinator_test.go
@@ -40,6 +40,7 @@ func TestCoordinatorAPI(t *testing.T) {
        config.MonitorConfig.Queries[0].Description = "Query"
        config.MonitorConfig.Queries[0].Query = "query"
        config.ClusterConfig.Agents[0].ApiURL = server.URL
+       config.ClusterConfig.MaxActiveUsers = 100
 
        t.Run("create/destroy", func(t *testing.T) {
                data := struct {
@@ -97,7 +98,6 @@ func TestCoordinatorAPI(t *testing.T) {
                }
 
                data.CoordinatorConfig.ClusterConfig.Agents[0].ApiURL = server.URL
-               data.CoordinatorConfig.ClusterConfig.MaxActiveUsers = 1000
 
                obj := e.POST("/create").WithQuery("id", id).WithJSON(data).
                        Expect().Status(http.StatusCreated).
```

By the way you can also run `make test` locally to verify it works.

agnivade:
@97amarnathk - Are you still interesting in working on this?

agarciamontoro:
Picking this where @97amarnathk left it to try and merge it.

agnivade:
ping @streamer45

agarciamontoro (approved):
This looks good to me now :tada: Let's wait for Claudio (he's on vacation) before we merge this.

streamer45 (approved):
Looks good :+1:

PR #480, opened 2022-03-10
make lint exclusions more specific

ashishbhate:
#### Summary

- Make lint exclusions more specific to issues in our code instead of skipping entire folders. This should prevent more lint errors from creeping in until we fix the existing issues. 

#### Ticket Link

- NA

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -50,9 +50,8 @@ linters:
 issues:
   exclude-rules:
     - linters:
-      # ignore golint error for a lot of packages for now
-      # This is mainly used to avoid errors regarding ids
+      # Used to avoid errors regarding ids
       # We can remove this section once we decide to fix them
       - golint
-      path: "loadtest|coordinator|logger|cmd|api|websocket|terraform"
+      text: "should be.*ID"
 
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -50,9 +50,8 @@ linters:
 issues:
   exclude-rules:
     - linters:
-      # ignore golint error for a lot of packages for now
-      # This is mainly used to avoid errors regarding ids
+      # Used to avoid errors regarding naming conventions
       # We can remove this section once we decide to fix them
       - golint
-      path: "loadtest|coordinator|logger|cmd|api|websocket|terraform"
+      text: "should be.*ID|CamelCase$|FileJSON`$|ConsoleJSON`$|APIURL`$|calling this Response$"
```
streamer45, line 56: What are the backticks for?

ashishbhate: The error message looks like 

```
logger/logger.go:17:2: struct field `ConsoleJson` should be `ConsoleJSON` (golint)
```

It's not part of the regex. It's part of the error message. I'm including it to anchor the exclusion regex to the end of the sentence.

streamer45: Thanks :+1:

streamer45 (approved):

PR #485, opened 2022-04-12
MM-43067: updates for CRT

ashishbhate:
#### Summary

- Only get threads when scrolling or viewing threads
- Pass CollapsedThreads parameter when viewing a channel

#### Ticket Link

- https://mattermost.atlassian.net/browse/MM-43067


loadtest/control/simulcontroller/actions.go
```diff
@@ -1604,12 +1614,11 @@ func (c *SimulController) updateThreadRead(u user.User) control.UserActionRespon
...
-			Extended:   false,
-			Deleted:    false,
-			Unread:     false,
-			Since:      0,
-			TotalsOnly: false,
+			PageSize: 25,
+			Extended: false,
+			Deleted:  false,
+			Unread:   false,
+			Since:    0,
```
lieut-data, line 1621: Is the omission of both `TotalsOnly` and `ThreadsOnly` here intentional? (Feels like the pattern above is to be explicit.)

lieut-data (approved):

loadtest/control/simulcontroller/actions.go
ashishbhate: Yep, intentional. This is an edge case - we make this API call only when the simulated user's threads aren't already loaded, so we need to fetch both the counts and the threads. 

I omitted both since they need to be false (which is the default). I can see why explicit would be better. I'll add them back in.

koox00:
> We don't make this API call in the webapp on team switch, so I'm not sure why we're doing it in the load tests. I haven't updated the load-test to fetch the counts the way the webapp does it. I think this merits some deeper investigation to check if the loadtests are replicating client behaviour exactly.

I think that we do fetch all teams' unread counts on switch, which fetches also thread counts but excluding DM/GM channels. 
And we also fetch current team's thread unread counts which includes DM/GMs.
At least that's what we used to do.

ashishbhate:
@koox00 I've added fetching of thread counts for that team, on team switch.

koox00 (approved):

PR #486, opened 2022-04-21
Added sidebar category coverage

agnivade:
GetSidebarCategories are called for every team load.

Create/Update SidebarCategories are called at the same rate
as delete posts. Verified from Community Grafana.


agnivade:
First step before deciding to add the index.

loadtest/control/simulcontroller/actions.go
```diff
@@ -556,6 +559,78 @@ func deletePost(u user.User) control.UserActionResponse {
...
+	// We pick a random channel from first category and move to second category.
+	channelToMove := control.PickRandomString(cat1.Channels)
+
+	// Find index
+	i := 0
+	for i = range cat1.Channels {
+		if cat1.Channels[i] == channelToMove {
+			break
+		}
+	}
```
streamer45, line 622: We can probably make this a util function since it's pretty generic.

loadtest/control/simulcontroller/actions.go
```diff
@@ -556,6 +559,78 @@ func deletePost(u user.User) control.UserActionResponse {
...
+	// Move from the first, and add to second.
+	cat1.Channels = append(cat1.Channels[:i], cat1.Channels[i+1:]...)
+	cat2.Channels = append(cat2.Channels, channelToMove)
```
streamer45, line 625: I am guessing this is safe for now but given we don't really copy I am thinking it could become a problem if we were to do anything on them from another goroutine (e.g. the websocket side).

streamer45 (approved):
Looks good, nothing blocking :+1:

loadtest/control/simulcontroller/actions.go
agnivade: yeah good point

agnivade:
One more round please. Just as a sanity check.

agnivade:
Gentle ping @streamer45

streamer45 (approved):
Looks good :+1:

PR #490, opened 2022-05-17
Update postgres version to 11.15 and mysql to 2.10.2

jwilander:
#### Summary
The postgres version went out of support and figured it was good to update mysql to the latest dot version too.


deployment/terraform/assets/variables.tf
```diff
@@ -60,8 +60,8 @@ variable "db_instance_class" {
 variable "db_engine_version" {
   type = map(any)
   default = {
-    "aurora-mysql"      = "5.7.mysql_aurora.2.10.0"
-    "aurora-postgresql" = "11.7"
```
streamer45, line 64: There was a reason we used this specific version as it's what Cloud adopted at the time. It would be interesting to know what they have landed on as of late.

streamer45 (approved):
Thanks :+1:

deployment/terraform/assets/variables.tf
jwilander: @stylianosrigas what version are you guys on in cloud these days?

stylianosrigas: We are on 12.8 for PostgreSQL.

jwilander: Ok, I'll merge this as is for now but it might be worth exploring bumping to 12.8 to match cloud.

PR #491, opened 2022-05-17
Remove blank query from coordinator default config

jwilander:
#### Summary
Not sure if this is the right solution or not but the default coordinator config failed on:

```
Error: failed to create coordinator: coordinator: coordinator api request error: could not create coordinator: could not validate configuration: Description is empty
```

if I didn't copy the sample coordinator config during the steps [here](https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/terraform_loadtest.md#optionally-configure-coordinator-and-load-test) which says it's optional to do.

The other option would be to add some of the queries from the sample in the code as a default.

### DIFF
--- a/api/client_test.go
+++ b/api/client_test.go
@@ -199,8 +199,6 @@ func TestCoordClientConcurrency(t *testing.T) {
 		defaults.Set(&ltConfig)
 		coordConfig.ClusterConfig.Agents[0].Id = coord.Id() + "-agent"
 		coordConfig.ClusterConfig.Agents[0].ApiURL = server.URL
-		coordConfig.MonitorConfig.Queries[0].Description = "Query"
-		coordConfig.MonitorConfig.Queries[0].Query = "query"
 		_, err := coord.Create(&coordConfig, &ltConfig)
 		require.NoError(t, err)
 		return coord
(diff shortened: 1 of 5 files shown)


jwilander:
Hmm from the tests failing I see that there's an assumption made that at least one query is present. Should I add a sample query as a default then?

agnivade:
It is failing here: https://github.com/mattermost/mattermost-load-test-ng/blob/2947baafc48e8ccada0096255c288b436037ac26/coordinator/coordinator_test.go#L25-L26. I think a length check should fix it.

streamer45:
> Hmm from the tests failing I see that there's an assumption made that at least one query is present. Should I add a sample query as a default then?

I think originally it didn't make sense for coordinator to be executed with no queries since it's main purpose was to figure out the supported users count. Things have evolved slightly since then and now there's a valid use case (bounded load-test).

I think we can just remove all instances of:

```go
coordConfig.MonitorConfig.Queries[0].Description = "Query"
coordConfig.MonitorConfig.Queries[0].Query = "query"
```

from the various tests and should be good.

jwilander:
Tests are green after removing those

isacikgoz (approved):

streamer45 (approved):
Looks good :+1:

PR #495, opened 2022-06-07
Using a newer DB class

agnivade:
r4 is long outdated.


streamer45 (approved):
Interestingly I've never used that default once.

PR #496, opened 2022-06-10
Add coverage for custom status

agnivade:


agnivade:
Did something different than what we discussed :P

loadtest/control/simulcontroller/controller.go
```diff
@@ -170,6 +170,14 @@ func (c *SimulController) Run() {
 			run:       deletePost,
 			frequency: 0.06,
 		},
+		{
+			run:       c.updateCustomStatus,
+			frequency: 0.05,
```
agnivade, line 175: Checked Community Grafana. It has nearly the same frequency as deletePost, may be slightly slower. Hence made it to 0.05.

streamer45 (approved):
Looks good :+1:

PR #498, opened 2022-07-21
MM-45829: Bump postgres version to 12.11

agarciamontoro:
#### Summary

I have executed two bounded load tests, `base` with Postgres v11.15 and `new` with v12.11. Both have been running for one hour after all the users were connected. The tests were executed with the following configuration:

- App instance type: `c5.2xlarge`
- DB instance type: `db.r5.2xlarge`
- Mattermost version: v7.1.1
- Dataset: DB bump from 6.1

There seems to be a non-negligible improvement in the API, store times and requests per second, while there's not any significant change in the rest. In conclusion, it's safe to bump the version and merge this PR.

avg-api-times
p99-api-times
avg-store-times
p99-store-times
rps
heap-in-use
stack-in-use
cpu-utilization
goroutines-in-use

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-45829


agnivade:
Wow!! Best news I've heard all week! This was super-unexpected.

cc @streamer45

agnivade (approved):

PR #500, opened 2022-07-29
Add coverage for post reminders

agnivade:


loadtest/control/simulcontroller/controller.go
```diff
@@ -162,6 +162,10 @@ func (c *SimulController) Run() {
 			run:       searchPosts,
 			frequency: 0.1,
 		},
+		{
+			run:       c.createPostReminder,
+			frequency: 0.1,
```
agnivade, line 167: This is just made-up for now. We'll replace this with the right number once the feature is live on community.

agnivade:
Strange, what has happened to golangci? It is failing in parts of code untouched by the PR.

agnivade:
@streamer45 - Do you know what's happening? golangci is failing randomly.

streamer45:
> @streamer45 - Do you know what's happening? golangci is failing randomly.

I'd suggest bumping the CI to go 1.18 and see if it fixes it.

agnivade:
Yeah, that's what I was thinking next. But strange that it would start to fail randomly

agnivade:
Man this gets harder, the cimg/go:1.18 has a new directory structure so installation is failing. I'll fix that on Monday. In the mean while, PR is good to review.

agnivade:
Wow finally. I had to upgrade Go and golangci to get it to work. @lieut-data also encountered the same issue at https://github.com/mattermost/mattermost-plugin-api/pull/119

Makefile
```diff
@@ -84,11 +84,7 @@ verify-gomod:
 check-style: golangci-lint
 
 golangci-lint:
-# https://stackoverflow.com/a/677212/1027058 (check if a command exists or not)
-	@if ! [ -x "$$(command -v golangci-lint)" ]; then \
-		echo "golangci-lint is not installed. Please see https://github.com/golangci/golangci-lint#install for installation instructions."; \
-		exit 1; \
-	fi; \
+	$(GO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.2
```
agarciamontoro, line 87: Should we run this only if golangci-lint is not present?

.circleci/config.yml
```diff
@@ -1,53 +1,11 @@
...
-      - save_cache:
-          name: Save golangci-lint cache
-          key: '<< parameters.prefix >>-golangci-lint-{{ arch }}-<< parameters.version >>'
-          paths: [<< parameters.gobin >>/golangci-lint]
-
 jobs:
   test:
     docker:
-      - image: circleci/golang:1.17
+      - image: cimg/go:1.18.4
```
agarciamontoro, line 6: Much cleaner :ok_hand:

loadtest/control/simulcontroller/controller.go
agarciamontoro: Should we create a ticket so we don't forget?

loadtest/user/userentity/actions.go
```diff
@@ -1329,3 +1329,15 @@ func (ue *UserEntity) RemoveCustomStatus(userID string) error {
...
+func (ue *UserEntity) CreatePostReminder(userID, postID string, targetTime int64) error {
+	_, err := ue.client.SetPostReminder(&model.PostReminder{
+		TargetTime: targetTime,
+		PostId:     postID,
+		UserId:     userID,
+	})
+	if err != nil {
+		return err
+	}
+	return nil
```
agarciamontoro, line 1342: This should have been caught by the linter, right?
```suggestion
	return err
```

agarciamontoro (changes requested):
Looks good, thank you! Just a change that should have been caught by the linter, if I'm not mistaken.

loadtest/user/userentity/actions.go
agnivade: Not really. I was just being consistent with the rest of the codebase here.

Makefile
agnivade: We are moving away from downloading a binary, to installing via the Go toolchain. If you run this command for the 2nd time, it doesn't do anything.

loadtest/control/simulcontroller/controller.go
agnivade: Done: https://mattermost.atlassian.net/browse/MM-46110

loadtest/user/userentity/actions.go
agarciamontoro: Ah, maybe we had this configured in Playbooks, then, I've always seen this flagged by the linter. Let's go with this pattern if it's used everywhere, then.

Makefile
agarciamontoro: Ah ok ok, that makes sense. Thanks!

agarciamontoro (approved):
:tada:

ashishbhate (approved):

loadtest/control/simulcontroller/actions.go
```diff
@@ -1813,3 +1813,24 @@ func (c *SimulController) getInsights(u user.User) control.UserActionResponse {
...
+func (c *SimulController) createPostReminder(u user.User) control.UserActionResponse {
+	ch, err := u.Store().CurrentChannel()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	post, err := u.Store().RandomPostForChannel(ch.Id)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
```
ashishbhate, line 1826: Non-blocker: 
Should this be a user action error if no posts are found in the channel store? 
IIUC, user action error counts towards to the error rate that decides when a load test should stop, so maybe this should be no-op instead of returning an error.

agnivade: Yeah good point, I can check that the channel does have posts. Thanks.

loadtest/user/userentity/actions.go
agnivade: I just remembered why I did this. Mostly this is to safe-guard against new code additions. If you want to add further logic down below, you would need to modify the old line as well. Not a big deal always, but can be missed sometimes.

agarciamontoro: Yeah, I see the potential benefit when adding new code, although it's not that important from my point of view. Either way, I'm 0/5 on this, not a big deal.

PR #502, opened 2022-08-04
Tweak release process

agarciamontoro:
#### Summary
This splits the release process in two steps, so that the updated config values are actually in the release itself. I documented the process there, so I won't explain more to check whether the docs are good :)

Let me know your thoughts.

#### Ticket Link
--

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -95,33 +95,48 @@ test:
 MATCH=v.+\/mattermost-load-test-ng-v.+-linux-amd64.tar.gz
 REPLACE=$(NEXT_VER)\/mattermost-load-test-ng-$(NEXT_VER)-linux-amd64.tar.gz
 TAG_EXISTS=$(shell git rev-parse $(NEXT_VER) >/dev/null 2>&1; echo $$?)
+BRANCH_NAME=bump-$(NEXT_VER)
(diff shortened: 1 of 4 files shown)


agarciamontoro:
Not assigning Claudio, since he's on PTO until next Wednesday, and I'd like to have this ready before I do the next release, which is a pre-requisite to run the 7.2-RC1 load-tests. All this is to say that if you could prioritize reviewing this, I'd really appreciate it :pray:

Makefile
```diff
@@ -95,33 +95,48 @@ test:
...
+PR_URL=https://github.com/mattermost/mattermost-load-test-ng/compare/master...$(BRANCH_NAME)?quick_pull=1&labels=2:+Dev+Review
+
+prepare-release:
+ifndef NEXT_VER
+	@echo "Error: NEXT_VER must be defined"
+else
+ifeq ($(TAG_EXISTS), 0)
+	@echo "Error: tag ${NEXT_VER} already exists"
+else
+	git checkout -b $(BRANCH_NAME) origin/master
```
agnivade, line 108: Maybe do a check if $(BRANCH_NAME) already exists?

agnivade (approved):

Makefile
agarciamontoro: Good catch, thanks! Done in https://github.com/mattermost/mattermost-load-test-ng/pull/502/commits/7ec8af8a12209922a34b55999f5c82a07ce9fb4f

isacikgoz (approved):

PR #510, opened 2022-08-30
check for gnuplot for comparison command

darkLord19:
Fixes https://github.com/mattermost/mattermost-server/issues/20884

### DIFF
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -10,6 +10,7 @@ import (
 	"io"
 	"io/ioutil"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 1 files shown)


agarciamontoro:
Thank you for the PR, @darkLord19. Sorry for the delay, I was on vacation and came back today. I'll take a look as soon as I can :)

cmd/ltctl/comparison.go
```diff
@@ -102,6 +103,10 @@ func writeReports(results []comparison.Result, outPath string) error {
 }
 
 func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
+	if _, err := exec.LookPath("gnuplot"); err != nil {
+		return fmt.Errorf("gnuplot is not installed. The comparison command requires it to be installed: %w", err)
+	}
```
agarciamontoro, line 108: The check looks good, but we should only do it if the value of `Output.GenerateGraphs` in the comparison config is true. So maybe it makes more sense to also move the check to `comparison.New`, after the default validation.

agarciamontoro (changes requested):

agarciamontoro:
Hey, @darkLord19, that was our bot marking PRs without activity for some days. Can I help further with this?

agarciamontoro (approved):
Looks great, @darkLord19, thank you! :)

PR #512, opened 2022-10-18
MM-47265: Set default_text_search_config to english

agarciamontoro:
#### Summary
There are multiple places where we assume that the default language in the database is `english` (e.g., the GIN indexes are hardcoded to `english`), but the `default_text_search_config` parameter in Aurora defaults to `simple`. It is not possible to change that in the config level in AWS, so these changes make sure that the parameter has the expected value by altering the DB itself in two places:
- When creating any new Terraform deployment
- When running a comparison, which needs to reset the param after dropping and recreating the DB.

This has been tested with:
- `go run ./cmd/ltctl deployment create`
- `go run ./cmd/ltctl comparison run`

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47265

### DIFF
--- a/comparison/loadtest.go
+++ b/comparison/loadtest.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/coordinator"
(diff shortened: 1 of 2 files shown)


isacikgoz (approved):
LGTM

agnivade:
> It is not possible to change that in the config level in AWS, 

This was the PR for that: https://github.com/mattermost/mattermost-load-test-ng/pull/462

@streamer45 - I forget what was the reason we didn't merge it. Can we merge it now?

streamer45:
> > It is not possible to change that in the config level in AWS,
> 
> This was the PR for that: #462
> 
> @streamer45 - I forget what was the reason we didn't merge it. Can we merge it now?

The reason is in https://github.com/mattermost/mattermost-load-test-ng/pull/462#issuecomment-1192808949

agarciamontoro:
The thing is that `default_text_search_config` is not available in DB parameters group, so even if we merge Claudio's PR (which we should at some point, it's just that other stuff has had more priority), we'll need the changes from this one too.
image

agnivade:
Oh never mind then

agnivade (approved):

PR #513, opened 2022-10-20
MM-47262: Bump terraform version 0.15

agarciamontoro:
#### Summary
This bump to v0.15 is the first step to get to the Terraform version we want: v1.3.

Please, review by commit, the commit messages explain the two main changes that ship here.

There's still a warning about the [S3 bucket ACL resources](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket_acl), but this was also shown with v0.14, so I'm fixing that in a separate PR.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47262


isacikgoz (approved):
LGTM, is there anything we need to change while running the `ltctl` command?

agarciamontoro:
> is there anything we need to change while running the ltctl command?

No :)

noxer (approved):
LGTM.

Nitpicks:
- We should start replacing the deprecated `ioutil` functions.
- It's not immediately relevant here but handling paths for file systems should be done with the `filepath` package instead of `path`.

agarciamontoro:
Thank you, @noxer, great catches! I just created help wanted tickets for both: https://mattermost.atlassian.net/browse/MM-47828 and https://mattermost.atlassian.net/browse/MM-47829 suggestions.
With regards to this PR, I'm still debugging an issue I found while testing this, but that I think it's unrelated to these changes. I'll pause merging until I find out what's happening there.

agarciamontoro:
I have tested this several times with:
1. `go run ./cmd/ltctl deployment {create,destroy}`
2. `go run ./cmd/ltctl comparison {run,destroy}`

In the `deployment` case everything is ok. The issue is in the `comparison` case: the first time I run `comparison run`, I get errors from all the agents, except for the 0th one, in the `remote-exec` block. However, running the same command a second time does start the comparison correctly and everything is nominal. I ran the same performance comparison I did with `7.4.0-rc1` and the results are similar.

The thing is that I can reproduce this exact same issue with load-test v1.6.0 and Terraform v0.14.11. The only difference is that the output from Terraform v0.15.5 is more verbose and more precise about what happened, but the problem is the same. Thus, the issue does not come from this PR and is something that maybe has always been there, and that I may have overlooked given the little feedback v0.14 returned.

In any case, it makes sense to merge this PR, since it's verified that the v7.4.0-rc1 performance comparison done with this Terraform v0.15 in this branch returned the same results.

PR #514, opened 2022-10-21
MM-47828: Replace deprecated ioutil package

wralith:
This pr replaces deprecated `ioutil` package usages with corresponding functions in `os` and `io` packages.
Also, there are 3 generated bindata.go files still uses `ioutil`, I am not sure about what to with those.

Fixes https://github.com/mattermost/mattermost-server/issues/21479

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"net"
 	"net/http"
 	"os"
@@ -290,7 +289,7 @@ func getServerVersion(serverURL string) (string, error) {
 		return version, fmt.Errorf("failed to get server version: %w", err)
 	}
 	defer resp.Body.Close()
-	io.Copy(ioutil.Discard, resp.Body)
+	io.Copy(io.Discard, resp.Body)
 
 	header := resp.Header["X-Version-Id"]
 	if len(header) > 0 {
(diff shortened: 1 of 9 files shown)


agarciamontoro:
Wow, that was quick! Thanks, @wralith :)

> Also, there are 3 generated bindata.go files still uses `ioutil`, I am not sure about what to with those.

Those files are auto-generated, so it's actually `go-bindata` who would need to be updated. That project is no longer maintained now, so we'll have to look for a replacement at some point. So no action needed from you there, thanks for asking!

I'll take a deeper look as soon as possible, but on first sight all looks good :)

isacikgoz (approved):
Thanks for the contribution @wralith 🎉 This looks good, for embedding files via generated files we can actually use `go:embed`. @agarciamontoro do you think it would be another cool ticket for the community?

noxer (approved):
Amazing (and fast!) work @wralith. LGTM.

agarciamontoro (approved):
Flawless! Thanks, @wralith :)

agarciamontoro:
> for embedding files via generated files we can actually use go:embed. @agarciamontoro do you think it would be another cool ticket for the community?

@isacikgoz, that's a good call. I'd need to look a bit more into that since I'm not familiar with `go:embed`, but if you are and you're open to create the ticket, please go ahead, I think it'd be a good help-wanted one.

PR #517, opened 2022-10-27
MM-47965: Revamp working directory management

agarciamontoro:
#### Summary

This set of changes revamps how the Terraform working directory is managed. Now, instead of creating a different temporary directory every time we run a command, we base all Terraform commands on the `deployment/terraform/assets` directory, where all Terraform source files live.

The motivation behind this is:
- The way things were designed, a temporary directory was only guaranteed to live during a single command run (a `go run ./cmd/ltctl {deployment,comparison} whatever`)
- We could pass that temporary directory to `-chdir` when running `terraform output`. If the temporary directory is there, this works. This is the case, for example, when the code calls `terraform output` as the last command during a `deployment create`.
- But, for example, `go run ./cmd/ltctl deployment info` calls `terraform output` independently, so there's no temporary directory anymore that we can pass to `-chdir` in this case. An alternative would be to run `terraform init` before the `terraform output`, but if we already have the `.terraform` directory there, then that's overkill.

So I decided to just use a directory in the repo for storing the .terraform directory (which before was stored in the root of the repo---see first commit) as well as the .tfstate files.

You can review this PR by commit, the commit messages should be self-explanatory.

The usage of the tool should not have changed at all, the commands are still the same.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47965


isacikgoz (approved):
Looks good, and the commit messages were nice touch to understand. Clear and neat 💯

agnivade:
Going to review this by today EOD.

agnivade (commented):
Yes, I was stuck with this directory management stuff when I first attempted upgrade from 0.14 to 0.15.

Couple of comments:
1. We used the asset generation to let binaries be able to create deployments as well. Because all assets would be compiled into the binary. I think this change assumes the existence of deployment/terraform/assets directory to be present. It would be good to not be dependent on any source directories at all.

2. Comparison mode is still unaffected by this right? It just creates deployment0.tfstate and deployment1.tfstate inside the assets dir right?

agarciamontoro:
>     1. We used the asset generation to let binaries be able to create deployments as well. Because all assets would be compiled into the binary. I think this change assumes the existence of deployment/terraform/assets directory to be present. It would be good to not be dependent on any source directories at all.

Ah! That makes sense, and it's something I had not considered, thanks for bringing it up. Then I'll have to rethink the whole approach, since this indeed assumes the existence of that directory.

The obvious alternative is letting everything more or less like it is right now and run `terraform init` even for running `deployment info`. I don't like that solution, though, since running `init` takes a lot of time regenerating the `.terraform` directory (that is, downloading the one provider we use now, `hashicorp/aws`) that could have been saved if other commands already ran.

The problem that needs to be solved here is that we can no longer separate the working directory (where the source files live) and the directory where the `.terraform` directory is generated. That could be done with 0.14, but not anymore. So if we use a temporary directory that will be cleaned between commands, we need to regenerate the `.terraform` directory every time we run anything. We could, however, create a `.loadtest` directory right where the binary is run, instead of the temporary one, and use that as the working directory, so we copy all source files (everything under `deployment/terraform/assets`) there and use the global option `-chdir=./.loadtest` for every command, so that both the `.terraform` directory and the state files are stored there. We don't cleanup that directory, though, to keep the `.terraform` directory intact.

Thoughts?

>     2. Comparison mode is still unaffected by this right? It just creates deployment0.tfstate and deployment1.tfstate inside the assets dir right?

Yes, that stays the same, this is handled by `getStatePath`.

agnivade:
Having a separate `.loadtest` directory sounds good. :+1:

agarciamontoro:
https://github.com/mattermost/mattermost-load-test-ng/pull/517/commits/76712edfe659e5fa89d1ada5c797ba701b11ae38 changes the approach to what we discussed: it relies on a fixed directory, but it lets the user configures where it is. It defaults to `/var/lib/mattermost-load-test-ng` (see the commit message to the why), but https://github.com/mattermost/mattermost-load-test-ng/pull/517/commits/171a791d67856be9b39b573901875accf1d1e3d6 documents what that directory is and how to change it, since creating the default needs root permissions. I tried to use `~/.loadtest` as the default, but it does not seem to like the tilde, so `/var/lib` it is.

Thoughts?

agnivade (approved):
Nice!

PR #523, opened 2022-11-15
[GH-21586] - Removed deprecated windows build function in Makefile

mattlam88:
#### Summary
The PR includes the removal of the deprecated build-windows command in the Makefile. Additionally, there is an update to the documentation that windows is no longer supported for make commands.

Please let me know if there is any additional documentation spots that may be a good candidate to add this change, as this is my first OSS PR!

#### Ticket Link
Fixes the following Github issue: https://github.com/mattermost/mattermost-server/issues/21586

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,6 @@ build-osx: ## Build the binary (only for OSX on AMD64).
 	env GOOS=darwin GOARCH=amd64 $(GO) build -o $(AGENT) $(AGENT_ARGS)
 	env GOOS=darwin GOARCH=amd64 $(GO) build -o $(API_SERVER) $(API_SERVER_ARGS)
 
-build-windows: ## Build the binary (only for Windows on AMD64).
-	@echo Build Windows amd64
-	env GOOS=windows GOARCH=amd64 $(GO) build -o $(AGENT) $(AGENT_ARGS)
-	env GOOS=windows GOARCH=amd64 $(GO) build -o $(API_SERVER) $(API_SERVER_ARGS)
-
 assets: ## Generate the assets. Install go-bindata if needed.
 	go install github.com/kevinburke/go-bindata/go-bindata@v3.23.0
 	go generate ./...
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -1,6 +1,6 @@
 ## Developer's workflow
 ### Notes
-The load-test tool does not support Windows
+The load-test tool does not support Windows.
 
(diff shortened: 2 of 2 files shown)


docs/developing.md
```diff
@@ -28,6 +28,8 @@ make package
 
 We use a combination of `Makefile` and [Goreleser](https://goreleaser.com/) to automate our release process.
 
+NOTE: Makefile no longer supports building the binary file using Windows on AMD64
+
```
agarciamontoro, line 32: This is true, but it's not the only problem. The binary cannot be built for Windows at all now (we use some unix-specific stuff), not only because it's not in Makefile. We can reword this to be more generic. Something like "The load-test tool does not support Windows". And maybe putting somewhere else, not under the `Releasing` section. Even at the beginning of this file. Thoughts, @mattlam88?

Makefile
```diff
@@ -35,11 +35,6 @@ build-osx: ## Build the binary (only for OSX on AMD64).
 	env GOOS=darwin GOARCH=amd64 $(GO) build -o $(AGENT) $(AGENT_ARGS)
 	env GOOS=darwin GOARCH=amd64 $(GO) build -o $(API_SERVER) $(API_SERVER_ARGS)
 
-build-windows: ## Build the binary (only for Windows on AMD64).
-	@echo Build Windows amd64
-	env GOOS=windows GOARCH=amd64 $(GO) build -o $(AGENT) $(AGENT_ARGS)
-	env GOOS=windows GOARCH=amd64 $(GO) build -o $(API_SERVER) $(API_SERVER_ARGS)
-
```
agarciamontoro, line 42: :tada:

agarciamontoro (changes requested):
Thank you for your work here, @mattlam88 :tada: I'm glad your first contribution to OSS is for this repo! I left a comment below, thanks!

docs/developing.md
mattlam88: That makes sense I'll update the PR to include that at the top

docs/developing.md
```diff
@@ -1,4 +1,6 @@
 ## Developer's workflow
+### Notes
+The load-test tool does not support Windows
```
agarciamontoro, line 3: Just a super small nit:
```suggestion
The load-test tool does not support Windows.
```

agarciamontoro (approved):
Cool! Thanks, @mattlam88 :tada:

Issue #525, opened 2022-12-02
Simulcontroller fail if username length less than 4


Gitnube:
On call searchGroupChannels an error occurs if the selected username length less than 4:
https://github.com/mattermost/mattermost-load-test-ng/blob/382e0d2549e9e1c5fc6c1eb9219f716d7f0e527b/loadtest/control/simulcontroller/actions.go#L1243

This case is fixed on searchChannels:
https://github.com/mattermost/mattermost-load-test-ng/blob/382e0d2549e9e1c5fc6c1eb9219f716d7f0e527b/loadtest/control/simulcontroller/actions.go#L1101-L1108

I propose on searchGroupChannels do the same

agnivade:
Fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/526

PR #526, opened 2022-12-02
fixed emulate typing username on searchGroupChannels for simulcontroller

Gitnube:
if length username less than 4 then an error occurs "slice bounds out of range"

#### Summary
Fixed length of string on emulate typing username on searchGroupChannels for simulcontroller

#### Ticket Link
https://github.com/mattermost/mattermost-load-test-ng/issues/525


agnivade (approved):
Thanks!

PR #529, opened 2022-12-12
Adding `TerraformStateDir` to deployer config

coltoneshaw:
#### Summary
You get the below error when this isn't added

```bash
ubuntu@ip-172-31-86-142:~/mattermost-load-test-ng$ go run ./cmd/ltctl deployment create
Error: failed to validate config: TerraformStateDir is empty
exit status 1
```

I manually added it to my `deployer.json` and it fixed the error

### DIFF
--- a/config/deployer.sample.json
+++ b/config/deployer.sample.json
@@ -21,6 +21,7 @@
   "EnableAgentFullLogs": true,
   "ProxyInstanceType": "c5.xlarge",
   "SSHPublicKey": "~/.ssh/id_rsa.pub",
+  "TerraformStateDir" : "/var/lib/mattermost-load-test-ng",
   "TerraformDBSettings": {
     "InstanceCount": 1,
     "InstanceEngine": "aurora-postgresql",


coltoneshaw:
It actually doesn't love this default config, for what it's worth. I changed mine to `~/terraform` and it works great.

agnivade (approved):

agarciamontoro:
>  I changed mine to ~/terraform and it works great.

Did you literally write `~/terraform`? That was my first option, but Terraform didn't like the tilde in there, so I had to resort to something most of users will have in their system.

agarciamontoro (approved):
In any case, I'm ok with this change. Users will now get a different error, as they won't have permissions over that directory by default. That's more meaningful than `"TerraformStateDir" is empty`, though.

coltoneshaw:
@agarciamontoro - You're right, I actually changed it to the absolute path of `/home/ubuntu/`. Looks like it didn't take my `~` when i initialized it.

agarciamontoro:
>  Looks like it didn't take my ~ when i initialized it.

We could do some form of pre-processing for this value and handle `~` implicitly. On the other hand, having to think where you're going to store the state is a good thing for the user.

PR #530, opened 2022-12-13
Add `S3BucketDumpURI` to `deployer.sample.json`

coltoneshaw:
## Summary

Adding a default value for `S3BucketDumpURI` to the deployer config. This value doesn't appear required, but this should remove confusion when going to set it.


agarciamontoro (approved):
Shipping it!

PR #532, opened 2022-12-13
Clarified the max active users

coltoneshaw:
Added some clarity to the max active users when attempting to loadtest over 52k users.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -34,8 +34,31 @@ Also the `Max Users Per Team` setting in Mattermost System Console should be eno
 
 ### Agent failing with `MaxActiveUsers is not compatible with max Rlimit value` error
 
-This means the maximum number of file descriptors is lower than what the agent needs to operate.  
+This means the maximum number of file descriptors is lower than what the agent needs to operate. 
+
 The following command can be run to raise the limit to the suggested value:
 ```sh
 ulimit -n VALUE
 ```
+
+#### Note
+For terraform deployments, this value is hard coded as 65535 in the `systemd` file for the loadtest api. If you are attempting to test over ~52k users you will need to manually update this file with the instructions below.
+
+1. ssh into your loadtest agents. You can see the agents available by running `go run ./cmd/ltctl ssh`.
+2. Modify the `/lib/systemd/system/ltapi.service` file with the below change.
+
+```diff
++ LimitNOFILE=150000
+- LimitNOFILE=65535
+```
+
+3. Restart the related processes
+
+```bash
+sudo systemctl daemon-reload
+sudo systemctl restart ltapi
+```
+
+You will have to run this for every loadtest agent you have. These will be appended by `agent-` when you run the `ltctl ssh` command above. 
+
+
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -42,16 +42,10 @@ ulimit -n VALUE
 ```
 
 #### Note
-For terraform deployments, this value is hard coded as 65535 in the `systemd` file for the loadtest api. If you are attempting to test over ~52k users you will need to manually update this file with the instructions below.
+For terraform deployments, this value is hard coded as 65535 in the `systemd` file for the loadtest api. If you need to change the value, you'll have to change the `LimitNOFILE` value in `/lib/systemd/system/ltapi.service` file to a higher value. 
 
 1. ssh into your loadtest agents. You can see the agents available by running `go run ./cmd/ltctl ssh`.
-2. Modify the `/lib/systemd/system/ltapi.service` file with the below change.
-
-```diff
-+ LimitNOFILE=150000
-- LimitNOFILE=65535
-```
-
+2. Modify the `/lib/systemd/system/ltapi.service` file with the new value.
 3. Restart the related processes
 
 ```bash
@@ -60,5 +54,6 @@ sudo systemctl restart ltapi
 ```
 
 You will have to run this for every loadtest agent you have. These will be appended by `agent-` when you run the `ltctl ssh` command above. 
+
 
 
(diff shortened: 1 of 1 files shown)


docs/faq.md
```diff
@@ -34,8 +34,31 @@ Also the `Max Users Per Team` setting in Mattermost System Console should be eno
...
+#### Note
+For terraform deployments, this value is hard coded as 65535 in the `systemd` file for the loadtest api. If you are attempting to test over ~52k users you will need to manually update this file with the instructions below.
+
+1. ssh into your loadtest agents. You can see the agents available by running `go run ./cmd/ltctl ssh`.
+2. Modify the `/lib/systemd/system/ltapi.service` file with the below change.
+
+```diff
++ LimitNOFILE=150000
+- LimitNOFILE=65535
+```
```
agnivade, line 53: The instance size of the agent also plays a role here. I'd rather advise to bump up the number of agents horizontally rather than bump it up vertically.

What I'd suggest is to just mention something like:

```
For terraform deployments, this value is hard coded as 65535 in the `systemd` file for the loadtest api. If you need to change the value, you'll have to change the `LimitNOFILE` value in `/lib/systemd/system/ltapi.service` file.
```

coltoneshaw: You're saying that if I wanted to do a load test with 100k users. I simply need to add additional agent nodes to my deployer.json?

coltoneshaw: Let me know if that clarification works. I left the instructions but removed the specifics, just to clear up potential questions. If we still don't want them, i'll drop it to just the one line.

agnivade: > You're saying that if I wanted to do a load test with 100k users. I simply need to add additional agent nodes to my deployer.json?

That is correct. Only on the agent side though. You'll need to similarly scale up the server nodes and the DB nodes. Although you will eventually get bottlenecked on the intra cluster traffic.

agnivade (approved):

agnivade:
I'll request Claudio to take another look.

coltoneshaw:
@agnivade - Thanks! For what it's worth, I'm actively helping a customer loadtest 75k because that's their business requirements and this comes from the steps I had to give them to unblock the max users.

docs/faq.md
coltoneshaw: Interesting. I tried that and it was still yelling at me, but let me give it another go now that I understand this a bit more. Is there a limit to agent nodes on a load test where it stops working?

streamer45:
A couple of thoughts:

- While this limit could ideally be made configurable I am thinking given this is a load testing tool it would also be acceptable to simply bump the hard-coded value to cover the higher load cases. I don't think this should be a problem and it would simplify a lot since it seems like overly complex for a user to have to change a file in potentially dozens of instances and restart services when it could be easily automated for free.
- I agree with Agniva that agents are made to be scaled up horizontally as there's so much they can handle. I'd stick with the recommended values (2000 users per `c5.xlarge` instance I believe) but simply deploy more of them.

streamer45:
> Oh wow, I didn't know this recommendation. I'm definitely far exceeding that. Do we list this somewhere that I just overlooked? How do I limit the number of users per agent?

There's a note in https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/setups.md#notes but can probably be surfaced a little better. Essentially the default is actually a maximum value that was proven to work well for most load-tests using default configs.

>  I assume the coordinator controls the number of users per agent with `MaxActiveUsers` and the config controls it locally?

Yes, ultimately the `UsersConfiguration.MaxActiveUsers` in `config.json` is what controls how many users are simulated per instance.

streamer45:
> So, thinking hypothetically. If i wanted to test 50k users I need 25 agent servers running a `c5.xlarge` ?

That's correct. Of course, depending on your setup (e.g. how many teams/channels you are testing on) at some point this won't scale linearly and you'll have to bump the instance class as well since simulated users will likely have to do more (e.g. receiving more events due to the increasing traffic).

coltoneshaw:
@streamer45 - Okay, this is awesome. I wonder if that's why my coordinator keeps dying too - https://community.mattermost.com/core/pl/7bbhe8egotftzghzomodbwayzh. If you have an idea where this could be surfaced better, I can submit a docs PR.

agarciamontoro:
What does everyone think about keeping these docs just in case, and bumping the hardcoded value as well, as suggested by Claudio?

coltoneshaw:
@agarciamontoro - I'd agree keeping them would be useful, since it's not extremely clear how to increase it if you do need to. I wonder how hard it would be to make the base value something configurable in the deployer.json. That would be the most ideal route from my perspective.

agarciamontoro:
> I wonder how hard it would be to make the base value something configurable in the deployer.json.

It shouldn't be too hard. These strings seem to use [Go's templating package](https://pkg.go.dev/text/template) for controlling the content, so you should be able to add a new config parameter and inject it when rendering the template.

If you're interested in trying, please do so. Otherwise, we can open a help wanted ticket.

agarciamontoro:
Friendly ping to review this, @streamer45. I guess we can merge as is and decide next steps independently of this PR.

streamer45 (commented):
I still feel this shouldn't be an issue users should face on an automated deployment as it seems like the manual fix is just adding unnecessary burden on them so my proposal stands of just bumping the value to something that covers even larger load-tests. 

With that said, I am going to unblock this and will defer any final decision to @agarciamontoro

agarciamontoro:
Thanks, @streamer45. Final decision is to merge these changes as well as bump the hardcoded value: https://github.com/mattermost/mattermost-load-test-ng/pull/543

I took the liberty to remove the specific number from the new docs in case we change the hardcoded value again, it's easy that those things get out of sync quickly :)

agarciamontoro (approved):
Thank you, @coltoneshaw, your help refining all these docs is invaluable <3

PR #533, opened 2022-12-14
Added `ServerVersion` to the usercontroller config

coltoneshaw:
Reading the docs [here](https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/loadtest_config.md#serverversion) and noticed the serverVersion setting was not included. This PR just includes it as a blank string, since that appears to be the default.

### DIFF
--- a/config/config.sample.json
+++ b/config/config.sample.json
@@ -28,7 +28,8 @@
         "Rate": 30.0,
         "Percentage": 0.3
       }
-    ]
+    ],
+    "ServerVersion" : ""
   },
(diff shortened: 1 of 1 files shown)


agarciamontoro (approved):
Huh, not sure when we have used this, but it doesn't hurt to have it in the sample. Thanks :)

isacikgoz (approved):
I think this was being used to test specific api calls if the server version supports. I checked and this is already being set by the api call. Anyway, wouldn't hurt to have it here.

PR #535, opened 2022-12-15
Added HasDB check to default text search config

coltoneshaw:
#### Summary
When attempting to use an existing database and letting the load tester build Mattermost, it will still attempt to configure some search settings. This just adds a check for `HasDB()` before running those commands. Did a quick look to see if any other things would benefit from this and couldn't find any.

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -149,6 +149,22 @@ func (t *Terraform) Create(initData bool) error {
 		}
 	}
 
+	if t.output.HasDB() {
+		// Aurora sets the default_text_search_config parameter to 'simple';
+		// a more sane default for a generic deployment is 'english'
+		if t.config.TerraformDBSettings.InstanceEngine == "aurora-postgresql" {
+			if err := t.setDefaultTextSearchConfig(extAgent, "pg_catalog.english"); err != nil {
+				return fmt.Errorf("could not modify default_search_text_config: %w", err)
+			}
+		}
+
+		if initData {
+			if err := t.createAdminUser(extAgent); err != nil {
+				return fmt.Errorf("could not create admin user: %w", err)
+			}
+		}
+	}
+
 	if t.output.HasAppServers() {
 		url := t.output.Instances[0].PublicDNS + ":8065"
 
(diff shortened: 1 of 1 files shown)


deployment/terraform/create.go
```diff
@@ -168,7 +168,7 @@ func (t *Terraform) Create(initData bool) error {
 
 		// Aurora sets the default_text_search_config parameter to 'simple';
 		// a more sane default for a generic deployment is 'english'
-		if t.config.TerraformDBSettings.InstanceEngine == "aurora-postgresql" {
+		if t.config.TerraformDBSettings.InstanceEngine == "aurora-postgresql" && t.output.HasDB() {
```
agnivade, line 171: I don't know why this if condition is inside `if t.output.HasAppServers() {`. It should be a standalone if condition, but I'll defer to Alejandro here.

agarciamontoro: Agree, we could refactor this whole thing as something like:
```go
if t.output.HasAppServers{
	// ...
	if err := pingServer("http://" + url); err != nil {
		return fmt.Errorf("error whiling pinging server: %w", err)
	}
}

if t.output.HasDB() {
	if t.config.TerraformDBSettings.InstanceEngine == "aurora-postgresql" {
		// ...
	}
	if initData {
		// ...
	}
}
```

agarciamontoro (changes requested):
Do you want to try the refactor discussed above, @coltoneshaw?

deployment/terraform/create.go
```diff
@@ -149,6 +149,22 @@ func (t *Terraform) Create(initData bool) error {
...
+	if t.output.HasDB() {
+		// Aurora sets the default_text_search_config parameter to 'simple';
+		// a more sane default for a generic deployment is 'english'
+		if t.config.TerraformDBSettings.InstanceEngine == "aurora-postgresql" {
+			if err := t.setDefaultTextSearchConfig(extAgent, "pg_catalog.english"); err != nil {
+				return fmt.Errorf("could not modify default_search_text_config: %w", err)
+			}
+		}
+
+		if initData {
```
coltoneshaw, line 161: @agarciamontoro - Would this be restricted based on if you have a DB or not? The way I'm reading this is that it's just if you're having it initialize data, which wouldn't be 100% tied to whether it's making a new DB or not, would it?

agnivade: Yes, there is no need to move this condition. This should remain under `output.HasAppServers()`

agnivade (changes requested):

deployment/terraform/create.go
agarciamontoro: It could be argued that the DB should be modified only in the case it's deployed by Terraform, and assume that the user wants full control (so no changes in the config nor in the initial data) if the DB is already provided by them.
Thoughts, @agnivade?

agnivade: Oh yeah good point. Approved.

agnivade (approved):

agarciamontoro (approved):
Shipping it

PR #537, opened 2023-01-10
MM-49581: Add path, method and code to HTTPRequestTimes

agarciamontoro:
#### Summary
The `loadtest_http_request_time_{sum,count,bucket}` metric is right now a `prometheus.Histogram`, so it doesn't have labels. This PR converts it to a `prometheus.HistogramVec`, and adds three labels to it:
1. Request path
2. Request method
3. Request status code

The observation of the time itself is unchanged, this PR only adds those dimensions.

This is always useful, but in this metric may be even more, since one of the main queries used by the coordinator to decrease users and eventually stop the tests uses this specific metric. Having some more fine-grained information there will help us analyze the results of the load-tests in more detail.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49581

### DIFF
--- a/loadtest/user/userentity/metrics.go
+++ b/loadtest/user/userentity/metrics.go
@@ -31,9 +31,13 @@ func (ue *UserEntity) incHTTPErrors(path, method string, status int) {
 	}
 }
 
-func (ue *UserEntity) observeHTTPRequestTimes(elapsed float64) {
+func (ue *UserEntity) observeHTTPRequestTimes(path, method string, status int, elapsed float64) {
 	if ue.metrics != nil {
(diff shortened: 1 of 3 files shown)


agarciamontoro:
It was a bit surprising to see that this metric didn't contain this information. Any idea why we didn't have it in the first place? Was it simply because we had forgotten?

agnivade (approved):

agnivade:
> It was a bit surprising to see that this metric didn't contain this information. Any idea why we didn't have it in the first place? Was it simply because we had forgotten?

Probably at that time, we didn't bother with granularity.

isacikgoz (approved):

PR #539, opened 2023-01-16
MM-49682: Move admin user creation right after server setup

agarciamontoro:
#### Summary
PR #535 inverted the order of server creation and admin user creation, so that the deployer was trying to create the user without having set up the server, and hence failed.

As an aside: we should start thinking about setting up some E2E tests; something as simple as "run deployment create and make sure that the server is up" would have prevented this. The thing is: I'm not sure what's the best way to do this without burning money on AWS every time we run CI.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-49682

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -157,12 +157,6 @@ func (t *Terraform) Create(initData bool) error {
 				return fmt.Errorf("could not modify default_search_text_config: %w", err)
 			}
 		}
-
-		if initData {
-			if err := t.createAdminUser(extAgent); err != nil {
-				return fmt.Errorf("could not create admin user: %w", err)
-			}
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
PR looks good to me. For the e2e testing, maybe we can run this manually with adding a tag or running a slash command to trigger mattermod?

agarciamontoro:
> For the e2e testing, maybe we can run this manually with adding a tag or running a slash command to trigger mattermod?

Yeah, and maybe restricting execution permissions that to our team for the time being. In any case, it gives me the chills to have something running (semi)automatically on AWS :joy: Maybe something for Q1

noxer (approved):

PR #541, opened 2023-01-17
MM-47263: Empty S3 bucket with AWS CLI on destroy

agarciamontoro:
#### Summary
This bug is a funny one, since it's not reproducible at all. I've seen it in the wild at some point after bumping to Terraform v1 (much less than before, but this may be just luck), although for me it usually was fixed with a second run of `deployment destroy`. However, it's been reported at least once that this happened even after three executions of `deployment destroy`, which always timed out at the 30-minute mark with only the S3 bucket left to destroy. The only solution that has consistently worked for each and every instance of this bug was to manually empty the S3 bucket, even while the `deployment destroy` command was executing: that is, running `aws --profile mm-loadtest s3 rm s3://bucket.id` concurrently with the destroy.

So that's what this PR does: it adds some defense code to run the `aws` command concurrently with the destroy. The third commit also adds a check to validate as soon as possible that the `aws` command is available in the system, so we can fail early if it's not there.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-47263

### DIFF
--- a/deployment/terraform/engine.go
+++ b/deployment/terraform/engine.go
@@ -118,3 +118,14 @@ func checkTerraformVersion() error {
 
 	return nil
 }
+
+// checkAWSCLI checks that the aws command is available in the system, and that
+// the profile that will be used is correctly configured
+func checkAWSCLI(profile string) error {
+	cmd := exec.Command("aws", "configure", "list", "--profile", profile)
+	err := cmd.Run()
+	if err != nil {
+		return fmt.Errorf("the AWS CLI is either not installed or the configured profile %q is not stored in the credentials; error: %w", profile, err)
+	}
+	return nil
+}
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -553,6 +553,10 @@ func (t *Terraform) preFlightCheck() error {
 		return fmt.Errorf("failed when checking terraform version: %w", err)
 	}
 
+	if err := checkAWSCLI(t.Config().AWSProfile); err != nil {
+		return fmt.Errorf("failed when checking AWS CLI: %w", err)
+	}
+
 	if !t.initialized {
 		if err := t.init(); err != nil {
 			return err
--- a/deployment/terraform/destroy.go
+++ b/deployment/terraform/destroy.go
@@ -3,11 +3,38 @@
 
 package terraform
 
+import (
+	"context"
+	"fmt"
+	"os/exec"
+	"strings"
+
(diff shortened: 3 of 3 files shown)


agnivade (commented):
Is the AWS CLI packaged with Terraform installation? If not, is there a possibility for us to use the AWS SDK instead, to remove the objects programmatically?

I'd try to avoid forcing users to install yet another CLI.

agarciamontoro:
@agnivade: It's not packaged with Terraform, no. I can check the possibility of using the SDK, the use-case we need it for [seems easy to implement](https://github.com/awsdocs/aws-doc-sdk-examples/blob/main/go/example_code/s3/s3_delete_objects.go). And then we can get rid of the other place where we use the AWS CLI. In hindsight, I don't know why I didn't consider this, thanks for bringing it up.

agnivade:
Oh is the AWS CLI already required to use the load test tool?

agarciamontoro:
> Oh is the AWS CLI already required to use the load test tool?

Yes, but it was [a recent addition](https://github.com/mattermost/mattermost-load-test-ng/pull/524). I'll check here and if it works, I'll refactor that one.

agnivade:
Ah I see. Let's push back the refactor as an HW task then. I want to move on to other things.

agnivade (approved):

agarciamontoro:
Sounds good, thanks for double-checking :)

agarciamontoro:
And another ping, @noxer :nerd_face:

deployment/terraform/engine.go
```diff
@@ -118,3 +118,14 @@ func checkTerraformVersion() error {
 
 	return nil
 }
+
+// checkAWSCLI checks that the aws command is available in the system, and that
+// the profile that will be used is correctly configured
+func checkAWSCLI(profile string) error {
+	cmd := exec.Command("aws", "configure", "list", "--profile", profile)
+	err := cmd.Run()
```
noxer, line 126: Just a nitpick: You can write those two as one line, since you're not using `cmd` anywhere else.

noxer (approved):
LGTM, just a very nitpicky nitpick.

deployment/terraform/engine.go
agarciamontoro: That's true, but I think this way is a bit more readable. I'll take the liberty to keep it since it's a nitpick :)

agarciamontoro:
> I'll open the HW ticket as soon as we merge this.

https://mattermost.atlassian.net/browse/MM-49972

PR #522, opened 2022-11-10
Load test post priority feature

koox00:
#### Summary

This PR is created to be able to load test Mattermost with PostPrioirty feature flag enabled.

Adds a config option `PercentUrgentPosts` for use in both `gencontroller` and `simulcontroller` which would be the percentage of posts created that will have the "urgent" priority.
Also enables the PostPriority feature flag, and ServiceSetting.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-46410

PR #542, opened 2023-01-20
[MM-47751] MPA: Send Persistent Notifications

vish9812:
#### Summary

Load test priority posts, ack posts and persistent notifications posts.

#### Ticket Link

JIRA - https://mattermost.atlassian.net/browse/MM-49930
Server PR - https://github.com/mattermost/mattermost-server/pull/21619


go.mod
```diff
@@ -96,19 +107,35 @@ require (
...
+	modernc.org/sqlite v1.20.1 // indirect
+	modernc.org/strutil v1.1.3 // indirect
+	modernc.org/token v1.1.0 // indirect
 )
 
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190515063710-7b18d6600f6b
 
 replace github.com/codegangsta/cli v1.22.1 => github.com/urfave/cli v1.22.1
+
+// replace github.com/mattermost/mattermost-server/v6 => ../mattermost-server
```
agarciamontoro, line 141: Leftover from debugging?

loadtest/control/simulcontroller/controller.go
```diff
@@ -234,6 +234,18 @@ func (c *SimulController) Run() {
...
+			frequency: 0.225,
+		},
+		{
+			run:       control.AckToPost,
+			frequency: 0.5,
+		},
+		// {
+		// 	run:       control.CreatePersistentNotificationPost,
+		// 	frequency: 5,
+		// },
```
agarciamontoro, line 248: Are these numbers based on real data? You can follow https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/coverage-frequency.md for coming up with the actual numbers.
Also: why is the last action commented out?

loadtest/store/memstore/store.go
```diff
@@ -1119,3 +1119,18 @@ func (s *MemStore) Thread(threadId string) (*model.ThreadResponse, error) {
 	}
 	return nil, ErrThreadNotFound
 }
+
+// PostsWithAckRequests returns IDs of the posts that asked for acknowledgment.
```
agarciamontoro, line 1123: Typo:
```suggestion
// PostsWithAckRequests returns IDs of the posts that asked for acknowledgement.
```

loadtest/store/memstore/store.go
```diff
@@ -1119,3 +1119,18 @@ func (s *MemStore) Thread(threadId string) (*model.ThreadResponse, error) {
...
+func (s *MemStore) PostsWithAckRequests() ([]string, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	var ids []string
+	for _, p := range s.posts {
+		if p.Metadata != nil && p.Metadata.Priority != nil && p.Metadata.Priority.RequestedAck != nil && *p.Metadata.Priority.RequestedAck {
```
agarciamontoro, line 1130: How I miss the optional chaining operator from Typescript with these never-ending conditionals :joy:

loadtest/store/memstore/store.go
```diff
@@ -1119,3 +1119,18 @@ func (s *MemStore) Thread(threadId string) (*model.ThreadResponse, error) {
 	}
 	return nil, ErrThreadNotFound
 }
+
+// PostsWithAckRequests returns IDs of the posts that asked for acknowledgment.
+func (s *MemStore) PostsWithAckRequests() ([]string, error) {
```
agarciamontoro, line 1124: Just thinking out loud, not asking for a change: Would it make sense to maintain a different list for these types of posts?

agarciamontoro (changes requested):
This looks great! I think we should change one bit of the implementation: instead of creating two new actions to create a post (one for ack and another for persistent notifications), I'd rather modify the `createPost` action and add the ack/persistent notification metadata to a percentage of the posts, similar to what @koox00 did for [the  PR](https://github.com/mattermost/mattermost-load-test-ng/pull/522). Thoughts?

The `AckToPost` action looks good, and the logic there with the `PostsWithAckRequests` is perfect. That one makes sense to be a separate action.

loadtest/control/simulcontroller/controller.go
vish9812: @koox00 Do we have real no. for Acks as well?

@agarciamontoro  The commented section is actually dependent on the bug I mentioned in the `Random` func. https://mattermost.atlassian.net/browse/MM-49975

loadtest/store/memstore/store.go
vish9812: So far there isn't any explicit need for it. But in future, if needed it surely can be created.

vish9812:
Will merge the separate actions for `Ack` and `PersistentNotification` posts into the existing `CreatePost`.

agarciamontoro:
> Edit: @agarciamontoro It doesn't seem a reliable way(https://github.com/mattermost/mattermost-load-test-ng/blob/master/loadtest/control/simulcontroller/actions.go#L764), if I want to keep these actions separate.
> As the `rand.Float64()` frequency will make it go through all the `if` conditions but here I want it to keep it separate from the other `ifs`.

What do you mean with keeping the actions separate? Do we need to have a post that is only urgent or only ack or only persistent? If not, we can have separate `if`s, as in:
```go
// 2% of the times post will have files attached.
if rand.Float64() < 0.02 {
	// attach files
}

// PercentUrgentPosts% of the times post will be urgent
if rand.Float64() < c.config.PercentUrgentPosts {
	// add urgency
}

// PercentAckPosts% of the times post will be ack-ready
if rand.Float64() < c.config.PercentAckPosts {
	// add ack
}

// PercentPersistentPosts% of the times post will have persistent notifications
if rand.Float64() < c.config.PercentPersistentPosts{
	// add persistency
}
```

Sorry if I'm misunderstanding something basic here.

loadtest/store/memstore/store.go
agarciamontoro: Ugh, you're right! The American spelling is so weird to me :joy:

loadtest/control/simulcontroller/controller.go
agarciamontoro: Sorry, I haven't had time to look into that bug yet. On the frequency for persistent notification, we can start with a very low number, then, and adjust when we have real data on it.

agarciamontoro: Btw, I'd love for you to also use the `MinServerVersion` field for the new actions. I know it's not anywhere else, but I've been asking for it for new actions, and I'll take care of old ones to set it.

loadtest/store/memstore/store.go
agarciamontoro: Sounds good, thanks!

vish9812:
For example in this case of ack and PNs:
```go
Metadata: &model.PostMetadata{
			Priority: &model.PostPriority{
				Priority:     model.NewString(model.PostPriorityUrgent),
				RequestedAck: model.NewBool(true),
			},
		}
```

```go
Metadata: &model.PostMetadata{
			Priority: &model.PostPriority{
				Priority:                model.NewString(model.PostPriorityUrgent),
				RequestedAck:            model.NewBool(false),
				PersistentNotifications: model.NewBool(true),
			},
		}
```

Now say, both of them are in their own `ifs` like 
// ack -> if < 2%
// PN -> if < 4%

Now, if the both conditions satisfy then only the later `if` PN post will be created with `ack` set as `false`. Even though, the 1st one was also true, it will be ignored as long as 2nd if is true.

So, the `ack` posts that were supposed to be created because there `if` condition was true, won't be created as long as 2nd `if` is also true.

Although, not sure if it's even a problem, just wanted to say, doesn't seem a reliable way of creating the correct % of posts, if multiple `ifs` can interfere with each other.

I'm not sure, if I've made it any clearer or more confusing 😕.

agarciamontoro:
@vish9812, thanks for the explanation. Can't we do something like this?

```go
metadata := &model.PostMetadata{
    Priority: &model.PostPriority{
        Priority:     model.NewString(model.PostPriorityUrgent),
    },
}

if ack {
    metadata.Priority.RequestedAck = true
}

if pn {
    metadata.PersistentNotification = true
}

// Use metadata to build the post
```

Or is it needed that when having PersistentNotification set to true, RequestedAck must be set to false?

vish9812:
> Or is it needed, for example, that when having PersistentNotification set to true, RequestedAck must be set to false?

@agarciamontoro yes, that's exactly what I need here.

agarciamontoro (changes requested):
Thanks for the hard work here @vish9812, and sorry for my delay on reviewing!

I left some comments below. Additionally, we need to do of the following:
1. Merge CreateAckPost and CreatePersistentNotificationPost into createPost.
2. Recompute the createPost frequency.

#### Alternative 1: merging the actions

For merging the actions, what we can do (sorry for not having understood the requirement previously) is compute relative percentages. Let me explain that: urgent posts, ack posts and persistent-notification posts are disjoint sets, but we only know their percentage with respect to all posts created. What we can do is compute what percentage each represents given that the others do not occur. In code, this would be something like:
```go
percUrgent = c.config.PercentUrgentPosts
conditionalPercAck = c.config.PercentAckPosts / (1 - c.config.PercentUrgentPosts)
conditionalPercPersistent = c.config.PercentPersistentPosts (1 - c.config.PercentUrgentPosts - c.PercentAckPosts)

if rand.Float64() < percUrgent {
	// add urgency
} else if rand.Float64() < conditionalPercAck {
	// add ack  
} else if rand.Float64() < conditionalPercPersistent {
	// add persistency
}
```
The computation `b / (1 - a)` gives the percentage of `b` given that `a` has not occurred. We can see this by representing all posts by a line L with length 1, then considering `a` the urgent posts and `b` the ack posts:
```
       a    b
     |---|-----|-----------| <- Length of L = 1
         ___________________
                1 - a
```
We know what percentage `a` represents over the whole line, and we know what percentage `b` represents over the whole line. To obtain the percentage that `b` represents over the line of length `1-a` (i.e., all posts that are *not* urgent), then we must simply divide `b` between `1-a`). That's why we need the `else if rand < b / (1 - a)`.

#### Alternative 2: recomputing frequencies

If we don't merge the actions, we need to recompute frequencies. Per https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/coverage-frequency.md, the `createPost` action has a frequency of 1.5 **by definition**, so that we can compare all other actions against it. However, this assumes that the `POST /api/v4/posts` endpoint univocally identifies the `createPost` action, and is not used in any other action.
This PR adds two additional actions that are identified by that endpoint, so we need to make it so that the sum of the frequencies of all these actions sum 1.5.

#### Summary
I first thought that alternative 1 was better, but it's true that computing those percentages may be a bit cumbersome, and they do hurt readability a lot, so we can go with alternative 2 if you're ok. Can you remind me of where you got the proposed frequencies in this PR? Is it from real data or an estimation?

I'm right now working on updating *all* action frequencies ([related doc](https://docs.google.com/document/d/1A4gRuJBymamP5cTgYkI-CJV3n4Sj4nF07Zn44seuaks/edit#)), so I can take care of computing the relative frequencies when I open the PR (probably tomorrow).

loadtest/store/memstore/store_test.go
```diff
@@ -771,3 +773,52 @@ func TestThreads(t *testing.T) {
...
+		// Concurrently read ack posts
+		posts, err := s.PostsWithAckRequests()
+		require.NoError(t, err)
+		for _, p := range posts {
+			mux.RLock()
+			assert.Contains(t, ackPosts, p)
+			mux.RUnlock()
+		}
+	}
+}
```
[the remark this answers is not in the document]
vish9812: Now come to think of it, I can actually remove the whole concurrency thing here. I was actually unit testing the `s.lock.RLock()` in the `PostsWithAckRequests` func which may not be _that_ necessary to unit-test.

vish9812:
@agarciamontoro Thanks for the great explanation for Alternative 1. But yeah I also agree, it would hurt readability.
 
So, as per alternative 2, it should be `CreateAckPost + CreatePersistentNotificationPost + CreatePost = 1.5` ?

> Is it from real data or an estimation?

It's an estimation.

agarciamontoro:
> So, as per alternative 2, it should be CreateAckPost + CreatePersistentNotificationPost + CreatePost = 1.5 ?

Yes. Although after https://github.com/mattermost/mattermost-load-test-ng/pull/577, it will be 1, not 1.5. So it's actually easier: if we know the percentage of ack posts and percentage of persistent-notifications posts, we can just set their frequency to that number, setting CreatePost's frequency to the remaining up to 1.0.

As soon as we have real data on that, we can adjust the frequencies.

loadtest/control/actions.go
```diff
@@ -215,6 +215,127 @@ func CreatePost(u user.User) UserActionResponse {
...
+		if len(cms) < 2 {
+			return UserActionResponse{Info: fmt.Sprintf("not enough users in the channel: %s", channel.Id)}
+		}
+	}
+
+	postOwnerID := u.Store().Id()
+	idx := rand.Intn(len(cms))
+	if cms[idx].UserId == postOwnerID {
+		idx = (idx + 1) % len(cms)
+	}
```
agarciamontoro, line 309: Neat! Maybe add a comment here, since it's not super obvious to a first-time reader what we're doing.

loadtest/control/simulcontroller/controller.go
```diff
@@ -130,7 +130,7 @@ func (c *SimulController) Run() {
 		},
 		{
 			run:       c.createPost,
-			frequency: 1.5,
+			frequency: 1.225,
```
agarciamontoro, line 133: Ok, let's keep this as is and I'll adjust in #577

agarciamontoro (approved):
Great! Thanks for the infinite patience and addressing all my nits (I left yet another one below :nerd_face:)

koox00 (approved):

agnivade:
@vish9812 @agarciamontoro - This is good to merge right? Since the feature PR has been merged.

vish9812:
Yes, the feature branch is merged but this repo is not pointing to the latest server-master which includes the path breaking changes of the sub-module `public`.
And if I don't update the `go.mod` to the latest server than the PN related loadtests won't run correctly, right?

agnivade:
You only need to update go.mod, if you need to use some new features from the latest server in your load-test client. Otherwise there isn't a need to update. We just point the  load-test to the new server.

vish9812:
`master` `go.mod` is currently pointing to `2023042` but my server-feature got merged after this. So, if I run the loadtests it would run against the current-pointed server, right? But then my tests will fail as the current-pointed server doesn't even have that feature.

So to make my tests pass, I'll need to update the `go.mod` to the latest which causes the `sub-module` paths problems.

agnivade:
>  So, if I run the loadtests it would run against the current-pointed server, right? But then my tests will fail as the current-pointed server doesn't even have that feature.

Current pointed server is just the server running at `ServerURL`. It doesn't have any relation to the go.mod of the load-test. Only if you need to import some method in a later version of Mattermost, then you need to update it.

agnivade (approved):
Would be also good to run a load test with this and compare with a version of master before the feature.

PR #545, opened 2023-01-29
MM-40056: Don't use sed if there are no IPs to replace

agarciamontoro:
#### Summary
#536 introduced a bug that caused the loading of the DB dump to fail when `PermalinkIPsToReplace` is empty, since the resulting `sed` command was malformed. These changes skip the `sed` command altogether when there is nothing to  replace.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50046

### DIFF
--- a/comparison/loadtest.go
+++ b/comparison/loadtest.go
@@ -156,7 +156,10 @@ func buildLoadDBDumpCmds(dumpFilename string, newIP string, permalinkIPsToReplac
 		sedRegex := fmt.Sprintf(`'s/%s/%s/g'`, match, replace)
 		replacements = append(replacements, sedRegex)
 	}
-	sedCmd := strings.Join(append([]string{"sed -r"}, replacements...), " -e ")
+	var sedCmd string
+	if len(replacements) > 0 {
+		sedCmd = strings.Join(append([]string{"sed -r"}, replacements...), " -e ")
+	}
 
 	var dbCmd string
 	switch dbInfo.Engine {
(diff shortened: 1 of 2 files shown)


comparison/loadtest_test.go
```diff
@@ -9,40 +9,61 @@ import (
...
+			Host:     "test-db-0.c0gzspkyf2r5.us-east-1.rds.amazonaws.com",
+			Engine:   "aurora-postgresql",
+		})
+		require.NoError(t, err)
+
+		// Check the slice contains only two commands: zcat and psql
+		require.Len(t, cmds, 2)
+		require.True(t, strings.HasPrefix(cmds[0], "zcat"))
+		require.True(t, strings.HasPrefix(cmds[1], "psql"))
+	})
```
agarciamontoro, line 29: This is the only new test in this file. The other part of the diff contain only whitespace changes.

comparison/loadtest.go
```diff
@@ -168,6 +171,10 @@ func buildLoadDBDumpCmds(dumpFilename string, newIP string, permalinkIPsToReplac
 		return []string{}, fmt.Errorf("invalid db engine %s", dbInfo.Engine)
 	}
 
+	if sedCmd == "" {
```
agnivade, line 174: Nit: Maybe invert the condition and check for `sedCmd != ""`? Feels more natural to me that way.

agnivade (approved):

comparison/loadtest.go
agarciamontoro: -1/5, so just inverted it :)

noxer (approved):
LGTM but why are we using `sed` at all? Couldn't we use the `regexp` package?

agarciamontoro:
> LGTM but why are we using sed at all? Couldn't we use the regexp package?

The existing code was already calling cli commands like zcat and mysql/psql to load the DB, so I decided to keep doing that. If we ever find it's too inefficient/difficult to maintain, we can change it.

PR #550, opened 2023-01-31
Bumping disk space to 100GiB

agnivade:
50GiB does not seem to be enough for multiple
load tests


agarciamontoro (commented):
I may wait a bit for this one, since disk space is not an issue I've ever seen. I'll try to reproduce when I try the GraphQL changes and try to understand the problem.

agnivade:
Makes sense

PR #552, opened 2023-02-01
make update-dependencies premiere

agarciamontoro:
#### Summary
Just testing the new rule before we need to do it in the next update-dependencies cycle :)

Steps were:
1. `make update-dependencies`
2. `make build-linux` to see if it compiled, which showed an error with the Prometheus API (yes, it changed in a minor version :shrug:)
3. `make test` to see if the tests ran successfully, which showed other instances of the same Prometheus API change in the test files.

#### Ticket Link
--


agarciamontoro:
From the dependabot PRs:
- [xmlquery](https://github.com/mattermost/mattermost-load-test-ng/pull/548) was updated here to the same version.
- But [fasthttp](https://github.com/mattermost/mattermost-load-test-ng/pull/548) was not, I guess because it's an indirect dependency (?) I'm not sure if we should try to update that one as well?

agnivade (approved):

noxer (approved):

PR #556, opened 2023-02-09
Use locally installed golangci-lint

streamer45:
#### Summary

It's been a while :) 

@coltoneshaw was having [some troubles](https://community.mattermost.com/core/pl/9iheekun8instro73reuafgquw) when building and it turns out we were not using the `golangci-lint` version we are installing but pointing to a system wide one most likely.


agarciamontoro (approved):
Huh. This may explain the linting differences we were getting in other PRs with CI. I assumed that was Go's version, but this seems more likely.

PR #558, opened 2023-02-10
MM-50356: Avoid a nil pointer dereference if RoundTrip fails

agarciamontoro:
#### Summary

Although @agnivade has already fixed this in the latest commit of [this open PR](https://github.com/mattermost/mattermost-load-test-ng/pull/544), I wanted to open this to discuss separately.

There are two alternatives to solve this:
1. What this commit does: use 0 as statusCode when RoundTrip fails. This gives more information to the metric, since 0 is not a valid HTTP status code, which informs that RoundTrip failed. However, if the users don't know this bit, seeing a 0 there in the metric data can be confusing.
2. Observe the metric only if t.transport.RoundTrip(req) finishes successfully. This avoids the confusion with a 0 statusCode, but it also skips all failing round trips, which most probably fail because of a timeout, which are useful to see in the metric.

I think having some potential confusion is better than skipping some data points, so I opted for one. Thoughts?

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-50356

### DIFF
--- a/loadtest/user/userentity/user.go
+++ b/loadtest/user/userentity/user.go
@@ -72,7 +72,11 @@ type ueTransport struct {
 func (t *ueTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	startTime := time.Now()
 	resp, err := t.transport.RoundTrip(req)
-	t.ue.observeHTTPRequestTimes(req.URL.Path, req.Method, resp.StatusCode, time.Since(startTime).Seconds())
+	var statusCode int
+	if resp != nil {
+		statusCode = resp.StatusCode
+	}
+	t.ue.observeHTTPRequestTimes(req.URL.Path, req.Method, statusCode, time.Since(startTime).Seconds())
 	if os.IsTimeout(err) {
 		t.ue.incHTTPTimeouts(req.URL.Path, req.Method)
 	}


streamer45:
Did this behaviour change after a Golang update? Surely we tested many kinds of request failures and timeouts before and I guess they always returned a non-nil response unless there was a regression on this side. 

As for how to deal with this case, as long as we successfully detect timeouts we should be fine and anything else causing a `nil` response could probably go into its own separate metric but not entirely opposed to the zero status code workaround which is likely the simplest fix although a little less clear.

agnivade:
I think it would be confusing to capture timeouts both in the `incHTTPTimeout`, and as a zero status metric in `observeHTTPRequestTimes`. What we could do is add another metric for all non-timeout errors to capture anything else, but I'd prefer to keep `observeHTTPRequestTimes` under `resp != nil`.

agarciamontoro:
> Did this behaviour change after a Golang update? Surely we tested many kinds of request failures and timeouts before and I guess they always returned a non-nil response unless there was a regression on this side.

@streamer45, I don't think anything changed in Golang side. The change was in #537, where I added some more info to the `HTTPRequestTimes` metric. One of the additional data points was the status code, for which I had to dereference the response (something that didn't happen until now, not without checking it first :)). As far as I understand, `RoundTrip` returns an error (and a nil response) when there's a timeout: at least that's how the code tracks it right now (see the lines right after the changes in this PR). The docs of the RoundTrip interface say: "RoundTrip must return err == nil if it obtained a response, regardless of the response's HTTP status code. **A non-nil err should be reserved for failure to obtain a response.**"

> I think it would be confusing to capture timeouts both in the incHTTPTimeout, and as a zero status metric in observeHTTPRequestTimes. What we could do is add another metric for all non-timeout errors to capture anything else, but I'd prefer to keep observeHTTPRequestTimes under resp != nil.

@agnivade, we were effectively capturing those in both metrics already, since `observeHTTPRequestTimes` was always called, even before #537 (it was not under `resp != nil`). And I think it still makes sense to include the times of the timeouts in observeHTTPRequestTimes, since for the client, the request did take all that time. I think it's interesting to see that when querying this metric.

All in all, I see three possibilities:
1. Keep things as they are in this PR.
2. Remove the status code for this specific metric in all cases.
3. Split the `HTTPRequestTimes` in two metrics: `HTTPRequestTimes` `FailedHTTPRequestTimes` (name to be determined), reserving the former for the cases where RoundTrip does not error and the latter for the RoundTrip failures.

I think I'm still inclined for option 1: option 2 removes information in most of the cases just for reducing confusion in one specific case, and option 3 seems to me more confusing (than the 0 status code) at the time of querying, having to take two metrics into account instead of one.

Still open to discussion, though :) Thoughts?

agnivade:
Sounds good, let's merge this. There are important things to worry about :)

streamer45 (approved):

PR #561, opened 2023-02-13
Remove build-windows dependency on build rule

agarciamontoro:
#### Summary
The `build-windows` rule was removed on #523, but we forgot to remove it from the list of dependencies for `build`. I guess I've been using `make package` and `make test` to test changes, and until now I had not run `make build` :shrug: 

#### Ticket Link
--

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ assets: ## Generate the assets. Install go-bindata if needed.
 	go generate ./...
 	go fmt ./...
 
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
It shouldn't be that hard to add Windows support back. I am thinking it's mostly about not running `ulimit` and `sysctl` specific stuff but anyway, I am clearly not going to fight for this :)

agarciamontoro:
My same exact thoughts, to be honest.

cpoile (approved):

PR #562, opened 2023-02-13
Merge createPost and createPostReply

agarciamontoro:
#### Summary
Action frequencies are computed against the `createPost` action: when possible, the [methodology](https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/coverage-frequency.md) is to compare the new API endpoint representing the new action against the `createPost` API endpoint; as the frequency of the `createPost` action is known, it's just a matter of multiplying that frequency by the ratio computed.

When reviewing the frequencies for https://mattermost.atlassian.net/browse/MM-50247, I realized that there is an imprecision in such methodology: the `createPost` **endpoint** does not represent the `createPost` **action**, as the endpoint is used for both creating root posts and for creating replies. Thus, it represents both the `createPost` and `createPostReply` actions. This makes the current frequency computation methodology invalid.

There are two solutions to this: we either change the methodology or the action:
1. Change the methodology to split the frequency of the `createPost` endpoint into `createPost` and `createPostReply` actions, according to the percentage of replies vs root posts in Community. This is possible, but I think it makes the computation more confusing, and if we go down this road, every time we add an action that creates a slightly different post, we'll have to change this computation.
2. Change how we tackle the `createPost` action, considering the posting of a message as a single action, regardless of the adjectives we give to the post. Inside such action, we would let the post be a reply, or an urgent action, or a post with persistent notifications, according to the percentages of such posts in Community over time.

I think 2. is the best solution here, and the one implemented in this PR. Specifically, I think it's better prepared for the future, as we will keep adding different properties to posts: instead of creating a single action for every nuance in how we send a post, we keep one action and decide its properties inside it under a `if rand.Float64() < c.PercentagePropertyX`. Note that this is how we already decide whether to attach files to a post or whether it's an urgent post.

In any case, I'm open to suggestions. I haven't tested this already, but the code is ready to review. Let me know your thoughts.

#### Ticket Link
--

### DIFF
--- a/loadtest/control/actions.go
+++ b/loadtest/control/actions.go
@@ -202,11 +202,25 @@ func CreatePost(u user.User) UserActionResponse {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
 
-	postId, err := u.CreatePost(&model.Post{
+	post := &model.Post{
 		Message:   "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
 		ChannelId: channel.Id,
 		CreateAt:  time.Now().Unix() * 1000,
-	})
+	}
+
+	// 85% of posts are replies
+	if rand.Float64() < 0.85 {
+		rootPost, err := u.Store().RandomPostForChannel(channel.Id)
+		if errors.Is(err, memstore.ErrPostNotFound) {
+			return UserActionResponse{Info: "no post to reply to"}
+		} else if err != nil {
+			return UserActionResponse{Err: NewUserError(err)}
+		}
+
+		post.RootId = rootPost.Id
+	}
+
+	postId, err := u.CreatePost(post)
 
 	if err != nil {
 		return UserActionResponse{Err: NewUserError(err)}
@@ -241,40 +255,6 @@ func EditPost(u user.User) UserActionResponse {
 	}
 
 	return UserActionResponse{Info: fmt.Sprintf("post updated, id %v -> %v", post.Id, postId)}
-}
-
-// CreatePostReply replies to a randomly picked post.
-func CreatePostReply(u user.User) UserActionResponse {
-	team, err := u.Store().RandomTeam(store.SelectMemberOf)
-	if err != nil {
-		return UserActionResponse{Err: NewUserError(err)}
-	}
-	channel, err := u.Store().RandomChannel(team.Id, store.SelectMemberOf)
-	if errors.Is(err, memstore.ErrChannelStoreEmpty) {
-		return UserActionResponse{Info: "no channels in store"}
-	} else if err != nil {
-		return UserActionResponse{Err: NewUserError(err)}
-	}
-
-	post, err := u.Store().RandomPostForChannel(channel.Id)
-	if errors.Is(err, memstore.ErrPostNotFound) {
-		return UserActionResponse{Info: "no post to reply to"}
-	} else if err != nil {
-		return UserActionResponse{Err: NewUserError(err)}
-	}
-
-	postId, err := u.CreatePost(&model.Post{
-		Message:   "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
-		ChannelId: channel.Id,
-		CreateAt:  time.Now().Unix() * 1000,
-		RootId:    post.Id,
-	})
-
-	if err != nil {
-		return UserActionResponse{Err: NewUserError(err)}
-	}
-
-	return UserActionResponse{Info: fmt.Sprintf("post reply created, id %v", postId)}
 }
 
 // AddReaction adds a reaction by the user to a random post.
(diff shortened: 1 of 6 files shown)


loadtest/control/actions.go
```diff
@@ -202,11 +202,25 @@ func CreatePost(u user.User) UserActionResponse {
...
-	postId, err := u.CreatePost(&model.Post{
+	post := &model.Post{
 		Message:   "Lorem ipsum dolor sit amet, consectetur adipiscing elit",
 		ChannelId: channel.Id,
 		CreateAt:  time.Now().Unix() * 1000,
-	})
+	}
+
+	// 85% of posts are replies
+	if rand.Float64() < 0.85 {
```
agarciamontoro, line 212: This number is taken from https://community.mattermost.com/private-core/pl/4otj7nxwxfb39q9ywp334gqzso, but I'm waiting for the results of the more general query requested in https://community.mattermost.com/private-core/pl/c89zba5dr3dutfjwks793ot9go

agarciamontoro:
Oh, and the frequencies will be fixed as soon as I open a PR for https://mattermost.atlassian.net/browse/MM-50247: I'll propose to set createPost frequency to 1 by definition and compute all others (the ones where it's possible) from there.

loadtest/control/actions.go
streamer45: This change is one of the main reasons I originally implemented it the way it is. If we start adding all these probability checks inside the generic actions we are essentially turning everything into a simulcontroller. In hindsight that may be okay as that's probably the only truly useful controller but mostly sharing a bit of context.

loadtest/control/simulcontroller/actions.go
```diff
@@ -682,86 +682,59 @@ func editPost(u user.User) control.UserActionResponse {
...
-
 	if err := sendTypingEventIfEnabled(u, channel.Id); err != nil {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
-	message, err := createMessage(u, channel, true)
+	// Select the post characteristics
+	isReply := rand.Float64() < c.config.PercentReplies
+	isUrgent := !isReply && (rand.Float64() < c.config.PercentUrgentPosts)
+	hasFilesAttached := rand.Float64() < 0.02
```
streamer45, line 698: Once upon a time there was a [ticket](https://mattermost.atlassian.net/browse/MM-24645). Sadly it never happened but I am happy to see that we are moving in that direction :)

streamer45 (approved):
I don't have strong opinions on this any longer (I certainly had back then) but to share some context I always thought it was better to keep actions as decoupled as possible and have the decision making at the highest level possible (controller's loop) to avoid polluting the code with an interminable series of if/else blocks. If we bring this to an extreme we could have a single action that runs everything based on a series of random variables but it would likely result in worse code overall.

With that said, I think the issue brought up here is valid as creating posts vs replies is an application detail so it makes more sense to unify them to avoid more confusion/incorrectness with tuning.

loadtest/control/simulcontroller/actions.go
agnivade: We need to bring you up to speed :)

loadtest/control/simplecontroller/controller.go
```diff
@@ -166,7 +166,6 @@ func (c *SimpleController) createActions(definitions []actionDefinition) error {
 		"CreateDirectChannel":  control.CreateDirectChannel,
 		"CreateGroupChannel":   control.CreateGroupChannel,
 		"CreatePost":           control.CreatePost,
-		"CreatePostReply":      control.CreatePostReply,
```
agnivade, line 169: We can keep the action in simpleController, no harm in that.

loadtest/control/actions.go
agnivade: Perhaps we can pass an additional param as `controllerType` and do this _only_ when it is a simulcontroller? Or otherwise, we can have a dedicated simulcontroller specific createPost and a generic createPost for simpleController.

But I agree that adding probability checks inside the top-level package functions is not a good idea.

agarciamontoro:
Actually, there's no need to change the generic CreatePost and CreatePostReply, let's keep the changes as isolated as possible

agnivade (approved):
This looks great

streamer45 (approved):
Looks good :+1:

agarciamontoro:
Running a quick smoke test here before merging this.

agarciamontoro:
After 90 minutes of test, the percentage of replies is closer to 60% than it is to the specified 85%. I wonder if there's something weird here.

agarciamontoro:
On a different note, the percentage of replies should be 15%, not 85% :facepalm:

agarciamontoro:
Results of the test with the modified percentage (only 18% of posts are replies): after being running for about 2.5h, the agents generated ~16.3K posts, out of which 12% are replies. I think it's safe to assume this is working now. I'll merge.

PR #563, opened 2023-02-15
MM-48315: Shutdown MM before running reset

agnivade:
`mattermost db reset` cannot run while an existing
MM instance is running. Therefore we shut it down prior
to running it.

https://mattermost.atlassian.net/browse/MM-48315

### DIFF
--- a/cmd/ltctl/reset.go
+++ b/cmd/ltctl/reset.go
@@ -79,23 +79,28 @@ func RunResetCmdF(cmd *cobra.Command, args []string) error {
 		clients []*ssh.Client
 	}{
 		{
-			msg:     "Resetting database",
+			msg:     "Shutting down MM server on primary...",
+			value:   "sudo systemctl stop mattermost",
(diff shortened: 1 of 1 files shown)


cmd/ltctl/reset.go
```diff
@@ -79,23 +79,28 @@ func RunResetCmdF(cmd *cobra.Command, args []string) error {
 		clients []*ssh.Client
 	}{
 		{
-			msg:     "Resetting database",
+			msg:     "Shutting down MM server on primary...",
+			value:   "sudo systemctl stop mattermost",
+			clients: []*ssh.Client{appClients[0]},
+		},
```
agarciamontoro, line 85: Do we need to do it on primary only?

agnivade: Yes, because we are running the `reset` cmd on primary only

agarciamontoro (approved):
Thanks, let's merge :)

PR #569, opened 2023-03-09
Added graphql request for loading channels and channelmembers

agnivade:
https://mattermost.atlassian.net/browse/MM-49702

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -1449,37 +1449,9 @@ func (ue *UserEntity) GetInitialDataGQL() error {
 			}
 	`}
 
-	buf, err := json.Marshal(input)
+	gqlResp, err := ue.getGqlResponse(input)
 	if err != nil {
 		return err
-	}
-
-	req, err := ue.prepareRequest(http.MethodPost,
-		getGQLURL(ue.client.URL),
-		bytes.NewReader(buf),
-		map[string]string{})
-	if err != nil {
-		return err
-	}
-
-	resp, err := ue.client.HTTPClient.Do(req)
-	if err != nil {
-		return err
-	}
-	defer closeBody(resp)
-
-	var gqlResp *graphql.Response
-	err = json.NewDecoder(resp.Body).Decode(&gqlResp)
-	if err != nil {
-		return err
-	}
-
-	if len(gqlResp.Errors) > 0 {
-		tmp := ""
-		for _, err := range gqlResp.Errors {
-			tmp += err.Error() + " "
-		}
-		return errors.New(tmp)
 	}
 
 	err = json.Unmarshal(gqlResp.Data, &q)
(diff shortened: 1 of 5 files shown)


agarciamontoro:
Tests are ongoing. I'll post results when they're finished.

loadtest/user/userentity/actions.go
```diff
@@ -1515,3 +1591,40 @@ func (ue *UserEntity) prepareRequest(method, url string, data io.Reader, headers
...
+	var gqlResp *graphql.Response
+	buf, err := json.Marshal(input)
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := ue.prepareRequest(http.MethodPost,
+		getGQLURL(ue.client.URL),
+		bytes.NewReader(buf),
+		map[string]string{})
```
agarciamontoro, line 1605: Nit: probably the compiler optimizes this, but what if we declare a single empty map outside of the function and use it here?

loadtest/user/userentity/actions.go
```diff
@@ -1515,3 +1591,40 @@ func (ue *UserEntity) prepareRequest(method, url string, data io.Reader, headers
...
+		return nil, err
+	}
+
+	resp, err := ue.client.HTTPClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+	defer closeBody(resp)
+
+	err = json.NewDecoder(resp.Body).Decode(&gqlResp)
```
agarciamontoro, line 1616: Super nit: can we move the definition of `gqlResp` right before its first usage, which happens in this line?

loadtest/user/userentity/utils.go
```diff
@@ -73,6 +73,14 @@ func convertToTypedRoles(input []gqlRole) []*model.Role {
 	return roles
 }
 
+func convertToRoleString(input []gqlRole) string {
+	roles := ""
+	for _, r := range input {
+		roles += " " + r.Name
+	}
+	return roles
+}
```
agarciamontoro, line 82: Another nit:
Why the space before and not after? I'd do something like
```suggestion
func convertToRoleString(input []gqlRole) string {
	roleNames := make([]string, 0, len(input))
	for _, r := range input {
		roleNames = append(roleNames, r.Name)
	}
	return strings.Join(roleNames, " ")
}
```

agarciamontoro (approved):
Code looks good! I left some nit comments, but nothing blocking.
I do have a general question: how are we using cursors? I see that both for Channels and ChannelMembers we use them, but I don't know how the] are actually implemented server side. Could you point me to the logic that handles that?

agnivade:
> I do have a general question: how are we using cursors? I see that both for Channels and ChannelMembers we use them, but I don't know how the] are actually implemented server side. Could you point me to the logic that handles that?

Sure: https://github.com/mattermost/mattermost-server/blob/febcac0308c7fb782322b900dc41a56f0e6d4404/api4/resolver.go#L18-L30

https://github.com/mattermost/mattermost-server/blob/febcac0308c7fb782322b900dc41a56f0e6d4404/api4/resolver.go#L75-L80

Does that help?

It's basically a base64 encoded string, that we decode and find the relevant ids to return the results after and then pass on to the SQL layer.

loadtest/user/userentity/actions.go
agnivade: Hurts readability. Probably zero impact on performance. You are thinking too low-level. :P

loadtest/user/userentity/actions.go
agnivade: Done

loadtest/user/userentity/graphql.go
```diff
@@ -68,3 +68,47 @@ type gqlTeamMember struct {
...
+
+type gqlChannel struct {
+	Cursor   string  `json:"cursor"`
+	ID       string  `json:"id"`
+	CreateAt float64 `json:"createAt"`
+	UpdateAt float64 `json:"updateAt"`
+	DeleteAt float64 `json:"deleteAt"`
+	Team     struct {
+		ID string `json:"id"`
+	}
```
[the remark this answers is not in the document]
agnivade: Yep, good one

loadtest/user/userentity/graphql.go
```diff
@@ -68,3 +68,47 @@ type gqlTeamMember struct {
...
+	Name              string         `json:"name"`
+	Header            string         `json:"header"`
+	Purpose           string         `json:"purpose"`
+	LastPostAt        int64          `json:"lastPostAt"`
+	LastRootPostAt    int64          `json:"lastRootPostAt"`
+	TotalMsgCount     int64          `json:"totalMsgCount"`
+	TotalMsgCountRoot int64          `json:"totalMsgCountRoot"`
+	CreatorID         string         `json:"creatorId"`
+	SchemeID          *string        `json:"schemeId"`
+	GroupConstrained  *bool          `json:"GroupConstrained"`
```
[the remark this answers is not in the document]
agnivade: Mistake :)

loadtest/user/userentity/utils.go
agnivade: Better :)

loadtest/user/userentity/actions.go
agarciamontoro: Fair enough :nerd_face:

agarciamontoro:
Ah, I see that we order by ID when querying and compare with the cursor.  Ok, ok, still a GraphQL noob, thanks for the info :)

agarciamontoro:
Merging, thanks for addressing the nits!

PR #570, opened 2023-03-13
Update setPostReminder with accurate frequencies

agnivade:
Now that it's deployed on community we can find the correct frequency.

Calculated by following: https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/coverage-frequency.md

### DIFF
--- a/loadtest/control/simulcontroller/controller.go
+++ b/loadtest/control/simulcontroller/controller.go
@@ -170,7 +170,7 @@ func (c *SimulController) Run() {
 		},
 		{
 			run:       c.createPostReminder,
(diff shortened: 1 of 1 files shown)


agarciamontoro (commented):
What time window did you use for this? In the analysis in https://docs.google.com/document/d/1A4gRuJBymamP5cTgYkI-CJV3n4Sj4nF07Zn44seuaks/edit#, I saw no actual usage of this endpoint, but that may be skewed by the choice of the window.

agnivade:
We just deployed the feature 2 weeks back.

PR #576, opened 2023-03-17
Add option to exclude file count

agnivade:
The webapp does not include the file count any more
while loading a channel. We update the API to incorporate that.

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -433,7 +433,18 @@ func viewChannel(u user.User, channel *model.Channel) control.UserActionResponse
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
-	if err := u.GetChannelStats(channel.Id); err != nil {
+	excludeFileCount := true
+	// 1% of the time, users will open RHS, which will include the file count as well.
+	// This is not an entirely accurate representation of events as we are mixing
+	// a normal viewChannel with a viewRHS event
+	// But we cannot distinguish between the two at an API level, so our action
+	// frequencies are also calculated that way.
+	// This is a good enough approximation.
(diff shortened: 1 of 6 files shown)


loadtest/control/simulcontroller/actions.go
```diff
@@ -433,7 +433,18 @@ func viewChannel(u user.User, channel *model.Channel) control.UserActionResponse
...
 
-	if err := u.GetChannelStats(channel.Id); err != nil {
+	excludeFileCount := true
+	// 1% of the time, users will open RHS, which will include the file count as well.
+	// This is not an entirely accurate representation of events as we are mixing
+	// a normal viewChannel with a viewRHS event
+	// But we cannot distinguish between the two at an API level, so our action
+	// frequencies are also calculated that way.
+	// This is a good enough approximation.
+	if rand.Float64() < 0.01 {
```
agarciamontoro, line 443: Does this number come from telemetry, then? Or how did you get it?

agnivade: Nope, just random. Will update from telemetry.

agnivade: It looks like since all RHS events are conflated into one, it's hard to understand how frequently are users opening channel info. I think so far we have to go with 1%. :man_shrugging:

agarciamontoro: It looks sensible, let's go with it.

agarciamontoro (approved):
Thanks!

loadtest/control/simulcontroller/actions.go
agnivade: I got the results back and it is around 1.2% approx!! Wow, what an estimate that was!

PR #580, opened 2023-04-14
MM-52110: Avoid requesting the config at the same time for all users

agarciamontoro:
#### Summary
#544 introduced GraphQL, and with it a way to detect whether it was enabled in the server. In order to do this, the agents need to retrieve the client config, but this was not done in the usual flow of the actions, but as soon as `Run()` was called, even before login. This created lots of requests to the config endpoint at the exact same, clogging the server and making the client request times go over the roof. This, in turn, made the coordinator decrease the users, never reaching more than 150-200 users. Something like:

image

I fixed this by retrieving the client config during the login action, storing it in the user's memstore. This is then used to populate a featureFlags struct---that contains just the GraphQLEnabled flag--- in the simulcontroller, which can be later read by the actions.

This is how four tests look like after the fix (note that the client times never go above 1s, while they easily reached the 10s timeout before):

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52110

### DIFF
--- a/loadtest/control/actions.go
+++ b/loadtest/control/actions.go
@@ -58,6 +58,11 @@ func Login(u user.User) UserActionResponse {
 		return UserActionResponse{Err: NewUserError(err)}
 	}
 
+	// Populate user config
+	if err := u.GetClientConfig(); err != nil {
+		return UserActionResponse{Err: NewUserError(err)}
+	}
+
 	// Populate teams and channels.
 	teamIds, err := u.GetAllTeams(0, 100)
 	if err != nil {
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -77,7 +77,7 @@ func (c *SimulController) reload(full bool) control.UserActionResponse {
 	}
 
 	var resp control.UserActionResponse
-	if c.isGQLEnabled {
+	if c.featureFlags.GraphQLEnabled {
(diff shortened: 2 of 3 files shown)


agarciamontoro:
The previous tests are probably enough, but I'm starting the v7.10 performance comparison again (only the unbounded tests) to verify it works after this fix.

agnivade (approved):

agnivade:
Also curious - how did it work when you ran the load tests with and without gql? Wasn't it using the same code?

agarciamontoro:
Yup, but I ran bounded load-tests there, so this issue was not evident  (I guess I didn't properly look at the client API times, because it should have been shown there). Then, when I did the comparison between graphql on/off and generated the graphs, I only focused on the stable part after all users had logged in; i.e. when this issue is already gone: if the test successfully passes the phase of all users logging in, then it behaves normally.

isacikgoz (approved):
Awesome 💯

agarciamontoro:
Unbounded load-tests (capped to 10k users because I ran half the agents) look normal:
Postgres:
image

MySQL:
image

The only thing I've noticed is that GraphQL seems to have a soothing effect on the wavy pattern of the unbounded tests; i.e.,  the coordinator does not need to decrease a large amount of users. With just a bit of them, things go back to normal and the coordinator is able to increase the users once again.

PR #581, opened 2023-04-25
MM-52474: Remove aws_s3_bucket_acl resource from Terraform

agarciamontoro:
#### Summary

Starting this month (April 2023), AWS changed the default access policies for new S3 buckets. This broke our current deployment logic, which relied on the `aws_s3_bucket_acl` resource. ACLs have been deprecated for a while in favor of all Block Public Access settings, which are now enabled by default, while ACLs are disabled by default.

Long story short, this change in AWS prevents the load-test from creating new S3 buckets with the current configuration, in all releases of the load-test.

This PR attempts to just remove the resource, and while a quick test proved this to work, I'm testing it again with a larger test to double check. I'm not very experienced with these policies from S3, so I wanted you guys to take a look as well.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52474


agarciamontoro:
[This issue](https://github.com/hashicorp/terraform-provider-aws/issues/28353) in Terraform AWS provider seems to back the outcome of my experiments here:

> With the new default security settings, a private ACL definition can likely be removed completely as the default security settings already ensure public access is restricted upon bucket creation.

isacikgoz (approved):
LGTM

streamer45 (approved):
:+1:

agarciamontoro:
Things look good in the unbounded Postgres test: https://snapshots.raintank.io/dashboard/snapshot/5xp6MychTTCnVgpIM5CYqE42hZcbL1U4

Merging and releasing a new minor version with the fix.

PR #583, opened 2023-04-26
Change the server's package from v6 to v8

kostaspt:
#### Summary
Now that the server module has been changed to v8.0, we need to update the packages we use in this repo as well to access the latest API client.

#### Ticket Link
N/A


agarciamontoro (approved):
Thanks for this! I really appreciate that you also updated the docs :+1:

I just tested locally that it builds and all looks good, thanks again :tada:

PR #586, opened 2023-05-09
MM-52134: Add Pyroscope

agarciamontoro:
#### Summary
The latest performance comparison (which I'm about to publish) has been successfully executed with Pyroscope support, so this is tested and ready to be merged. Take a look at the profiles from the comparison, which is still up:
- http://3.80.114.74:4040/?query=mattermost.cpu%7B%7D&from=1683543095&until=1683577028
- http://34.238.117.190:4040/?query=mattermost.alloc_space%7B%7D&from=1683542957&until=1683556465
- http://3.84.137.35:4040/?query=mattermost.inuse_space%7B%7D&from=1683545078&until=1683580762
- http://23.22.163.153:4040/?query=mattermost.alloc_objects%7B%7D&from=1683545308&until=1683559877

#### Screenshots
And some screenshots in case you're reviewing this when the instance is already down:
##### Alloc-use space for a bounded test
image
##### CPU for an unbounded test
image

A couple of notes:
- For the performance comparison, I conservatively bumped the disk space to 200GiB in case Pyroscope data could clog the disk. However, after 23 hours of the instances running and with the largest test taking 9.5 hours, the disk with the most data had only 13GiB used. Thus, it's not needed to bump the disk size at all.
- The scrape interval is set to 60 seconds. I think we can safely increase this frequency, but this is the value tested, so let's use this one for now and keep iterating on future tests.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52134


isacikgoz (approved):
LGTM 👍

agnivade (approved):
Nice one

PR #589, opened 2023-05-23
MM-52856 Inject actions

wiggin77:
#### Summary
This PR adds a feature to load-test that allows actions to be injected into a running load test.  An example use case is to have all active users simulate a client reload simultaneously (thundering herd).  

Injected actions are run once, and at the very next opportunity, taking priory over scripted/scheduled actions.  Supports injecting actions via API or CLI.  Supports injecting actions directly to agents or coordinator.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52856


loadtest/control/simulcontroller/controller.go
```diff
@@ -118,136 +122,191 @@ func (c *SimulController) Run() {
...
+		}
+
+		// check for injected actions
+		select {
+		case ia := <-c.injectedActionChan:
+			injectedAction = &ia
+			continue
+		default:
+		}
+
```
agarciamontoro, line 309: This whole `for` is getting quite confusing. The structure is something like this, right?
```go
for {
  // 1. run any injected action
  // 2. check for new injected actions and skip the next iteration
  // 3. run a normal action
  // 4. wait for the idle time between actions OR receive a new injected action immediately and go to 1
}
```
I think 1+2 (in particular the `continue` at 2) and the OR in 4 are confusing. Now, I don't have good solutions:

I thought of moving 1+2 to a `checkAndRunInjectedAction` function, maybe swapping the order there: if injectedAction is nil, then read the channel and update injectedAction, then run it if not nil. But it's also quite confusing.

Any thoughts?

loadtest/control/simplecontroller/controller.go
```diff
@@ -107,9 +111,38 @@ func (c *SimpleController) Run() {
...
+				c.status <- c.newInfoStatus(resp.Info)
+			}
+			injectedAction = nil
+		}
+
+		// check for injected actions even if no scripted actions are present
+		select {
+		case ia := <-c.injectedActionChan:
+			injectedAction = ia
+			continue Loop
```
agarciamontoro, line 133: Do we need the label here?

agarciamontoro (changes requested):
First pass done, this is amazing! Thanks, @wiggin77 :)

I think the general design is perfect, I am only concerned about the complexity in the `for` loop in the controllers, we can discuss in the comment I added there. I also added a bazillion of nit comments, feel free to ignore those if you don't agree.

agarciamontoro:
Adding @coltoneshaw here. I don't expect a dev review, just a quick glance to see if this looks good for the use case you have.

api/server.go
```diff
@@ -95,6 +96,7 @@ func SetupAPIRouter(coordLog, agentLog *mlog.Logger) *mux.Router {
 	c.HandleFunc("/{id}/status", a.getCoordinatorStatusHandler).Methods("GET")
 	c.HandleFunc("/{id}/run", a.runCoordinatorHandler).Methods("POST")
 	c.HandleFunc("/{id}/stop", a.stopCoordinatorHandler).Methods("POST")
+	c.HandleFunc("/{id}/inject", a.coordinatorInjectActionHandler).Methods("POST").Queries("action", "{[a-zA-Z]+}")
```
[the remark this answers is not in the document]
wiggin77: Action IDs are all mixed case alpha's.  This is just to help catch errors.

cmd/ltctl/loadtest.go
```diff
@@ -156,3 +156,49 @@ func RunLoadTestStatusCmdF(cmd *cobra.Command, args []string) error {
...
+	}
+
+	t, err := terraform.New("", config)
+	if err != nil {
+		return fmt.Errorf("failed to create terraform engine: %w", err)
+	}
+
+	action := args[0]
+
+	status, errInject := t.InjectAction(action)
```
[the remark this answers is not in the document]
wiggin77: There isn't a definitive list of actions. In fact different controllers support different actions (with some overlap).  I guess I could try to create a list of all the known actions.  It's a trade-off between creating and maintaining that list, versus the potential to accidentally use an unsupported action.

wiggin77:
> First pass done

Thanks @agarciamontoro.  I appreciate you doing a thorough review.

cmd/ltctl/loadtest.go
```diff
@@ -156,3 +156,49 @@ func RunLoadTestStatusCmdF(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return err
+	}
+
+	usersCount, err := getUsersCount(helper)
+	if err != nil {
+		return err
+	}
+
+	printCoordinatorStatus(status, errInfo, usersCount)
```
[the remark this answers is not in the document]
wiggin77: I guess it's not necessary.  I was just being consistent with the other commands.  I think I'll remove it, as it may be misleading since the results of the action are likely not reflected in these numbers at this point.

coordinator/cluster/cluster.go
```diff
@@ -219,3 +219,14 @@ func (c *LoadAgentCluster) Status() (Status, error) {
...
+// InjectAction injects an action into all the agents. The action is run once,
+// at the next possible opportunity.
+func (c *LoadAgentCluster) InjectAction(actionID string) error {
+	for _, agent := range c.agents {
+		if _, err := agent.InjectAction(actionID); err != nil {
+			return fmt.Errorf("cluster: failed to inject action %s for agent: %w", actionID, err)
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
wiggin77: The actual agent.InjectAction is simply a queuing operation.  I think adding concurrency here is overkill.  I can collect errors in a `merror` and let all agents get the action injected, however if one fails at this stage the whole load-test likely needs to be modified and restarted.

loadtest/control/simulcontroller/controller.go
```diff
@@ -47,15 +49,17 @@ func New(id int, user user.User, config *Config, status chan<- control.UserStatu
...
-		disconnectChan: make(chan struct{}),
-		stopChan:       make(chan struct{}),
-		stoppedChan:    make(chan struct{}),
-		wg:             &sync.WaitGroup{},
+		id:                 id,
+		user:               user,
+		status:             status,
+		rate:               1.0,
+		config:             config,
+		injectedActionChan: make(chan userAction, 10),
```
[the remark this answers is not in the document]
wiggin77: Yes, potentially.  A script could do that and a bit of buffering can avoid blocking for most simple cases.

loadtest/control/simplecontroller/controller.go
wiggin77: If the code is changed in the future, such as another for loop added, this will ensure it does not break and explicitly indicates which loop the continue is targeting.

loadtest/control/simulcontroller/controller.go
```diff
@@ -118,136 +122,191 @@ func (c *SimulController) Run() {
...
 		},
 		{
+			name:      "GetInsights",
 			run:       c.getInsights,
 			frequency: 0.011,
 		},
 	}
+	for _, ua := range actions {
+		c.actionMap[ua.name] = ua
+	}
```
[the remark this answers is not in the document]
wiggin77: I'm not sure what you mean.  The hard-coded list of actions is above.  There are no duplicates currently, and it wouldn't matter if one was added later, provided the actions did the same thing.

loadtest/control/simplecontroller/controller.go
wiggin77: IMO, if there is a `break` or `continue` at any nesting level inside a labeled loop then the label should be used.

coordinator/cluster/cluster.go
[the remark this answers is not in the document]
wiggin77: I added an merror.

loadtest/control/simulcontroller/controller.go
```diff
@@ -302,3 +363,42 @@ func (c *SimulController) sendFailStatus(reason string) {
...
+
+	// include some actions that are not normally supported by SimulController
+	switch actionID {
+	case "Reload":
+		action = userAction{
+			name: "Reload",
+			run:  func(_ user.User) control.UserActionResponse { return c.reload(false) },
+		}
+	default:
+		action, ok = c.actionMap[actionID]
```
streamer45, line 382: Should we make sure this access is not concurrent with the initialization in `Run()`. They technically happen in separate goroutines so although unlikely it could be racy.

loadtest/loadtest.go
```diff
@@ -245,6 +246,24 @@ func (lt *LoadTester) Status() *Status {
...
+		return ErrNotRunning
+	}
+
+	merr := merror.New()
+
+	for _, ctrl := range lt.activeControllers {
+		if resp := ctrl.InjectAction(action); resp.Err != nil {
+			merr.Append(resp.Err)
+		}
+	}
```
streamer45, line 263: Related to the previous comment, we could be keeping `lt.mut` locked for several seconds which isn't ideal.

loadtest/control/simulcontroller/controller.go
streamer45: I wonder if we could run the injected action instead of the picked one to simplify and have a single `action` variable. The rest of the logic doesn't really care whether it's injected or picked.

streamer45 (commented):
Nice addition! Left a couple of comments on top of Alejandro's review.

loadtest/control/simplecontroller/controller.go
wiggin77: Loop has been refactored and no label is needed.

loadtest/control/simulcontroller/controller.go
```diff
@@ -302,3 +363,42 @@ func (c *SimulController) sendFailStatus(reason string) {
...
+			}
+		}
+	}
+
+	select {
+	case c.injectedActionChan <- action:
+		return control.UserActionResponse{
+			Info: fmt.Sprintf("Action %s queued successfully", actionID),
+		}
+	case <-time.After(time.Second * 15):
```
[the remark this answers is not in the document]
wiggin77: Fair point.  The idea around the queuing is that someone may inject a small batch of actions to perform some one-time simulation of real client activity.  Currently it will batch up to 10 actions, which seems reasonable, but going beyond that should really be done via a controller config.  So, I agree going beyond 10 actions in a short interval should fail immediately. Fixed.

loadtest/control/simulcontroller/controller.go
wiggin77: I moved the initialization of the actions list and map to the constructor.  They are only read after New.

loadtest/loadtest.go
wiggin77: Agreed.  I now release the lock quickly by making a defensive copy of the active controllers slice.

loadtest/control/simulcontroller/controller.go
wiggin77: I refactored the loop for SimulController and SimpleController.  I think this should address the concerns.

loadtest/loadtest.go
```diff
@@ -245,6 +246,29 @@ func (lt *LoadTester) Status() *Status {
...
 }
 
+func (lt *LoadTester) InjectAction(action string) error {
+	var activeControllers []control.UserController
+
+	lt.mut.Lock()
+
+	if lt.status.State != Running {
+		return ErrNotRunning
+	}
```
streamer45, line 256: Looks like we may return with the mutex locked here. But with the timeout gone now, I think we may as well lock for the whole function to keep it simple.

loadtest/control/simulcontroller/controller.go
```diff
@@ -302,3 +371,42 @@ func (c *SimulController) sendFailStatus(reason string) {
 func (c *SimulController) sendStopStatus() {
 	c.status <- control.UserStatus{ControllerId: c.id, User: c.user, Info: "user stopped", Code: control.USER_STATUS_STOPPED}
 }
+
+// InjectAction allows a named UserAction to be injected that is run once, at the next
+// available opportunity. These actions can be injected via the coordinator via
+// CLI or Rest API.
+func (c *SimulController) InjectAction(actionID string) control.UserActionResponse {
```
streamer45, line 378: I may have missed this but is there any special reason for not returning a simple `error` from this method? Given the action is run asynchronously I am not sure the added complexity of the wrapper is necessary.

streamer45 (commented):
Nice improvements, left just a couple of additional comments.

loadtest/loadtest.go
wiggin77: Good catch.  I removed the defensive copy, lock for whole method, and changed to a RLock since we are only reading `lt.status` and `lt.activeControllers`.

loadtest/control/simulcontroller/controller.go
wiggin77: I think I was following a pattern from elsewhere in the app.  Fixed.

streamer45 (approved):
Thanks @wiggin77, there's a test failing it seems but other than that it looks great :+1:

wiggin77:
@agarciamontoro can you please have a look again.

loadtest/control/simulcontroller/controller.go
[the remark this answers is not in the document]
agarciamontoro: Ah, I saw that you refactored this into `getActionList` and `getActionMap`. So we could either add my suggestion to `getActionMap` or add a test for `SimulController.New` that checks that both `actionList` and `actionMap` in the created controller have the same length.

loadtest/control/simulcontroller/controller.go
```diff
@@ -12,269 +12,338 @@ import (
...
 
 		select {
 		case <-c.stopChan:
 			return
 		case <-time.After(control.PickIdleTimeMs(c.config.MinIdleTimeMs, c.config.AvgIdleTimeMs, c.rate)):
+		case ia := <-c.injectedActionChan: // run injected actions immediately
+			c.runAction(&ia)
 		}
 	}
+}
```
agarciamontoro, line 326: Much simpler, thank you!

agarciamontoro (changes requested):
Amazing work! The only blocking thing is the test on the controllers New functions, to make sure that the map and the list are of the same length (what I commented [here](https://github.com/mattermost/mattermost-load-test-ng/pull/589#discussion_r1209922575))

Btw, it's cool that we can also use the NoopController to inject actions. It's like an army of sleeper agents logged in, just waiting for an action to run. Neat.

agarciamontoro (approved):
It seems there are three actions in the simulcontroller with missing names. The test seems to be working!

Pre-approved for when the tests pass, thank you!

PR #590, opened 2023-05-24
MM-52896: Bump Postgres version to 14.8

agarciamontoro:
#### Summary
More than 50% of servers sending telemetry use a Postgres version greater than or equal to 13, so before performing the ceiling tests, we should upgrade the one used here, which is still 12.11.

I decided to go with the latest minor version in 14.x to get the newest version that didn't cause any additional work (we need slight changes when using Postgres 15.x, as stated in https://docs.mattermost.com/install/trouble-postgres.html). End of life for 14.x is November 12, 2026.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52896


agarciamontoro:
@DHaussermann, for testing this we'll need to run bounded and unbounded tests with the old and the new versions and compare them. Let's sync in Community.

agnivade (approved):

streamer45:
Historically we've been using the minimum supported version to ensure we test against the likely "weakest" performance wise. I am not opposed to the changes but something to keep in mind. If we go in this direction we may as well revisit MySQL and use v8 if it makes sense.

streamer45:
> I'm thinking that maybe we should simply convert that into a config setting.

That would be quite helpful to do more advanced comparisons. Should we open a ticket?

streamer45 (approved):

agarciamontoro:
> Should we open a ticket?

There it is: https://mattermost.atlassian.net/browse/MM-52938

DHaussermann:
I did load testing on this in late June. Based on comparisons I have put together the following summary.

**Postgres Upgrade load test summary**

- After an extra configuration change to RDS setup load testing with new Postgres 14.7 is working as expected.
- Some tooling issues were encountered that had to be worked through but, they were unrelated to the Postgres version change.
- No outstanding issues remain as a result of the version bump.

- **Bonded test results**:
  - CPU Utilization saw a small but noticeable improvement in performance. On average CPU utilization decreased by at least 5% and more at some points. 
  - Memory heap usage decreased very slightly with 14.7
  - Go Routines in use did increase slightly while the test was running with 14.7
  - API Times shows by far the biggest difference. On average the times seem to be at least 10% faster when running on 14.7 
  - Store times saw no appreciable changes


- **Unbonded test results**:
  - Not many clear differences in the metrics 
  - For most of the data the results are quite similar and don’t show much of a trend where on DB version is consistently performing better than the other
  - CPU utilization does show a very small difference on average where 12.11 utilization status is consistently a bit higher. So, there is a very slight improvement for 14.7 near the end of the test
  - The number of connected devices did not reach quite as high on the newer 14.7 database. Both tests ended when the count was between 14000 and 15000 but the 14.7 test ended with somewhere in the area of 50 fewer connected devices. 
  
  There were some concerned with Request per Second identified by @agarciamontoro He will post an update.

agarciamontoro:
Thanks, @DHaussermann! I did run a pair of bounded tests yesterday, with the old and new versions of Postgres.
Both tests had 6000 users concurrently connected, the results come from a one-hour time window, and they both ran v8.0.1 of the server. The version of the load-test was latest `master` for the base test, and this branch, commit [f204e63](https://github.com/mattermost/mattermost-load-test-ng/pull/590/commits/f204e639268214931ead57ca9eb74ee94b90d331) for the new test.
The results show no significant difference, except for P99 store times, that seem to have improved with the new version.
All considered, I think it's safe to merge this PR.
avg-api-times
avg-store-times
cpu-utilization
goroutines-in-use
heap-in-use
p99-api-times
p99-store-times
rps
stack-in-use

PR #591, opened 2023-05-29
MM-52932: Improve mem utilization of agents

agnivade:
- Tune down queue sizes
- Properly clear off maps and slices during memstore clear.
- Add reaction queue as well.
- Improve Cqueue implementation to be generic.

https://mattermost.atlassian.net/browse/MM-52932

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -363,11 +363,12 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 			Password:     password,
 		}
 		store, err := memstore.New(&memstore.Config{
-			MaxStoredPosts:          500,
-			MaxStoredUsers:          1000,
-			MaxStoredChannelMembers: 1000,
-			MaxStoredStatuses:       1000,
-			MaxStoredThreads:        500,
+			MaxStoredPosts:          250,
+			MaxStoredUsers:          500,
+			MaxStoredChannelMembers: 500,
+			MaxStoredStatuses:       500,
+			MaxStoredThreads:        250,
+			MaxStoredReactions:      10,
 		})
 		if err != nil {
 			return nil, err
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -23,18 +23,19 @@ type MemStore struct {
 	clientConfig        map[string]string
 	emojis              []*model.Emoji
 	posts               map[string]*model.Post
-	postsQueue          *CQueue
+	postsQueue          *CQueue[model.Post]
 	teams               map[string]*model.Team
 	channels            map[string]*model.Channel
 	channelStats        map[string]*model.ChannelStats
 	channelMembers      map[string]map[string]*model.ChannelMember
-	channelMembersQueue *CQueue
+	channelMembersQueue *CQueue[model.ChannelMember]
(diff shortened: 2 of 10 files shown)


api/agent.go
```diff
@@ -363,11 +363,12 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
-			MaxStoredUsers:          1000,
-			MaxStoredChannelMembers: 1000,
-			MaxStoredStatuses:       1000,
-			MaxStoredThreads:        500,
+			MaxStoredPosts:          250,
+			MaxStoredUsers:          500,
+			MaxStoredChannelMembers: 500,
+			MaxStoredStatuses:       500,
+			MaxStoredThreads:        250,
+			MaxStoredReactions:      10,
```
streamer45, line 371: I know it's a bit more work but I'd really like to see these exposed through the agent's config. We currently have a use case of generating the data but real clients don't really have limits to how many entities they can store so from a simulation quality perspective it would be best to keep these higher when possible.

streamer45 (approved):
Brilliant work, would be great to do a quick comparison, maybe on the existing cluster.

api/agent.go
agnivade: >  but real clients don't really have limits to how many entities they can store so from a simulation quality perspective it would be best to keep these higher when possible.

It'll be a tradeoff essentially. In real life, the webapp can accumulate upto 1GiB of mem. (It's showing 500MiB right now). That would be very very costly in terms of resources. And I think that was one of the reason we didn't choose to go the route of headless browser emulation.

But I agree that it will be good to have them exposed via agent's config. Created: https://mattermost.atlassian.net/browse/MM-52951

agnivade:
> Brilliant work, would be great to do a quick comparison, maybe on the existing cluster.

Yep, will leave that to Alejandro to do that on his existing cluster.

agarciamontoro (approved):
Amazing work, and beautiful usage of generics :heart_eyes:

I left a nit and a question for my own understanding, nothing blocking! :)

loadtest/store/memstore/store.go
```diff
@@ -775,21 +776,22 @@ func (s *MemStore) SetEmojis(emoji []*model.Emoji) error {
...
+	// if it points to the same memory location. If so, we delete it since it means the queue is full.
+	// This is done to keep the data pointed by the map consistent with the data stored in the queue.
+	r := s.reactionsQueue.Get()
+	if rs, ok := s.reactions[r.PostId]; ok && rs[len(s.reactions[r.PostId])-1] == r {
+		rs[len(s.reactions[r.PostId])-1] = nil
+		s.reactions[r.PostId] = rs[:len(rs)-1]
+	}
+
+	*r = *reaction
+	s.reactions[r.PostId] = append(s.reactions[r.PostId], r)
```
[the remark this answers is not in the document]
agnivade: The basic idea is that Cqueue is a ring buffer of pointers and as we keep adding reactions, the buffer gets filled up. When it gets full, it means the next pointer would already exist in the map (since it's circular). So we delete the entry in the map, and overwrite the pointer value in the buffer to be the new one.

This way we keep the total number of entries in the map to be limited.

PR #592, opened 2023-05-30
MM-52961: Set CurrentChannel properly in genController

agnivade:
The createReply action depended on the currentChannel being set.
Without that, it was throwing an error of channel not found.
So we add a custom switchChannel action to achieve that.

https://mattermost.atlassian.net/browse/MM-52961

### DIFF
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -222,8 +222,14 @@ func (c *GenController) createReply(u user.User) control.UserActionResponse {
 
 	var rootId string
 	var channelId string
-	channel, err := u.Store().CurrentChannel()
+	team, err := u.Store().RandomTeam(store.SelectMemberOf)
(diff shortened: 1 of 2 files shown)


agnivade:
Done

streamer45 (approved):

agarciamontoro (commented):
I've faced something similar when adding more actions to the gencontroller for https://mattermost.atlassian.net/browse/MM-52956, but I opted for choosing a random channel in those actions, instead of relying on the current channel. I think that that is easier to understand, since the gencontroller does not need to know the channel the user is in, only the channels the user is part of, and usually pick one randomly.

Thoughts?

agnivade:
It would be a bit complicated to share the code because you'll run into additional if conditions. Let me just use random channel directly during creating a reply.

agarciamontoro (approved):
Looks good, thank you!

PR #596, opened 2023-06-01
Improve pickRandomKeyFromMap to use generics

agnivade:
Things become to ergonomic with generics.
You don't even need to have some tests because of
the compile-time validation.

### DIFF
--- a/loadtest/store/memstore/random.go
+++ b/loadtest/store/memstore/random.go
@@ -6,7 +6,6 @@ package memstore
 import (
 	"errors"
 	"math/rand"
-	"reflect"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store"
 	"github.com/mattermost/mattermost-server/server/v8/model"
@@ -162,7 +161,7 @@ func (s *MemStore) randomUser() (model.User, error) {
 		if err != nil {
 			return model.User{}, err
 		}
-		user := s.users[key.(string)]
+		user := s.users[key]
 		if user == nil || user.Id == "" {
 			return model.User{}, ErrInvalidData
 		}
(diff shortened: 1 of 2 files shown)


agnivade:
Sorry for spending some time in non-priority things :P

streamer45:
> Sorry for spending some time in non-priority things :P

I understand you couldn't resist :)

loadtest/store/memstore/random.go
```diff
@@ -6,7 +6,6 @@ package memstore
 import (
 	"errors"
 	"math/rand"
-	"reflect"
```
streamer45, line 9: This is what makes this PR totally worth it in my opinion.

loadtest/store/memstore/random.go
```diff
@@ -350,24 +349,24 @@ func (s *MemStore) RandomCategory(teamID string) (model.SidebarCategoryWithChann
...
 }
 
-func pickRandomKeyFromMap(m interface{}) (interface{}, error) {
-	val := reflect.ValueOf(m)
-	if val.Kind() != reflect.Map {
-		return nil, errors.New("memstore: not a map")
+func pickRandomKeyFromMap[K comparable, V any](m map[K]V) (K, error) {
+	var def K
+	if len(m) == 0 {
+		return def, ErrEmptyMap
```
streamer45, line 362: Purely out of curiosity, can't you initialize K in place?

streamer45 (approved):
:100:

loadtest/store/memstore/random.go
agnivade: I tried to do that but couldn't think of anything at the moment. I can't do `K{}` obviously, and nor can I do `new(K)` because we are returning K not *K. What else am I missing?

loadtest/store/memstore/random.go
agnivade: @isacikgoz will be sad :joy:

loadtest/store/memstore/random.go
streamer45: Looks like you can do `*new(K)`, jeez :roll_eyes:

agnivade: oh dear ..

agarciamontoro (approved):
Oh no, I think we've lost you to the generics fever!

PR #600, opened 2023-06-06
MM-52704: Limit SSH SG to only current IP

agnivade:
https://mattermost.atlassian.net/browse/MM-52704


deployment/terraform/assets/cluster.tf
```diff
@@ -17,6 +17,14 @@ data "aws_region" "current" {}
 
 data "aws_caller_identity" "current" {}
 
+data "http" "my_public_ip" {
+  url = "https://ifconfig.me/ip"
```
streamer45, line 21: Could we use https://checkip.amazonaws.com/ as a slightly more "official" service?

streamer45 (approved):
Nice, hopefully it won't cause too much pain due to dynamic IPs.

deployment/terraform/assets/cluster.tf
```diff
@@ -520,7 +528,7 @@ resource "aws_security_group_rule" "metrics-ssh" {
   from_port         = 22
   to_port           = 22
   protocol          = "tcp"
-  cidr_blocks       = ["0.0.0.0/0"]
+  cidr_blocks       = ["${local.ifconfig_result}/32"]
```
streamer45, line 531: I can see this becoming a bit of an annoyance as we tend to share access to grafana panels and this would essentially limit that without some manual intervention. Not sure if we could consider keeping this open.

deployment/terraform/assets/cluster.tf
agnivade: Oh interesting, wasn't aware of that. Will check

agnivade: I checked, and while AWS refers to that site to check for public IP, I couldn't find any official documentation for the tool. But given it's maintained by AWS, we can use that.

agnivade:
ping @agarciamontoro

PR #601, opened 2023-06-06
feat: added application readiness timeout option to te deployer

fmartingr:
#### Summary

In my first usage with the deployer I kept reaching timeouts on the Mattermost installation after the infrastructure was deployed, maybe due to only having one application instance.

I've added a new configuration parameter to set the timeout for the ping handler that checks for the application to be running correctly so I don't need to edit the timeout in code.

#### Ticket Link

--

### DIFF
--- a/config/deployer.sample.json
+++ b/config/deployer.sample.json
@@ -42,6 +42,7 @@
   "MattermostDownloadURL": "https://latest.mattermost.com/mattermost-enterprise-linux",
   "MattermostLicenseFile": "",
   "MattermostConfigPatchFile": "",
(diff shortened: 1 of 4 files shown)


agnivade:
https://github.com/mattermost/mattermost-load-test-ng/pull/599. I think 60 seconds should be sufficient for now.

agnivade:
Felipe - we merged #599. I'd say we can make this configurable if we start seeing issues with 60s as well. Otherwise, I don't want to add too many config knobs unnecessarily. LMK your thoughts.

PR #602, opened 2023-06-07
Add a FAQ to clarify how to configure a test to be bounded or unbounded

agarciamontoro:
#### Summary
Felipe was reading this bit of the docs and we realized it was slightly confusing, so I just reworded it to have three separate sections:
- Definition of bounded test
- Definition of unbounded test
- Configuration needed to choose between one or the other

Let me know if it's better now, we can keep iterating on it if needed.

#### Ticket Link
--


fmartingr (approved):
LGTM! Thanks!!

agnivade (approved):
Lovely

PR #604, opened 2023-06-08
Reduce number of calls to GetPostsForChannel

agnivade:
The problem with using a random channel would be that
there is a probability of the same channel coming up
again and we would unnecessarily make the GetPostsForChannel
call again for the same channel.

This was costly and would waste resources. Instead, we just
get the posts for the channel we just joined. This way
we don't get posts for the same channel again.


streamer45 (approved):
Thanks :+1:

PR #605, opened 2023-06-09
Improve gencontroller global counters handling

streamer45:
#### Summary

There were several instances where we would return early and forget to decrease the counters. PR tries to make it a little less error prone by using named returns in combination with a defer function. Also converting some info logs to warning to help discriminating whether the action was completed or not.


agnivade (approved):

agarciamontoro (approved):
Oh, this is smart! Thanks, this does simplify things and will avoid future bugs.

PR #607, opened 2023-06-13
MM-53080: Choose DM user ids from a list

agnivade:
Rather than selecting a random user from the store,
we pre-create all combinations of userIDs and go
through the list and create the DM channels.

https://mattermost.atlassian.net/browse/MM-53080

### DIFF
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -132,16 +132,9 @@ func (c *GenController) createDirectChannel(u user.User) (res control.UserAction
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
-	// TODO: make the selection a bit smarter and pick someone
-	// we don't have a direct channel with already.
-	user, err := u.Store().RandomUser()
-	if errors.Is(err, memstore.ErrLenMismatch) {
-		return control.UserActionResponse{Warn: "not enough users to create direct channel"}
-	} else if err != nil {
-		return control.UserActionResponse{Err: control.NewUserError(err)}
-	}
+	pair := st.getUserPair()
 
-	channelId, err := u.CreateDirectChannel(user.Id)
+	channelId, err := c.sysadmin.CreateDirectChannelWithUser(pair[0], pair[1])
 	if err != nil {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
@@ -422,7 +415,7 @@ func (c *GenController) joinChannel(u user.User) control.UserActionResponse {
 		}
 		resp = control.UserActionResponse{Info: fmt.Sprintf("joined channel %s", channelID)}
 
-		if err := c.user.GetPostsForChannel(channelID, 0, 60, collapsedThreads); err != nil {
+		if err := c.sysadmin.GetPostsForChannel(channelID, 0, 60, collapsedThreads); err != nil {
 			return control.UserActionResponse{Err: control.NewUserError(err)}
 		}
 	}
--- a/loadtest/control/gencontroller/state.go
+++ b/loadtest/control/gencontroller/state.go
@@ -5,6 +5,8 @@ package gencontroller
 
 import (
 	"sync"
+
+	combinations "github.com/mxschmitt/golang-combinations"
 )
 
 type state struct {
@@ -15,8 +17,15 @@ type state struct {
 	// This is used to store the global list of channelIDs for the agents
 	// to choose from while trying to join a channel. This only contains Open/Private
 	// channels.
-	channels              []string
-	channelsMut           sync.Mutex
+	channels    []string
+	channelsMut sync.Mutex
+	// This is used to store the global list of userIDs.
+	// Used to generate DM channels.
+	users                 []string
+	usersMut              sync.RWMutex
+	userCombinations      [][]string
+	userCombinationsOnce  sync.Once
+	userCombinationsIndex int
 	followedThreadsByUser map[string]map[string]bool
 	followedThreadsMut    sync.RWMutex
 }
@@ -59,6 +68,7 @@ func init() {
 		longRunningThreads:    make(map[string]*ThreadInfo),
 		channels:              []string{},
 		followedThreadsByUser: make(map[string]map[string]bool),
+		users:                 []string{},
 	}
 }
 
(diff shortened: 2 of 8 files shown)


loadtest/control/gencontroller/actions.go
```diff
@@ -422,7 +415,7 @@ func (c *GenController) joinChannel(u user.User) control.UserActionResponse {
 		}
 		resp = control.UserActionResponse{Info: fmt.Sprintf("joined channel %s", channelID)}
 
-		if err := c.user.GetPostsForChannel(channelID, 0, 60, collapsedThreads); err != nil {
+		if err := c.sysadmin.GetPostsForChannel(channelID, 0, 60, collapsedThreads); err != nil {
```
streamer45, line 418: Slightly unrelated question but how are we guaranteeing that users will have some posts in store to pick from, for example, to reply or react?

streamer45 (approved):
Nice 👍

loadtest/control/gencontroller/actions.go
agnivade: Aha good point there.. yeah that's a bug. I used sysadmin because, normal users weren't able to load posts from private channels. Which is why I used sysadmin for `AddChannelMember`  as well. Any suggestions?

streamer45: The problem is that after https://github.com/mattermost/mattermost-load-test-ng/pull/604 we essentially removed the call except for this case so it may become an issue even if we fix just this one. As things stand I suppose users will only reply and react to their own posts. 

I think we need to find a compromise to at least load posts every now and then, not as often as before to avoid killing the db but with some lower frequency maybe through a dedicated action. Another option, although likely more complex, would be to hook up the websocket to get the broadcasts but it may be overkill.

agnivade: I am currently writing a solution where I just use the sysadmin HTTP client to make the API call, and then use the normal user's store to store the posts. Essentially moving up the userentity logic to this upper layer. It's not nice, but then we need to do the same for `AddChannelMember` as well. Because otherwise, the channel members get added to the sysadmin's store, not the current user's store.

loadtest/control/gencontroller/state.go
```diff
@@ -112,6 +122,35 @@ func (st *state) storeChannelID(channelID string) {
...
+// getUserPair returns the pair at the current index
+// and increments the index
+func (st *state) getUserPair() []string {
+	st.usersMut.Lock()
+	defer st.usersMut.Unlock()
+
+	pair := st.userCombinations[st.userCombinationsIndex]
+	st.userCombinationsIndex++
+	return pair
+}
```
agarciamontoro, line 152: Maybe we should check the array length here

loadtest/control/gencontroller/actions.go
```diff
@@ -416,13 +410,28 @@ func (c *GenController) joinChannel(u user.User) control.UserActionResponse {
...
+				if err != nil {
+					return control.UserActionResponse{Err: control.NewUserError(err)}
+				}
+
+				err = c.joinTeamWithId(teamID, u)
+				if err != nil {
+					return control.UserActionResponse{Err: control.NewUserError(err)}
+				}
+				return control.UserActionResponse{Info: fmt.Sprintf("user wasn't a team member. Added to team %s", teamID)}
+			}
```
agarciamontoro, line 429: What if instead of this, we just join all teams for all users? As an init action, replace the current `joinTeam` by a new `joinAllTeams`.

agarciamontoro (commented):
Changes look great, thanks! I only have one suggestion: what if instead of handling the errors caused for users not being on a team, we simply join all teams in the first place? I left a comment below with some more info on this idea, we can discuss it there.

Apart from that, all looks good! I'm gonna give it a test locally.

loadtest/control/gencontroller/state.go
agarciamontoro: Indeed, I got a panic here. I don't know why it was empty, though, maybe it's some issue with my config.
```
panic: runtime error: index out of range [0] with length 0

goroutine 182 [running]:
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*state).getUserPair(0xc0001e2460)
	/home/alejandro/projects/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/state.go:149 +0xf5
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*GenController).createDirectChannel(0xc0003ce540, {0x15248e8, 0xc001dfd800})
	/home/alejandro/projects/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/actions.go:136 +0x215
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*GenController).runActions(0xc0003ce540, 0xc002793b20?, 0xc002b8fa68)
	/home/alejandro/projects/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/controller.go:232 +0x78
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*GenController).Run(0xc0003ce540)
	/home/alejandro/projects/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/controller.go:160 +0xa66
github.com/mattermost/mattermost-load-test-ng/loadtest.(*LoadTester).addUser.func1()
	/home/alejandro/projects/mattermost/mattermost-load-test-ng/loadtest/loadtest.go:131 +0x22
created by github.com/mattermost/mattermost-load-test-ng/loadtest.(*LoadTester).addUser
	/home/alejandro/projects/mattermost/mattermost-load-test-ng/loadtest/loadtest.go:130 +0x38b
exit status 2
```

loadtest/control/gencontroller/actions.go
agnivade: Hmm .. but that'll only work with dbs with existing teams right? Or do you mean create all the necessary teams first, then join to all those teams, then create necessary channels, and then go on with the rest of the actions.

agarciamontoro: Ah, that's true, I had not thought about that. But yes, we can
1. Create all teams
2. Join all teams
3. Create all channels
4. Do the rest

agnivade: Hmm .. this is getting more and more complicated.. Let's see what @streamer45 thinks.

PR #608, opened 2023-06-13
MM-53142: Join all teams and track global DMs

agarciamontoro:
#### Summary
Preliminary commits:
- https://github.com/mattermost/mattermost-load-test-ng/commit/d45a06b7e43999cb8dbb1261483aa120dd866e28: Verify the number of DMs in gencontroller's config to ensure that the number of users can satisfy the number of DMs.
- https://github.com/mattermost/mattermost-load-test-ng/commit/a60a299d4101175a689158bb3af189bafc0de0ee: Use sysadmin client in `AddChannelMember` and `GetPostsForChannel` (stolen from Agniva's changes in https://github.com/mattermost/mattermost-load-test-ng/pull/607)

Main commits:
- https://github.com/mattermost/mattermost-load-test-ng/commit/fd5e523d567f033aa134eb71c45bd7d33fc8ad77: Join all teams right after they are created and before creating channels. This adds a new phase to the gencontroller, which now is split into:
  1. Signup, login, join a single team and create a random team (up to the limit)
  2. (Wait for all teams to be created)
  3. Join all teams
  4. Create all channels
  5. Run the rest of the actions (create posts, add reactions, follow threads...)
- https://github.com/mattermost/mattermost-load-test-ng/commit/bbd2df1c0109bcac8248c8be3b2e7ed5ff8e7450: Globally track all DMs created, so that we don't attempt to create the same one twice. This is technically not needed, since we're using the custom Mattermost binary with https://github.com/mattermost/mattermost/pull/23680, but it will speed things up, since the retry logic will happen in place (we still randomly pick the users, but we keep trying until we get one with which a DM can be created), instead of in the server.

This PR does not tackle the lack of posts for users, I'm opening another one with a fix for that.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53142

### DIFF
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -430,13 +430,13 @@ func (c *GenController) joinChannel(u user.User) control.UserActionResponse {
 	if cm.UserId == "" {
 		// We use sysadmin to add channel in case it's a private channel.
 		// Otherwise normal users don't have permissions to join a private channel.
-		err = c.user.AddChannelMember(channelID, u.Store().Id(), c.sysadmin.Client())
+		err = c.sysadmin.AddChannelMember(channelID, u.Store().Id())
 		if err != nil {
 			return control.UserActionResponse{Err: control.NewUserError(err)}
 		}
 		resp = control.UserActionResponse{Info: fmt.Sprintf("joined channel %s", channelID)}
 
-		if err := c.user.GetPostsForChannel(channelID, 0, 60, collapsedThreads, c.sysadmin.Client()); err != nil {
+		if err := c.user.GetPostsForChannel(channelID, 0, 60, collapsedThreads); err != nil {
 			return control.UserActionResponse{Err: control.NewUserError(err)}
 		}
 	}
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -129,7 +129,7 @@ func (c *GenController) createDirectChannel(u user.User) (res control.UserAction
 
 	// Here we make a call to GetUsers to simulate the user opening the users
 	// list when creating a direct channel.
-	if _, err := u.GetUsers(0, 100); err != nil {
+	if _, err := u.GetUsers(0, c.numUsers); err != nil {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand"
+	"strings"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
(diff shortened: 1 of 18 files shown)


loadtest/control/gencontroller/controller.go
```diff
@@ -122,6 +123,21 @@ func (c *GenController) Run() {
 		}
 	}
 
+	// Wait for all teams to be created, so all users can join them all.
+	for st.get(StateTargetTeams) != c.config.NumTeams {
+		time.Sleep(time.Second)
+	}
+
+	// Join all teams
+	if resp := c.joinAllTeams(c.user); resp.Err != nil {
```
agnivade, line 132: This just made me realize. createTeam is just called once per user, because it's not repeated. So we can never create more teams than the number of users. I guess this doesn't come up in practice, but it's an important limitation to note.

agnivade:
Slightly unrelated, I got this:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xebc250]

goroutine 151 [running]:
github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore.(*MemStore).RandomPost(_, _)
        /home/agniva/mattermost/mattermost-load-test-ng/loadtest/store/memstore/random.go:247 +0x530
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*GenController).createReply(0xc0005301c0, {0x1503378, 0xc0005b8240})
        /home/agniva/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/actions.go:329 +0x5a8
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*GenController).runActions(0xc0005301c0, 0xc0001439d0?, 0xc000f79920)
        /home/agniva/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/controller.go:231 +0x78
github.com/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller.(*GenController).Run(0xc0005301c0)
        /home/agniva/mattermost/mattermost-load-test-ng/loadtest/control/gencontroller/controller.go:209 +0x1325
github.com/mattermost/mattermost-load-test-ng/loadtest.(*LoadTester).addUser.func1()
        /home/agniva/mattermost/mattermost-load-test-ng/loadtest/loadtest.go:131 +0x22
created by github.com/mattermost/mattermost-load-test-ng/loadtest.(*LoadTester).addUser
        /home/agniva/mattermost/mattermost-load-test-ng/loadtest/loadtest.go:130 +0x38b
exit status 2
```

The channel needs to be checked if it's nil or not. I'll send a separate PR.

agnivade (approved):
Looks good. I am still getting the "no threads to follow" warning even after applying the two commits. So it never finishes and keeps looking for threads to follow. No idea why.

loadtest/control/gencontroller/state.go
```diff
@@ -134,3 +140,41 @@ func copyThreadInfo(src *ThreadInfo) *ThreadInfo {
...
 	return dst
 }
+
+func orderUsers(user1, user2 string) (string, string) {
+	if user1 < user2 {
+		return user1, user2
+	}
+
+	return user2, user1
+}
```
streamer45, line 150: This feels more of a util, not state specific, I'd move it to the util file.

loadtest/control/gencontroller/actions.go
```diff
@@ -132,21 +133,32 @@ func (c *GenController) createDirectChannel(u user.User) (res control.UserAction
 		return control.UserActionResponse{Err: control.NewUserError(err)}
```
streamer45, line 133: Do we need to perhaps adjust the fetch size in the previous line to ensure we have enough users in store or is it covered somewhere else?

loadtest/control/gencontroller/actions.go
```diff
@@ -132,21 +133,32 @@ func (c *GenController) createDirectChannel(u user.User) (res control.UserAction
...
-	// TODO: make the selection a bit smarter and pick someone
-	// we don't have a direct channel with already.
-	user, err := u.Store().RandomUser()
-	if errors.Is(err, memstore.ErrLenMismatch) {
-		return control.UserActionResponse{Warn: "not enough users to create direct channel"}
-	} else if err != nil {
-		return control.UserActionResponse{Err: control.NewUserError(err)}
-	}
+	userID := u.Store().Id()
+	maxAttempts := 2 * (c.numUsers - st.numDMs(u.Store().Id()))
```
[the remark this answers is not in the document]
agarciamontoro: Whoops, I had a comment here, then removed the whole thing, then readded it but now without the comment :nerd_face: It's twice the number of users that can still be chosen. I'll add a comment.

loadtest/control/gencontroller/controller.go
agarciamontoro: Yup, that's true. I'll add a config validation.

loadtest/control/gencontroller/state.go
agarciamontoro: It's definitely a util, but right now very specific to the userDMs field and only used there. I think I'd keep it here until other code path needs to use it.

loadtest/control/gencontroller/actions.go
agarciamontoro: Great point! Maybe we can call it with the maximum number of users, since this number shouldn't be too large in practice. Thoughts?

loadtest/user/user.go
```diff
@@ -119,7 +122,7 @@ type User interface {
 	// SearchPosts performs a search for posts in the given teamId with the given terms.
 	SearchPosts(teamId, terms string, isOrSearch bool) (*model.PostList, error)
 	// GetPostsForChannel fetches and stores posts in a given channelId.
-	GetPostsForChannel(channelId string, page, perPage int, collapsedThreads bool) error
+	GetPostsForChannel(channelId string, page, perPage int, collapsedThreads bool, client *model.Client4) error
```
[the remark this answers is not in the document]
agarciamontoro: Let me give that a shot.

loadtest/control/gencontroller/actions.go
streamer45: That sounds reasonable to me :+1:

streamer45: Just make sure we don't hit a page limit server side or we may have to page through.

loadtest/control/gencontroller/actions.go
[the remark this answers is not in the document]
agarciamontoro: Fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/608/commits/2f2064c0a86141af762c802f06491e63712b1a39

loadtest/control/gencontroller/controller.go
agarciamontoro: Fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/608/commits/81adf7665e33880042bdc56958cfdfa98af910e2

loadtest/control/gencontroller/actions.go
agarciamontoro: As far as I can see, there is no server-side limit for this endpoint. The perPage parameter is the one used.

agarciamontoro: We're limited by the `memstore.Config.MaxStoredUsers`, though. That's right now set to 500, so it's exactly what we need, but maybe we should rethink all this in the future.

agarciamontoro:
@streamer45, here's my proposal: I removed the sysadmin change in this PR, so if you don't have any other comments, we can merge it. Then, to unblock the current efforts, I'll open a PR only with those changes, without the intent to merge, but only to use in the branch we're using for the tests. Thoughts?

agarciamontoro:
Here's the PR mentioned above: https://github.com/mattermost/mattermost-load-test-ng/pull/611

streamer45 (approved):
Thanks :+1:

PR #609, opened 2023-06-13
MM-53154: Add getPosts action

agarciamontoro:
#### Summary
Add a simple `1getPosts` action in the gencontroller that retrieves the first 200 posts of a random channel the user is a member of.

The frequency is set to 100, which is quite low, unless the number of channels or posts configured is even lower. Not sure if we should be too concerned about that, I'm open to discussing it.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53154

### DIFF
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -533,3 +533,23 @@ func (c *GenController) followThread(u user.User) (res control.UserActionRespons
 
 	return control.UserActionResponse{Info: fmt.Sprintf("followed thread %s", threadId)}
 }
+
+func (c *GenController) getPosts(u user.User) (res control.UserActionResponse) {
+	team, err := u.Store().RandomTeam(store.SelectMemberOf)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+	channel, err := u.Store().RandomChannel(team.Id, store.SelectMemberOf)
+	if errors.Is(err, memstore.ErrChannelStoreEmpty) {
+		return control.UserActionResponse{Warn: "no channels in store"}
+	} else if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	collapsedThreads := false
+	if err := c.user.GetPostsForChannel(channel.Id, 0, 200, collapsedThreads); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("got posts for channel %q", channel.Id)}
+}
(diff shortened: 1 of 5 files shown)


loadtest/control/gencontroller/actions.go
```diff
@@ -533,3 +533,23 @@ func (c *GenController) followThread(u user.User) (res control.UserActionRespons
...
+	}
+	channel, err := u.Store().RandomChannel(team.Id, store.SelectMemberOf)
+	if errors.Is(err, memstore.ErrChannelStoreEmpty) {
+		return control.UserActionResponse{Warn: "no channels in store"}
+	} else if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	collapsedThreads := false
+	if err := c.user.GetPostsForChannel(channel.Id, 0, 200, collapsedThreads); err != nil {
```
agnivade, line 550: FYI- you'll have to pass nil here once the other PR is merged.

loadtest/control/gencontroller/controller.go
```diff
@@ -192,6 +192,13 @@ func (c *GenController) Run() {
 			frequency:  int(c.config.NumFollowedThreads),
 			idleTimeMs: 1000,
 		},
+		// This action does not generate data, but is needed for all other actions
+		// to have data to work with
+		"getPosts": {
+			run:        c.getPosts,
+			frequency:  100,
```
agnivade, line 199: 100 is fine for lower number of posts. But for a batch size of 20M, this should be 10k. We can do a test run and then tune if needed.

agnivade (approved):

loadtest/control/gencontroller/controller.go
streamer45: Yeah, can we perhaps make this somewhat proportional with a low enough constant? Like 1/1000 of total posts or something. We can play with it and see I suppose but I'd avoid the static value.

streamer45 (approved):

loadtest/control/gencontroller/controller.go
agnivade: Yeah, something like max(100, totalPosts/1000) should be good.

loadtest/control/gencontroller/actions.go
agarciamontoro: Ok, will add it in the other PR, I'm merging this one first.

loadtest/control/gencontroller/controller.go
agarciamontoro: Good. Done in https://github.com/mattermost/mattermost-load-test-ng/pull/609/commits/749a330ca14fdb93891ad7c33d4bc70a7f84c626

PR #613, opened 2023-06-15
MM-52331: Automate dump ingestion for normal deployments

agnivade:
I moved some of the common code under deployment. There's
still slight repetition but it's only at a high level.

I wanted to have some level of freedom in case the workflow
changes between a deployment run and a comparison run. That's
why I refactored only the low-level plumbing.

https://mattermost.atlassian.net/browse/MM-52331

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -26,7 +26,22 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
-	return t.Create(true)
+
+	// If DBDumpURI is not set, we seed data. Otherwise,
+	// we just load the dump.
+	initData := config.DBDumpURI == ""
+	err = t.Create(initData)
+	if err != nil {
+		return fmt.Errorf("failed to create terraform env: %w", err)
+	}
+
+	if !initData {
+		err = t.IngestDump()
+		if err != nil {
+			return fmt.Errorf("failed to create ingest dump: %w", err)
+		}
+	}
+	return nil
 }
 
 func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -84,7 +84,7 @@ type Config struct {
 	// The file is expected to be gzip compressed.
 	// This can also point to a local file if prefixed with "file://".
 	// In such case, the dump file will be uploaded to the app servers.
-	DBDumpURI       string `default:""`
+	DBDumpURI string `default:""`
 	// An optional list of IPs present in the posts from the DB dump
 	// that contain permalinks to other posts. These IPs are replaced,
 	// when ingesting the dump into the database, in every post that
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -79,6 +79,11 @@ type Config struct {
 	TerraformStateDir string `default:"/var/lib/mattermost-load-test-ng" validate:"notempty"`
 	// URI of an S3 bucket whose contents are copied to the bucket created in the deployment
 	S3BucketDumpURI string `default:"" validate:"s3uri"`
+	// An optional URI to a MM server database dump file
+	// to be loaded before running the load-test.
+	// The file is expected to be gzip compressed.
+	// This can also point to a local file if prefixed with "file://".
+	// In such case, the dump file will be uploaded to the app servers.
 	DBDumpURI       string `default:""`
 	// An optional list of IPs present in the posts from the DB dump
 	// that contain permalinks to other posts. These IPs are replaced,
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -79,6 +79,13 @@ type Config struct {
 	TerraformStateDir string `default:"/var/lib/mattermost-load-test-ng" validate:"notempty"`
 	// URI of an S3 bucket whose contents are copied to the bucket created in the deployment
 	S3BucketDumpURI string `default:"" validate:"s3uri"`
+	DBDumpURI       string `default:""`
+	// An optional list of IPs present in the posts from the DB dump
+	// that contain permalinks to other posts. These IPs are replaced,
+	// when ingesting the dump into the database, in every post that
+	// uses them with the public IP of the first app instance, so that
+	// the permalinks are valid in the new deployment.
+	PermalinkIPsToReplace []string `validate:"each:ip"`
 }
 
 // TerraformDBSettings contains the necessary data
--- a/deployment/terraform/dump.go
+++ b/deployment/terraform/dump.go
@@ -1,0 +1,123 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package terraform
+
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/mattermost/mattermost-load-test-ng/deployment"
+	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
+	"github.com/mattermost/mattermost-server/server/v8/platform/shared/mlog"
+)
+
+func (t *Terraform) IngestDump() error {
+	output, err := t.Output()
+	if err != nil {
+		return err
+	}
+
+	extAgent, err := ssh.NewAgent()
+	if err != nil {
+		return err
+	}
+
+	if len(output.Instances) < 1 {
+		return fmt.Errorf("no app instances deployed")
+	}
+
+	appClients := make([]*ssh.Client, len(output.Instances))
+	for i, instance := range output.Instances {
+		client, err := extAgent.NewClient(instance.PublicIP)
+		if err != nil {
+			return fmt.Errorf("error in getting ssh connection %w", err)
+		}
+		defer client.Close()
+		appClients[i] = client
+	}
+
+	dumpURI := t.config.DBDumpURI
+	fileName := filepath.Base(dumpURI)
+	mlog.Info("Provisioning dump file", mlog.String("uri", dumpURI))
+	if err := deployment.ProvisionURL(appClients[0], dumpURI, fileName); err != nil {
+		return err
+	}
+
+	// stop
+	// reset
+	// load dump
+	// start
+
+	stopCmd := deployment.Cmd{
+		Msg:     "Stopping app servers",
+		Value:   "sudo systemctl stop mattermost",
+		Clients: appClients,
+	}
+
+	resetCmd, err := getResetCmd(t.config, output, appClients)
+	if err != nil {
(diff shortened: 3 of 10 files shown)


streamer45:
@agnivade Would this benefit from some testing? Dylan may be able to help if that's the case.

agnivade:
Yeah sure, I've personally tested it. But if @DHaussermann wants to give it a go, that's okay as well :+1: .

deployment/utils.go
```diff
@@ -0,0 +1,103 @@
...
+
+type DBSettings struct {
+	UserName string
+	Password string
+	DBName   string
+	Host     string
+	Engine   string
+}
+
+// ProvisionURL takes a URL pointing to a file to be provisioned.
```
agarciamontoro, line 28: I know this comment is copied over from the other location, but could we specify here that we can use both http(s):// and file:// as protocols here in the docs?

deployment/terraform/dump.go
```diff
@@ -0,0 +1,123 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/mattermost/mattermost-load-test-ng/deployment"
+	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
+	"github.com/mattermost/mattermost-server/server/v8/platform/shared/mlog"
+)
+
+func (t *Terraform) IngestDump() error {
```
agarciamontoro, line 16: Can we add some documentation here? Mainly what config is used and an overview of the commands it executes.

deployment/config.go
```diff
@@ -79,6 +79,13 @@ type Config struct {
 	TerraformStateDir string `default:"/var/lib/mattermost-load-test-ng" validate:"notempty"`
 	// URI of an S3 bucket whose contents are copied to the bucket created in the deployment
 	S3BucketDumpURI string `default:"" validate:"s3uri"`
+	DBDumpURI       string `default:""`
```
agarciamontoro, line 82: We should add the comment here as well.

cmd/ltctl/main.go
```diff
@@ -26,7 +26,22 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("failed to create terraform env: %w", err)
+	}
+
+	if !initData {
+		err = t.IngestDump()
+		if err != nil {
+			return fmt.Errorf("failed to create ingest dump: %w", err)
+		}
+	}
+	return nil
```
agarciamontoro, line 44: Does this logic take into account whether the DB already had something in it? I'm thinking of the following usecase:
1. I run `deployment create` to deploy the cluster. I had DBDumpURI configured so the dump is ingested.
2. I test something, generate a report, whatever
3. Now I change something in the deployment config (let's say for example that I add another node to the app nodes)
4. I run `deployment create` again so that the app node is created.

With this change, `deployment create` resets the DB as well, right? I guess this is ok for most usecases, but I'm wondering whether we should surface this somehow and make the user confirm that they are ok resetting the DB. Thoughts?

agarciamontoro (commented):
Yay, the changes look great, thanks! I have only one blocking comment that I want to discuss before approving.

cmd/ltctl/main.go
agnivade: Yep, we reset the DB. This is the behavior for comparison tests as well right?

Without resetting, I don't think it's even possible to restore a dump right? Then you have to take care of edge-cases like conflicts in IDs and things like that. Maybe we can give a warning at the start, but it seems pretty normal to me.

DHaussermann:
Okay I should be able to to take a look tomorrow when I'm done the load test I'm currently running 👍

cmd/ltctl/main.go
agarciamontoro: > This is the behavior for comparison tests as well right?

Yes, it is. But this is a more manual process, and I think it would be somewhat normal for people to play with the cluster specs running `deployment create` repeatedly without necessarily wanting to reset the DB.

I think the proper solution for this would be to have some flag that you can set to enable/disable the DB reset. Which is something we should do in other places of the CLI, but that's not something for this PR of course.

deployment/config.go
agnivade: done

deployment/terraform/dump.go
agnivade: done

deployment/utils.go
agnivade: done

cmd/ltctl/main.go
agarciamontoro: Good, that makes

agarciamontoro (changes requested):
Just one more thing: can you add the new config fields to the deployer config docs?

agnivade:
Ah good point. For sure

agnivade:
@agarciamontoro - PTAL

agarciamontoro (approved):
Looks great, thanks!

agnivade:
@DHaussermann - please go ahead with testing this one. Thanks!

deployment/config.go
```diff
@@ -79,6 +79,18 @@ type Config struct {
...
+	// The file is expected to be gzip compressed.
+	// This can also point to a local file if prefixed with "file://".
+	// In such case, the dump file will be uploaded to the app servers.
+	DBDumpURI string `default:""`
+	// An optional list of IPs present in the posts from the DB dump
+	// that contain permalinks to other posts. These IPs are replaced,
+	// when ingesting the dump into the database, in every post that
+	// uses them with the public IP of the first app instance, so that
+	// the permalinks are valid in the new deployment.
+	PermalinkIPsToReplace []string `validate:"each:ip"`
```
streamer45, line 93: I haven't followed the original implementation but asking here for clarity. Does using the first app node's IP to replace have performance consequences? I'd expect either the SiteURL to be used to allow for proper load balancing or otherwise equally distribute to all the available nodes IPs.

streamer45 (approved):
Nice! Great refactoring as well :100:

deployment/config.go
agnivade: That is actually true. In case of HA deployments, we can use the nginx siteURL. @agarciamontoro - thoughts?

agarciamontoro: Yeah, that's a good point. Created https://mattermost.atlassian.net/browse/MM-53348 to track this.

agarciamontoro (approved):

DHaussermann:
Hi @agnivade I think I understand the core concept here. Instead of a separate dump step after creation, I can pass in a Mattermost DB dump as a `.gz` and the create step would then also apply this DB dump for me correct?

Assuming I understand, I'm not seeing this work. Perhaps I'm trying this wrong. 

When I run create from this branch, I end up with an empty DB where the initial admin account has not yet been created. 

The way I tried to use this was to take the url to the DB that I normally dumb manually `https://lt-public-data.s3.amazonaws.com/12M_610_psql.sql.gz` and provide this in the new string you accept in the `DBDumpURI`. From what I see, I don't think this values was added to the `deployer.json` so, I saved it directly in the `config.go`
image
image

Can please you provide some guidance here on if I tried to test this correctly?

agnivade:
Hey Dylan, yep you got the concept right.

You need to add `DBDumpURI` to deployer.json. It wasn't added to deployer.sample.json. I just pushed a commit today that fixes it.

From your output, it appears that the deployment failed because it couldn't find the socket location for local mode. You would need to fix that.

When the deployment goes through, you will see logs like:

```
info  [2023-06-28 14:39:56.947 +05:30] Server status is OK                           caller="terraform/create.go:615"
info  [2023-06-28 14:39:56.947 +05:30] Deployment complete.                          caller="terraform/create.go:192"
info  [2023-06-28 14:39:59.201 +05:30] Provisioning dump file                        caller="terraform/dump.go:46" uri="https://lt-public-data.s3.amazonaws.com/100M_710_psql.sql.gz"

info  [2023-06-28 14:52:24.996 +05:30] Stopping app servers                          caller="terraform/dump.go:91"
debug [2023-06-28 14:52:24.996 +05:30] Running cmd                                   caller="terraform/dump.go:93" cmd="sudo systemctl stop mattermost"
info  [2023-06-28 14:52:25.740 +05:30] Resetting database                            caller="terraform/dump.go:91"
debug [2023-06-28 14:52:25.740 +05:30] Running cmd                                   caller="terraform/dump.go:93" cmd="export PGPASSWORD='<>' && dropdb -U mmuser -h agnivalt-wr.cluster-custom-
c0gzspkyf2r5.us-east-1.rds.amazonaws.com agnivaltdb && createdb -U mmuser -h agnivalt-wr.cluster-custom-c0gzspkyf2r5.us-east-1.rds.amazonaws.com agnivaltdb && psql -U mmuser -h agnivalt-wr.cluster-custom-c0gzspk
yf2r5.us-east-1.rds.amazonaws.com agnivaltdb -c 'ALTER DATABASE agnivaltdb SET default_text_search_config TO "pg_catalog.english"'"
info  [2023-06-28 14:52:27.471 +05:30] Loading DB dump                               caller="terraform/dump.go:91"
debug [2023-06-28 14:52:27.471 +05:30] Running cmd                                   caller="terraform/dump.go:93" cmd="zcat 100M_710_psql.sql.gz | psql 'postgres://mmuser:<>@agnivalt-wr.cluste
r-custom-c0gzspkyf2r5.us-east-1.rds.amazonaws.com/agnivaltdb?sslmode=disable'"
```

DHaussermann:
Thanks @agnivade this is very cool and saves a huge chuck of the setup process 🎉 
I see it working now
image
I'll poke at this a but and think on if there are more complex scenarios to tests. 

For now as to the comment: 
> "From your output, it appears that the deployment failed because it couldn't find the socket location for local mode. You would need to fix that."

I realize this is outside the scope of this PR but, I'm not sure that's something I can control directly but rather something Terraform is doing. @agarciamontoro is this something that can be solved. I notice n the UI I also get the websocket banner.
image
0/5 when saw this locally dev helped me solve it by adding `"AllowCorsFrom": "*",` to the `config.json
Should I open an issue for this?

agnivade:
The `AllowCorsFrom` is only needed when you are accessing MM from a different host. I don't think that is applicable in this context. Sometimes, the deployment is just flaky and runs into issues. One needs to manually log in to the app host to see what's happened and then resume the deployment again. 

If you can see any genuine bug in the deployment, then of course feel free to create a ticket. Otherwise, random flaky bugs would be hard to fix. 

Overall, I assume you were able to test this PR properly?

DHaussermann (approved):
Confirmed Tested and passed
- `DBDumpURI` now sits in the `deployer.json`
- Create works to apply the DB dump when the `.gz` has been provided
- No issues when running create again to apply a change from `deployer.json` such as adding an agent
- Running create with the `DBDumpURI` empty runs create as before without the DB dump applied

LGTM!

Thanks for the help @agnivade

DHaussermann:
@agnivade I have no permission to adjust labels in this repo. But, review is complete 👍

PR #614, opened 2023-06-15
Some minor improvements

agnivade:
- Corrected some log lines
- Calling GetUsers just once because it was overloading the servers and we didn't need that any more after all users have logged in.
- Paginating through GetUsers response

### DIFF
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -33,12 +33,17 @@ func logout(u user.User) control.UserActionResponse {
 	return control.UserActionResponse{Info: "logged out"}
 }
 
-func (c *GenController) login(u user.User) control.UserActionResponse {
-	res := control.Login(u)
-	if res.Err == nil && res.Warn == "" {
-		st.incUsers()
+func (c *GenController) login(u user.User) (res control.UserActionResponse) {
+	if !st.inc(StateTargetUsers, int64(c.numUsers)) {
+		return control.UserActionResponse{Info: "target number of users reached"}
 	}
-	return res
+	defer func() {
+		if res.Err != nil || res.Warn != "" {
+			st.dec(StateTargetUsers)
+		}
+	}()
+
+	return control.Login(u)
 }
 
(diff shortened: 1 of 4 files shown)


loadtest/control/gencontroller/actions.go
```diff
@@ -33,6 +33,14 @@ func logout(u user.User) control.UserActionResponse {
 	return control.UserActionResponse{Info: "logged out"}
 }
 
+func (c *GenController) login(u user.User) control.UserActionResponse {
+	res := control.Login(u)
+	if res.Err == nil && res.Warn == "" {
+		st.incUsers()
+	}
```
streamer45, line 40: What would happen if somehow one of the users fails to login? Is the action being repeated or we end up stuck waiting forever?

streamer45 (approved):
Great :+1:

loadtest/control/gencontroller/actions.go
agnivade: The latter. We don't have target checks for user logins, that's why I kept it as a single counter. But I realize that one failure, and we end up with an infinite loop.

Do you mind adding a fix? It would be just about converting this field to a StateTarget and changing the actions to a loop. Or something close to that.

streamer45: Okay, pushed the changes.

loadtest/control/gencontroller/controller.go
```diff
@@ -104,7 +105,7 @@ func (c *GenController) Run() {
 
 	for i := 0; i < len(initActions); i++ {
```
agnivade, line 106: This doesn't retry the action right? We probably need to break this out of the loop and call this via `c.runActions`.

streamer45: Init actions are executed again if they error as the loop counter is decremented so it should be okay I am thinking.

agnivade: Ah yes! Missed the `i--` :)

agarciamontoro (approved):
Code looks good, and the tests have been successful using these changes, so I'd say we can safely mergee.

PR #615, opened 2023-06-15
Fix init command

streamer45:
#### Summary

Changes in https://github.com/mattermost/mattermost-load-test-ng/pull/608 broke the `init` command as we used to set `InitialActiveUsers` to zero and add users later on, which would result in the config validation check for `gencontroller` to fail. 

I am not entirely sure why we used to add the users after starting the load-test but I can't see a strong reason to keep doing that. Please let me know if I am missing something.

### DIFF
--- a/cmd/ltagent/init.go
+++ b/cmd/ltagent/init.go
@@ -41,7 +41,7 @@ func isInitDone(serverURL, userPrefix string) (bool, error) {
 	return userentity.New(ueSetup, ueConfig).Login() == nil, nil
 }
 
-func genData(lt *loadtest.LoadTester, numUsers int64) error {
+func genData(lt *loadtest.LoadTester, numUsers int) error {
 	if err := lt.Run(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


cmd/ltagent/init.go
```diff
@@ -91,12 +85,12 @@ func RunInitCmdF(cmd *cobra.Command, args []string) error {
 	if ok, err := isInitDone(config.ConnectionConfiguration.ServerURL, userPrefix); err != nil {
 		return err
 	} else if ok {
-		mlog.Warn("init already done")
+		log.Warn("init already done")
```
streamer45, line 88: Unrelated but this log was never showing for me because we would exit right away. I added a defer above to flush the logger so it gets showed.

cmd/ltagent/init.go
```diff
@@ -50,14 +50,7 @@ func genData(lt *loadtest.LoadTester, numUsers int64) error {
 		mlog.Info("loadtest done", mlog.String("elapsed", time.Since(start).String()))
 	}(time.Now())
 
-	for lt.Status().NumUsersAdded != numUsers {
-		if _, err := lt.AddUsers(10); err != nil {
-			return fmt.Errorf("failed to add users %w", err)
-		}
-		time.Sleep(5 * time.Second)
-	}
```
agnivade, line 58: Maybe I am missing something, but if you don't add users, how will you have multiple controllers?

streamer45: We are setting the `InitialActiveUsers` to the number of users given. The difference is that now they'd start immediately, while before we'd ramping them up in batches.

agnivade: My bad. `lt.Run` has that logic internally. Sorry, a bit late here. My brain isn't working clearly :)

agnivade (approved):

agarciamontoro (approved):
Looks good, thanks!

PR #616, opened 2023-06-16
Sleep 200-500 ms after adding user to team

agarciamontoro:
#### Summary
There seems to be a data race in HA environments between adding a user to a member and asking for the channels in that team. We introduce a sleep (from 200 to 500 milliseconds) between those two calls to let the cluster catch up and avoid the read-after-write issue.

### DIFF
--- a/loadtest/control/gencontroller/actions.go
+++ b/loadtest/control/gencontroller/actions.go
@@ -465,7 +465,20 @@ func (c *GenController) joinAllTeams(u user.User) control.UserActionResponse {
 		if err := u.AddTeamMember(team.Id, userId); err != nil {
 			return control.UserActionResponse{Err: control.NewUserError(err)}
 		}
+
+		// The following sleep avoids a read-after-write issue that happens as follows:
+		//     1. The AddTeamMember call right before this comment adds the user to the team.
+		//     2. The teamMember and its roles are saved in the sessions cache, so the previous
+		//        request invalidates the cache in the node that processed the request.
+		//     3. This cache invalidation is sent to all other nodes in the cluster.
+		//     4. The GetChannelsForTeam call right after the sleep reads the cache to get the
+		//        teamMember and its roles.
+		// If 4 happens before 3 finishes, and the GetChannelsForTeam is processed by a different
+		// node than the one that processed AddTeamMember, then the node reads its cache thinking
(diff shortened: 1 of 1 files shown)


agnivade (approved):

streamer45 (approved):
Nice, if we ever plan to merge this let's write a big comment on top.

agnivade (approved):

streamer45 (approved):
Thanks!

agarciamontoro:
@agnivade, @streamer45, I'm proposing to merge this, since it seems to be working as expected. Let me know your thoughts :)

streamer45:
> @agnivade, @streamer45, I'm proposing to merge this, since it seems to be working as expected. Let me know your thoughts :)

I am thinking that it really depends on whether we merge https://github.com/mattermost/mattermost-load-test-ng/pull/617. If we add some proper retry logic then we technically don't need a sleep inside as eventually it should succeed. I still need to tweak the above PR to make sure that's the case but maybe it's a little more robust as an approach. Thoughts?

agnivade:
#617 definitely looks more logically sound, but we'll definitely need to add some sleep. Because even if we fix the retry, without sleep, you will never succeed and keep failing.

streamer45:
> #617 definitely looks more logically sound, but we'll definitely need to add some sleep. Because even if we fix the retry, without sleep, you will never succeed and keep failing.

There's always going to be some sort of sleep/wait mechanism before retrying the action as a whole. I am thinking that should be enough?

agnivade:
As long as that is proportional to the number of users, it should be fine. Because a low amount of sleep with 600 users also doesn't work.

agarciamontoro:
I'm ok closing this and merging #617, although merging both (this one as preventive code, #617 in case something goes wrong) shouldn't be a problem either. @streamer45, thoughts?

streamer45:
@agarciamontoro I'd propose to pause merging of either PRs until we get to a stable run. There's no strong need to have any of this on `master`. I'd like to test a bit before committing to any of this.

agarciamontoro:
Sounds good. Let's pause on merging then :)

agarciamontoro:
Btw, the latest successful batches have been generated without this PR, using only #617. I'm leaning towards closing this one.

agarciamontoro:
Closing this one without merging, we did not use it in the latest successful batches.

PR #619, opened 2023-07-05
Let deployments be made with zero agents

agnivade:
Sometimes, we just need a server only setup
to test out something. Or we may already have
an agent cluster and just want to deploy a
server side setup.

### DIFF
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -36,7 +36,7 @@ type Config struct {
 	// Type of the EC2 instance for app.
 	AppInstanceType string `default:"c5.xlarge" validate:"notempty"`
 	// Number of agents, first agent and coordinator will share the same instance.
-	AgentInstanceCount int `default:"2" validate:"range:[1,)"`
+	AgentInstanceCount int `default:"2" validate:"range:[0,)"`
 	// Type of the EC2 instance for agent.
 	AgentInstanceType string `default:"c5.xlarge" validate:"notempty"`
 	// Logs the command output (stdout & stderr) to home directory.
--- a/deployment/terraform/info.go
+++ b/deployment/terraform/info.go
@@ -47,9 +47,9 @@ func displayInfo(output *Output) {
 		for _, agent := range output.Agents {
 			fmt.Println("- " + agent.Tags.Name + ": " + agent.PublicIP)
 		}
-	}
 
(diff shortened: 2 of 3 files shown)


agarciamontoro (approved):
Another data point that supports the "double tool" theory about this repo. Hopefully we can split it into deployer tool and load-test tool at some point :)

deployment/config.go
```diff
@@ -36,7 +36,7 @@ type Config struct {
 	// Type of the EC2 instance for app.
 	AppInstanceType string `default:"c5.xlarge" validate:"notempty"`
 	// Number of agents, first agent and coordinator will share the same instance.
-	AgentInstanceCount int `default:"2" validate:"range:[1,)"`
+	AgentInstanceCount int `default:"2" validate:"range:[0,)"`
```
streamer45, line 39: Is this all that's needed? I remember having to comment certain checks in the past to make this work well.

streamer45: Cool. Would it be possible to fix this case? It could still be useful to show the state of things regardless of agents being created.

https://github.com/mattermost/mattermost-load-test-ng/blob/63a37f5353f808a9b436673a9236eede2f5a3add/deployment/terraform/info.go#L22-L26

agnivade: Yes, indeed that's one thing that I saw. Will fix.

agnivade:
@streamer45 - PTAL

deployment/terraform/info.go
```diff
@@ -47,10 +42,13 @@ func displayInfo(output *Output) {
...
-	for _, agent := range output.Agents {
-		fmt.Println("- " + agent.Tags.Name + ": " + agent.PublicIP)
+	if output.HasAgents() {
+		fmt.Println("Load Agent(s):")
+		for _, agent := range output.Agents {
+			fmt.Println("- " + agent.Tags.Name + ": " + agent.PublicIP)
+		}
 	}
+
 	fmt.Println("Coordinator: " + output.Agents[0].PublicIP)
```
streamer45, line 52: We likely need to wrap this one or :boom:

agnivade: Hehe nice one

streamer45 (approved):
Thanks :+1:

PR #621, opened 2023-07-11
MM-51553: Update AMI to use latest LTS

agnivade:
I was getting issues while installing the linux-tools-* package.
Bumping up the AMI version to use the latest package. Also, the AMI
hasn't been updated in a while. So it's good to do it.

https://mattermost.atlassian.net/browse/MM-51553


streamer45 (approved):
It's been a while, I just hope this is not going to randomly break things.

streamer45:
Yes, I was more worried about obscure failures like what we saw with a certain customer but agree this had to happen eventually.

agarciamontoro (approved):
Looks good! I'm wondering wheter we should batch the testing for this repo in a per-release test step, or do it per PR. In any case, we need testing here.

agnivade:
I'd say we can do it in a case by case basis. For this one, yes I'd lean on testing the PR itself.

DHaussermann:
@agarciamontoro I'm not sure the simplest way to test this.

I'm on Mac OS. I do have a linux piece of hardware available I can try and push up to the latest LTS and install this particular tool. Or perhaps there is some type of Docker solution for setup here?

Let me know if you have any thoughts.

agnivade:
@DHaussermann - All you need to do here is do a deployment with this and smoke test the server. That should be it. It does not matter what hardware you have locally.

agarciamontoro:
Yup, what @agnivade said. What this PR changed is the AMI used in the deployed instances, so it's just a matter of making a deployment and run a test. No need even to compare two tests, just make sure that running a test is successful and there's no regression.

agnivade:
@DHaussermann - Did you get a chance to look into this?

DHaussermann:
@agnivade I hit an unrelated tooling issue when I tried to test this where the code in the Terraform tooling was not able to record my IP address during the create process.
@agarciamontoro was helping me debug this but we could not solve it.

I will try this again on my Friday.

agnivade:
Gave it a bit of smoke testing. Nothing major overall. One problem that I noticed was that the websocket connection kept getting disconnected. I am assuming it's to do with AWS and my ISP. Also, somehow the `getChannelMember` call always gave 404. But I ran another test with master and that also gave same results so it's nothing to do with this specific change. 

Some interesting logs:
```
{"timestamp":"2023-08-07 05:44:15.725 Z","level":"warn","msg":"failed to get public IP address for local interface","caller":"app/plugin_api.go:979","plugin_id":"com.mattermost.calls","origin":"main.(*logger).Warn log.go:99","localAddr":"127.0.0.1","error":"failed to get public address: write udp4 127.0.0.1:8443->52.72.139.62:3478: sendto: invalid argument"}
```

Don't think it has anything to do with this change, but just wanted to highlight. @streamer45

streamer45:
> Gave it a bit of smoke testing. Nothing major overall. One problem that I noticed was that the websocket connection kept getting disconnected. I am assuming it's to do with AWS and my ISP. Also, somehow the `getChannelMember` call always gave 404. But I ran another test with master and that also gave same results so it's nothing to do with this specific change.

That's interesting. Regardless of this change, does it make sense to look into this? `getChannelMember` should not fail consistently I suppose.

> Some interesting logs:
>
> ```
> {"timestamp":"2023-08-07 05:44:15.725 Z","level":"warn","msg":"failed to get public IP address for local interface","caller":"app/plugin_api.go:979","plugin_id":"com.mattermost.calls","origin":"main.(*logger).Warn log.go:99","localAddr":"127.0.0.1","error":"failed to get public address: write udp4 127.0.0.1:8443->52.72.139.62:3478: sendto: invalid argument"}
> ```
> 
> Don't think it has anything to do with this change, but just wanted to highlight. @streamer45

I think that log is expected, although avoidable. In order to automatically figure out the public IP of the instance, we loop on all available interfaces and try to connect to the public STUN server. However, sending a udp message from the loopback device to a public IP is likely always going to fail. The following log line should be a successful one as eventually we should hit the interface actually routed to the public IP (172.x.x.x in our VPC).

agnivade:
> That's interesting. Regardless of this change, does it make sense to look into this? getChannelMember should not fail consistently I suppose.

It's something to do with our gencontroller I think. Makes sense to investigate. I can open a ticket unless you were already planning to.

streamer45:
> > That's interesting. Regardless of this change, does it make sense to look into this? getChannelMember should not fail consistently I suppose.
> 
> It's something to do with our gencontroller I think. Makes sense to investigate. I can open a ticket unless you were already planning to.

Not really, but agree, let's open a ticket and whoever finds the bandwidth can look in to it. Thanks!

agnivade:
https://mattermost.atlassian.net/browse/MM-53940

PR #623, opened 2023-07-14
MM-53644: Prevent panic when a metric value is not present

agnivade:
Sometimes it can happen that a p99 value for a corresponding avg metric
is not there. We need to handle that case.

https://mattermost.atlassian.net/browse/MM-53644


streamer45 (approved):
Thanks :+1:

agarciamontoro (approved):
Thanks, the check looks good! And thank you for the tests :) Do we have an example of why this could happen in a real-life scenario? I'm still a bit fuzzy on why this happens at all

agnivade:
Ideally it should not happen, but this seems like Prometheus is doing some optimizations. Because the query gets only a single value computing for the entire time range. So it might be the case that it got a value for avg but maybe it considered the value to be insignificant for a p99 metric. Then you will have a crash.

PR #625, opened 2023-07-21
MM-53777: Review fullReload

agarciamontoro:
#### Summary
This is the first investigation I did back then when we identified that there were discrepancies.

I'm opening this as a draft PR so that we are all on the same page, but it still needs more testing.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53777


agarciamontoro:
Closed in favor of #628

PR #627, opened 2023-07-25
MM-53730: Add more coverage

agnivade:
We add coverage for clicking a user profile and clicking
a permalink.

These result in additional coverage for getChannelMember
and getTeamMember.

https://mattermost.atlassian.net/browse/MM-53730


loadtest/control/simulcontroller/controller.go
```diff
@@ -192,6 +192,16 @@ func getActionList(c *SimulController) []userAction {
 			run:       control.CreatePersistentNotificationPost,
 			frequency: 0.05,
 		},
+		{
+			name:      "ClickUserProfile",
+			run:       c.clickUserProfile,
+			frequency: 0.03,
```
agnivade, line 198: @agarciamontoro - This frequency is calculated assuming 1.0 of `createPost` from SNOW and then multiplying the ratio. But I'll let you handle it since you have been recalculating it.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,152 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	oldestPost, err := u.Store().Post(postIds[0])
+	if err != nil && !errors.Is(err, memstore.ErrPostNotFound) {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	ch, err := u.Store().Channel(oldestPost.ChannelId)
```
agarciamontoro, line 1917: oldestPost can be nil at this point if `errors.Is(err, memstore.ErrPostNotFound)` is true, right?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,152 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+				if ch.Type != model.ChannelTypePrivate {
+					if err := u.AddChannelMember(ch.Id, u.Store().Id()); err != nil {
+						return control.UserActionResponse{Err: control.NewUserError(err)}
+					}
+				}
+			} else {
+				return control.UserActionResponse{Err: control.NewUserError(err)}
+			}
+		}
+	}
```
agarciamontoro, line 1956: This is a bit hard to follow. Thoughts on moving it into a separate helper function so we can return early and avoid the five-level indentation?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,152 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+
+	if err := u.GetChannelMembersForUser(u.Store().Id(), currentTeam.Id); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	// Switch to channel.
+	if resp := viewChannel(u, ch); resp.Err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(resp.Err)}
+	}
```
agarciamontoro, line 1976: Not something to change in this PR, this is something I realized in the frequencies analysis, but worth pointing out:
We call the `viewChannel` helper action in many places, so we need to fix the frequency of the `switchChannel` action (which is based on the `viewChannel` handler, of course).

loadtest/control/simulcontroller/controller.go
agarciamontoro: What endpoint did you use to compare against `createPost`?

agarciamontoro (changes requested):
Looks good! I left some comments below, but the approach is great, thank you! I see that the permalink logic is quite complex, thanks for taking the time to reflect it here :)

loadtest/control/simulcontroller/controller.go
agnivade: `clickUserProfile` would call out getTeamMember and getChannelMember. I saw that getChannelMember is 10x more frequent than getTeamMember and this is the only common place from where getTeamMember is called. So I selected the getTeamMember's frequency for clickUserProfile, and getChannelMember for clickPermalink. Assuming people click permalinks more often than they open user profiles.

loadtest/control/simulcontroller/actions.go
agnivade: It won't be, because `u.GetPostThreadWithOpts` inserts the posts into the store. So let me just return as an error if the post isn't found in the store.

loadtest/control/simulcontroller/actions.go
agnivade: Right .. it's tricky.

loadtest/control/simulcontroller/actions.go
agnivade: I removed one layer. It looks okay to me now. Let me know if you still think a separate helper function is necessary.

agarciamontoro: Ah, that's better, thanks!

loadtest/control/simulcontroller/controller.go
agarciamontoro: Do those actions happen to have a unique telemetry event we can inspect? Relying on endpoints that are used elsewhere should be our last resort, so maybe telemetry can help us here.

agnivade: Nope, just checked. There is no event for "profile picture click". There's something called "api_channels_permalink" but I don't think it's to do with clicking a permalink. So I guess we are on our own here.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,156 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	// If it's a DM/GM channel, the webapp still sends the current team
+	// the user is part of
+	if ch.TeamId == "" {
+		team, err := u.Store().CurrentTeam()
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+		if team != nil {
+			ch.TeamId = team.Id
+		}
```
streamer45, line 1866: Should we perhaps warn or error if the current team is missing?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,156 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	for _, p := range posts {
+		// Found a permalink.
+		if index := strings.Index(p.Message, "/pl/"); index != -1 {
+			// If there are multiple permalinks found, we will click on the last one
+			// in the channel. This naturally leads to an effect of having clicked all
+			// permalinks which emulates organic behavior.
+			start := index + len("/pl/")
+			idLen := 26 // All IDs are always 26-char long
+			postID = p.Message[start : start+idLen]
+		}
```
streamer45, line 1900: Thoughts on having the parsing logic in a util function?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,156 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	}
+	if postID == "" {
+		return control.UserActionResponse{Info: "no permalink found"}
+	}
+
+	// We fetch the post thread.
+	postIds, _, err := u.GetPostThreadWithOpts(postID, "", model.GetPostsOptions{
+		CollapsedThreads: true,
```
streamer45, line 1908: Do we need to use `control.CollapsedThreadsEnabled(u)` here?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1831,3 +1832,156 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	currentTeam, err := u.Store().CurrentTeam()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	} else if currentTeam == nil {
+		return control.UserActionResponse{Err: control.NewUserError(errors.New("current team should be set"))}
+	}
+
+	if _, err := u.GetChannelsForTeamForUser(currentTeam.Id, u.Store().Id(), true); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
```
streamer45, line 1971: Does the webapp ask for deleted channels as well?

streamer45 (commented):
Nice work, left some questions/comments.

loadtest/control/simulcontroller/actions.go
agnivade: Hmm .. let me confirm that. I didn't check the `is_deleted` flag.

agarciamontoro:
Sorry for the change, @agnivade, but could you rebase this branch on top of `feat/frequencies.review.nube` to merge it there? You may get weird conflicts by doing a plain rebase, so what I did in other cases was to copy the branch I wanted to rebase, hard-reset the original one to `feat/frequencies.review.nube` and cherry-pick the commits from the copied branch (I'm sure there are more clever ways to do it, but it worked in my case).

loadtest/control/simulcontroller/actions.go
agnivade: done

loadtest/control/simulcontroller/actions.go
agnivade: Yep, done

loadtest/control/simulcontroller/actions.go
agnivade: It doesn't.

loadtest/control/simulcontroller/controller.go
agarciamontoro: I see that `getTeamMember` is also called in the team members dropdonw and in some places in the admin console. Assuming we can ignore those, I'm ok with the frequencies as you computed them.

agarciamontoro (approved):

loadtest/control/simulcontroller/controller.go
agnivade: Yep, I saw those. I think we can ignore them. Those are like super rare things.

streamer45 (approved):
Looks good! :+1:

PR #629, opened 2023-07-26
Use new frequencies based on cloud customer

agarciamontoro:
#### Summary
These are the initial frequencies computed using the [Grafana](https://docs.google.com/spreadsheets/d/1l6cHa1vW3dCFI3-LqqFdiJTmYQD0FadoI3oPHkKPIZA/edit#gid=604980166), [DB](https://docs.google.com/spreadsheets/d/1l6cHa1vW3dCFI3-LqqFdiJTmYQD0FadoI3oPHkKPIZA/edit#gid=759740678) and [telemetry](https://docs.google.com/spreadsheets/d/1l6cHa1vW3dCFI3-LqqFdiJTmYQD0FadoI3oPHkKPIZA/edit#gid=539193463) data from the cloud customer.

#### Ticket Link
--

PR #630, opened 2023-07-27
MM-53734: Added coverage for ws reconnect

agnivade:
There are a bunch of API calls we make
when we recover from a ws reconnect.

For simplicity, we don't actually disconnect
and reconnect the connection. But rather emulate
the set of calls that the webapp does when that happens.

https://mattermost.atlassian.net/browse/MM-53734


agnivade:
@agarciamontoro - I would say we need to run another test with the base nube branch + this PR + https://github.com/mattermost/mattermost-load-test-ng/pull/627 + https://github.com/mattermost/mattermost-load-test-ng/pull/626 (once updated to remove the systemPing) and measure that with https://github.com/mattermost/mattermost-load-test-ng/pull/629 to get a final reading.

These PRs work with lots of API endpoints, so it's easy for the frequencies to change again.

loadtest/user/userentity/actions.go
```diff
@@ -694,8 +694,11 @@ func (ue *UserEntity) GetTeamMembersForUser(userId string) error {
 
 // GetUsersByIds fetches and stores the specified users.
 // It returns a list of those users' ids.
-func (ue *UserEntity) GetUsersByIds(userIds []string) ([]string, error) {
-	users, _, err := ue.client.GetUsersByIds(context.Background(), userIds)
+func (ue *UserEntity) GetUsersByIds(userIds []string, since int64) ([]string, error) {
+	opts := &model.UserGetByIdsOptions{
+		Since: since,
+	}
+	users, _, err := ue.client.GetUsersByIdsWithOptions(context.Background(), userIds, opts)
```
agarciamontoro, line 701: It seems that the server code applies the `u.UpdateAt > since` in the SQL query if the query parameter `since` is non-empty, even if it equals "0". Should we treat `since = 0` here as a special case and avoid passing the options? Something like (not tested):
```suggestion
func (ue *UserEntity) GetUsersByIds(userIds []string, since int64) ([]string, error) {
	opts :=  &model.UserGetByIdsOptions{}
	if since > 0 {
		opts.Since = since
	}
	users, _, err := ue.client.GetUsersByIdsWithOptions(context.Background(), userIds, opts)
```

loadtest/store/memstore/store.go
```diff
@@ -858,6 +858,19 @@ func (s *MemStore) GetUser(userId string) (model.User, error) {
...
 
+// Users returns all users in the store.
+func (s *MemStore) Users() ([]model.User, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	users := make([]model.User, len(s.users))
+	for _, u := range s.users {
+		users = append(users, *u)
+	}
```
agarciamontoro, line 869: If we use append we need to define the capacity, not the length, right?
```suggestion
	users := make([]model.User, 0, len(s.users))
	for _, u := range s.users {
		users = append(users, *u)
	}
```

loadtest/control/simulcontroller/utils.go
```diff
@@ -164,3 +164,11 @@ func findIndex(haystack []string, needle string) int {
...
 	return -1
 }
+
+func keys[K comparable, V any](in map[K]V) []K {
+	keys := make([]K, 0, len(in))
+	for k := range in {
+		keys = append(keys, k)
+	}
+	return keys
+}
```
agarciamontoro, line 174: Beautiful generics :heart:

loadtest/control/simulcontroller/controller.go
```diff
@@ -192,6 +192,11 @@ func getActionList(c *SimulController) []userAction {
 			run:       control.CreatePersistentNotificationPost,
 			frequency: 0.05,
 		},
+		{
+			name:      "WSReconnect",
+			run:       c.reconnectWebSocket,
```
agarciamontoro, line 197: Nit just for consistency: the other actions use the name of the function as the `name` field.
```suggestion
			name:      "ReconnectWebSocket",
			run:       c.reconnectWebSocket,
```

loadtest/control/simulcontroller/actions.go
```diff
@@ -1838,3 +1838,87 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+
+func (c *SimulController) reconnectWebSocket(u user.User) control.UserActionResponse {
+	team, err := c.user.Store().CurrentTeam()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	} else if team != nil {
+		// The webapp makes these requests only if the team
+		// is set in the state.
+
+		if _, err := u.GetChannelsForTeamForUser(team.Id, u.Store().Id(), false); err != nil {
```
agarciamontoro, line 1850: Nit
```suggestion
	if err != nil {
		return control.UserActionResponse{Err: control.NewUserError(err)}
	}

	// The webapp makes these requests only if the team
	// is set in the state.
	if team != nil {
		if _, err := u.GetChannelsForTeamForUser(team.Id, u.Store().Id(), false); err != nil {
```

loadtest/control/simulcontroller/actions.go
```diff
@@ -1838,3 +1838,87 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+		}
+		prefs, err := u.Store().Preferences()
+		if err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+
+		for _, p := range prefs {
+			switch {
+			case p.Category == model.PreferenceCategoryDirectChannelShow:
+				userIds[p.Name] = true
```
agarciamontoro, line 1891: Should this be UserId?
```suggestion
				userIds[p.UserId] = true
```

loadtest/control/simulcontroller/actions.go
```diff
@@ -1838,3 +1838,87 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	}
+
+	userIds := make([]string, len(users))
+	for i, u := range users {
+		userIds[i] = u.Id
+	}
+	// Assuming reconnect happened just now.
+	if _, err := u.GetUsersByIds(userIds, time.Now().UnixMilli()); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
```
agarciamontoro, line 1921: Should we use the time when the reconnectWebSocket function starts?

agarciamontoro (changes requested):
Lot of work to get our GetWebappPlugins! Thanks for all the effort. I left some comments below.

loadtest/control/simulcontroller/actions.go
agnivade: Ah no, UserId is the user to whom the preference belongs to. What we want are the userIDs which are in the DM channel category to show.

loadtest/control/simulcontroller/actions.go
agnivade: I don't see a huge difference eiher ways.

loadtest/user/userentity/actions.go
agnivade: No need. See the `(c *Client4) GetUsersByIdsWithOptions` code. It only sets the query param, if `since` is non-empty.  Which automatically does what you want.

agarciamontoro: Aaah, went straight to the server code, I had not checked the client code. Good then!

loadtest/control/simulcontroller/actions.go
agarciamontoro: Aaaah, got it, thanks!

loadtest/control/simulcontroller/actions.go
agarciamontoro: Ok, let's use this, I'm 0/5

agarciamontoro (approved):
Thanks!

loadtest/control/simulcontroller/actions.go
```diff
@@ -1840,6 +1840,91 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+		latestCreateAt := posts[len(posts)-1].CreateAt
+		if _, err := u.GetPostsSince(currentCh.Id, latestCreateAt, crtEnabled); err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+
+		// unique user ids from all posts in current Channel + users from DM prefs
+		userIds := make(map[string]bool)
+		for _, p := range posts {
+			userIds[p.UserId] = true
+		}
```
streamer45, line 1884: Could become a `UsersForPosts` util of some kind.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1840,6 +1840,91 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	// All users in state.
+	users, err := u.Store().Users()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	userIds := make([]string, len(users))
+	for i, u := range users {
+		userIds[i] = u.Id
+	}
```
streamer45, line 1919: Would it make sense to have a `UserIds()` store method that calls the generic `keys()` util you implemented on the stored map?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1840,6 +1840,91 @@ func (c *SimulController) createPostReminder(u user.User) control.UserActionResp
...
+	// Assuming reconnect happened just now.
+	if _, err := u.GetUsersByIds(userIds, time.Now().UnixMilli()); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("reconnected ws for user %s", u.Store().Id())}
```
streamer45, line 1925: I suspect the user id here is redundant as we should always print it out from the higher level.

streamer45 (approved):
Nice work!

loadtest/control/simulcontroller/actions.go
agnivade: removed.

loadtest/control/simulcontroller/actions.go
agnivade: Slightly leaning towards no here. Returning user objects is the low level function which the store should expose. Returning userIds on top of it, is extra data transformation that should be done in the higher layer. From an API standpoint, it would be redundant to return users and userIds both from the same interface.

loadtest/control/simulcontroller/actions.go
agnivade: done

PR #631, opened 2023-07-27
Store FF in MemStore

streamer45:
#### Summary

Unrelated to the current effort but while working on https://mattermost.atlassian.net/browse/MM-53825 I noticed we were storing FF in the controller which seems a bit odd. I moved that to the `MemStore` and added some basic parsing so that all existing boolean flags should get loaded automatically. Ideally we'd rely on `model.FeatureFlags` but the way flags are returned through the client config makes it hard without resorting to hardcodes  or `reflect` usage.

### DIFF
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -236,7 +236,7 @@ func (s *MemStore) SetClientConfig(config map[string]string) {
 	defer s.lock.Unlock()
 	s.clientConfig = config
 
-	// Popule FF
+	// Populate FF
 	ffPrefix := "FeatureFlag"
 	s.featureFlags = map[string]bool{}
 	for k, v := range s.clientConfig {
--- a/loadtest/store/memstore/store.go
+++ b/loadtest/store/memstore/store.go
@@ -239,6 +239,8 @@ func (s *MemStore) SetClientConfig(config map[string]string) {
 	// Populate FF
 	s.featureFlags = map[string]bool{}
 	for k, v := range s.clientConfig {
+		// We avoid an extra call to strings.HasPrefix by checking the returned length.
+		// If the prefix matches then the returned string must be shorter.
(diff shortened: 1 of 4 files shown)


agnivade (approved):

loadtest/store/memstore/store.go
```diff
@@ -225,6 +235,18 @@ func (s *MemStore) SetClientConfig(config map[string]string) {
...
+	// Populate FF
+	s.featureFlags = map[string]bool{}
+	for k, v := range s.clientConfig {
+		if ffKey := strings.TrimPrefix(k, "FeatureFlag"); len(ffKey) < len(k) {
+			v, err := strconv.ParseBool(v)
+			if err != nil {
+				continue
+			}
+			s.featureFlags[ffKey] = v
+		}
```
agarciamontoro, line 248: Took me a second to understand the condition here. Could we do the following, even if there's some repetition?
```suggestion
		if strings.HasPrefix(k, "FeatureFlag") {
			v, err := strconv.ParseBool(v)
			if err != nil {
				continue
			}
			s.featureFlags[strings.TrimPrefix(k, "FeatureFlag")] = v
		}
```

agarciamontoro (approved):
Looks good, thanks! I left a comment below, but non-blocking :)

loadtest/store/memstore/store.go
streamer45: Funny that's how it was and then I simplified in https://github.com/mattermost/mattermost-load-test-ng/pull/631/commits/2762b4ebab76520ae41d272057f619b1039c1470

agarciamontoro: Oh, I had not seen that :joy: I understand why you made the simplification, but my brain had to go through the non-simplified version to understand it.

streamer45: I am inclined to keep the simplification but will add a comment to explain what's going on.

agarciamontoro: Sounds good, thanks!

PR #633, opened 2023-08-01
MM-53729: Load profile images while doing user autocomplete

agnivade:
This logic was missed out and since we already do autocomplete,
therefore we make it more realistic.

https://mattermost.atlassian.net/browse/MM-53729

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -966,7 +966,7 @@ func createMessage(u user.User, channel *model.Channel, isReply bool) (string, e
 		if err != nil {
 			return "", err
 		}
-		if err := emulateMention(channel.TeamId, channel.Id, user.Username, u.AutocompleteUsersInChannel); err != nil && !errors.Is(err, errNoMatch) {
+		if err := emulateMention(u, channel.TeamId, channel.Id, user.Username, u.AutocompleteUsersInChannel); err != nil && !errors.Is(err, errNoMatch) {
 			return "", err
 		}
 		message += "@" + user.Username + " "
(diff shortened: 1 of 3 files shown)


loadtest/user/userentity/actions.go
```diff
@@ -974,10 +974,10 @@ func (ue *UserEntity) AutocompleteUsersInChannel(teamId, channelId, username str
 	}
 	usersMap := make(map[string]bool, len(users.Users)+len(users.OutOfChannel))
 	for _, u := range users.Users {
-		usersMap[u.Username] = true
+		usersMap[u.Id] = true
 	}
 	for _, u := range users.OutOfChannel {
-		usersMap[u.Username] = false
+		usersMap[u.Id] = false
```
agarciamontoro, line 980: Oh good catches!

agarciamontoro (approved):
Looks good! I think this will be more than enough, actually. Let's see in the next test. Thanks for the added coverage :)

loadtest/control/simulcontroller/actions.go
```diff
@@ -1126,6 +1126,9 @@ func searchPosts(u user.User) control.UserActionResponse {
 			if err != nil {
 				return control.UserActionResponse{Err: control.NewUserError(err)}
 			}
+			if err := getProfileImageForUsers(u, keys(users)); err != nil {
```
streamer45, line 1129: Unrelated but did we ever review the TODOs above? Noting that we are still living with some arbitrary choices which eventually will affect the output rates.

loadtest/user/userentity/actions.go
streamer45: Do these change have any practical differences? Aren't username unique as well?

Oh I guess they are needed now to be able to feed the ids to `getProfileImageForUsers` but it was working fine before.

streamer45 (approved):
Looks good :+1:

loadtest/control/simulcontroller/actions.go
agnivade: Don't think we did. I guess the challenge here would be that we cannot directly count the number of mentions because that wouldn't be allowed. What we might do is look for indirect things like rates for `autocompleteUsersInTeam` and other similar things to give a rough idea.

PR #634, opened 2023-08-01
[MM-53348] Use proxy instance IP to substitute permalinks IPs

streamer45:
#### Summary

PR partially addresses the issue of permalinks IPs substitution not being properly load balanced. If the target deployment has a proxy we use its IP to do the substitution.

We are left with the case of multiple app servers without a proxy which I am not sure if it's worth investing more time on it. The implementation would be much more complex, likely requiring a bash script or small Go program to make it work since we'd have to pick IPs dynamically when doing the substitution.

I added a warning if we hit this case but happy to hear your thoughts on the above.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-53348

### DIFF
--- a/comparison/loadtest.go
+++ b/comparison/loadtest.go
@@ -208,7 +208,7 @@ func initLoadTest(t *terraform.Terraform, buildCfg BuildConfig, dumpFilename str
 		Clients: []*ssh.Client{appClients[0]},
 	}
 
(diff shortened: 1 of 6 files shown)


agnivade (approved):

agnivade:
Ooh I like that. Good idea!

agarciamontoro:
This is actually based on an idea @isacikgoz proposed back when I originally implemented this, but I didn't fully understand it at the time, and as we needed the feature sooner than later (in hindsight it's clear that we could have spend some more time on it :joy:), we merged it as is today.
Anyway, ticket created: https://mattermost.atlassian.net/browse/MM-53909

PR #635, opened 2023-08-02
MM-53560: Do not set param group name if not defined

agnivade:
We were unconditionally setting the param group name
for an RDS cluster even when there were no parameters defined.

This would lead to an issue re-creating the cluster because
the param group woudn't really exist, but the db cluster
got created. Ideally, this should have thrown an error from
Terraform while trying to attach a param group that doesn't
exist. But apparently it can indeed attach a dummy group.

So we are left with manually checking if we need to set
a group or not.

https://mattermost.atlassian.net/browse/MM-53560


agarciamontoro (approved):
Oh, that's some weird behaviour from Terraform. An additional question: is the actual `db_params_group` resource created only when there is at least one param? Or should we add some more conditions there?

streamer45 (approved):
Nice find!

agnivade:
@agarciamontoro - This is getting merged to master. Please feel free to cp to the nube branch.

agarciamontoro:
> @agarciamontoro - This is getting merged to master. Please feel free to cp to the nube branch.

Do we actually need to do that? The cluster I'm using to test is already deployed and I don't plan to redeploy the RDS cluster at any point, so we can probably get away without cherry-picking to the nube branch.

agnivade:
Sure, totally up to you. Just wanted to give a heads up.

PR #636, opened 2023-08-03
MM-53845: Use external bucket

agnivade:
Add ability to use an external S3 bucket. This allows
us to restore a bucket before hand and just point to it.

https://mattermost.atlassian.net/browse/MM-53845


agnivade:
cc @coltoneshaw - Now you can remove the forked commit.

agnivade:
This also means we can get rid of the copy bucket logic and passing an external bucket URI. LMK if you want me to do this as part of this PR itself.

agarciamontoro (commented):
Changes look good! But I have a dumb question: how easy is it to get all the info from the bucket manually? I mainly mean the `AccessKeyId` and the `SecretAccessKey`. Can we get those from our accounts in the load-test? Is there another way we can specify it?
I vaguely recall Colton and Stu had some problems getting that information (not sure if it was maybe a customer's specific issue), but I may be mistaken.

agarciamontoro (approved):

streamer45 (approved):
Nice improvement :+1:

PR #638, opened 2023-08-04
MM-53845: Attach to an existing rds cluster

agnivade:
This is not completely straightforward and there
might be better ways to achieve this, but this feels
like a reasonable first step.

Our first problem is to be able to import the restored
cluster into our terraform state and add instances
to the cluster.

My initial approach was to add the import block into the
terraform file itself, but the main challenge was that
the import block does not allow targets to be variables.

So I had to move the command to be run as a CLI.

The next challenge was to attach security groups
once they are imported. This was necessary because
if a cluster is imported, rather than created, then
the security groups don't get attached by Terraform
automatically since the block is entirely skipped.

So we have to spawn a shell command to attach the security
groups.

Once these are done, terraform works as expected and
idempotently. I have tested with deploying multiple times
and it works fine.

The only catch is that destroying a setup does not
delete the cluster. That one we have to manually.
Don't think that should be a big deal. Again, we can
write another shell command to delete it, but leaving it
out for now.

https://mattermost.atlassian.net/browse/MM-53845


deployment/config.go
```diff
@@ -102,7 +102,7 @@ type TerraformDBSettings struct {
 	// Number of DB instances.
 	InstanceCount int `default:"1" validate:"range:[0,)"`
 	// Type of the DB instance.
-	InstanceType string `default:"db.r4.large" validate:"notempty"`
+	InstanceType string `default:"db.r6g.large" validate:"notempty"`
```
streamer45, line 105: Was this meant to go in? As far as I understand we are waiting for the reference architecture to decide which instance class to default to going forward.

deployment/config.go
```diff
@@ -113,6 +113,8 @@ type TerraformDBSettings struct {
 	EnablePerformanceInsights bool `default:"false"`
 	// A list of DB specific parameters to use for the created instance.
 	DBParameters DBParameters
+	// ClusterIdentifier indicates to point to an existing cluster
+	ClusterIdentifier string `default:""`
```
streamer45, line 117: To understand, would any of the above settings be applicable when `ClusterIdentifier` is used?

deployment/terraform/engine.go
```diff
@@ -40,6 +40,38 @@ func (t *Terraform) runCommand(dst io.Writer, args ...string) error {
 	mlog.Debug("Running terraform command", mlog.String("args", fmt.Sprintf("%v", args)))
 	cmd := exec.CommandContext(ctx, terraformBin, args...)
 
+	return _runCommand(cmd, dst)
+}
+
+func (t *Terraform) attachSG() error {
```
streamer45, line 46: Would it make sense to rename this to something a little more specific? As it is, it sounds like it could be used for any resource but it has a very limited use case right now.

deployment/terraform/engine.go
```diff
@@ -40,6 +40,38 @@ func (t *Terraform) runCommand(dst io.Writer, args ...string) error {
...
+	awsBin := "aws"
```
streamer45, line 47: Not opposed to this for time's sake but should we consider the more programmatic way of integrating with the official Golang AWS SDK as a possible future improvement?

streamer45 (approved):
Thanks, agreed it may feel a bit hacky but the rest of the deployment logic is far from perfect so I think it will do fine.

> The only catch is that destroying a setup does not
delete the cluster. That one we have to manually.

If we agree this is the right approach for the time being, I'd be leaning towards implementing the above now. In general I feel we should work towards a more predicable and user friendly tool and try to move away from leaving this sort of "gotchas" which may backfire in the future.

deployment/config.go
agnivade: No we already changed the defaults in deployer.sample.json. This got missed out. So yes, this is intended.

deployment/config.go
agnivade: Yep, it would. All of them should be applicable I believe.

deployment/terraform/engine.go
agnivade: Yep, you already have https://github.com/mattermost/mattermost/issues/22150. We can add this one to that as well.

deployment/terraform/engine.go
agnivade: ok

agnivade:
> If we agree this is the right approach for the time being, I'd be leaning towards implementing the above now.

Done

agarciamontoro (changes requested):
Neat! It's indeed a bit hacky, but a great first step, thanks for looking into it!
The only blocking thing is that we need to document the new `ClusterIdentifier` in the `docs/config/deployer.md`

If I understood everything correctly, the flow then would be:
1. Set up the cluster through the AWS backup thingie.
2. Set `ClusterIdentifier` in the `config/deployer.json` file to the ID of the created cluster.
3. Run `deployment create` normally.

I have two questions:
- How do we exactly do 1?
- This flow is only available for `deployment create`, not `comparison run`, right?

deployment/terraform/engine.go
```diff
@@ -40,6 +40,28 @@ func (t *Terraform) runCommand(dst io.Writer, args ...string) error {
...
+
+func (t *Terraform) runAWSCommand(ctx context.Context, args []string) error {
+	awsBin := "aws"
+	if _, err := exec.LookPath(awsBin); err != nil {
+		return fmt.Errorf("aws not installed. Please install aws. (https://aws.amazon.com/cli): %w", err)
+	}
+
+	var cancel context.CancelFunc
+	if ctx == nil {
+		ctx, cancel = context.WithTimeout(context.Background(), cmdExecTimeoutMinutes*time.Minute)
```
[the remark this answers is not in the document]
agnivade: The s3 bucket empty has its own context. Only the delete cluster and set security group have this timeout. The delete cluster is an async command, as in, it does not wait till the cluster is actually deleted. So 1 min is fine for that. Setting a SG is sync, but a relatively quick operation. 

Happy to bump it up to something higher if you prefer.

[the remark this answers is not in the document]
agarciamontoro: Ah perfect! It makes sense as a default, I guess, since we can overwrite it from the caller. Thanks for the explanation :)

agarciamontoro (approved):
:rocket:

PR #639, opened 2023-08-18
MM-54174: Fix WebsocketReconnect panic

agarciamontoro:
#### Summary
There are rare cases when the current channel seems to have no posts. In those cases, we skip loading the statuses.

Reviewing tip: the diff is way nicer when hiding whitespaces. I just wrapped a block of code inside an `if` :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54174


agnivade (approved):
Nice!

And yes, hide whitespace indeed looks better.

PR #640, opened 2023-08-18
MM-53646: First version of the release process

agarciamontoro:
#### Summary
This PR proposes a new release process, with a fixed timeline and owners, to ensure released versions are well tested and stable. It is simple by design, we can always iterate on it when needed.

Adding everyone in the team that may be part of it as stated in the doc as reviewers.

Everything in this document is up to discussion, there's nothing set in stone, so feel free to disagree with some or even all of it :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53646

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,6 +1,6 @@
 # Release Process
 
-This document outlines the release process of the load-test tool. It is codified in [a Playbook](https://community.mattermost.com/playbooks/playbooks/95pa6a8fcp8p8xsn4frqejcide) that the team follows monthly in the Community server. If you don't have access to the playbook, please ping the team.
+This document outlines the release process of the load-test tool. It is codified in [a Playbook](https://community.mattermost.com/playbooks/playbooks/95pa6a8fcp8p8xsn4frqejcide) that the team follows monthly in the Community server. If you don't have access to the playbook, please ping `@xyzteam` in the [Developers: Server channel](https://community.mattermost.com/core/channels/developers-server).
 
 ## Requirements
 
--- a/docs/release.md
+++ b/docs/release.md
@@ -51,10 +51,28 @@ In the case that the primary Dev and QA are not available, we define Dev and QA 
 - Dev backups: @streamer45, @agnivade, @isacikgoz
 - QA backups: @agarciamontoro, @streamer45, @agnivade, @isacikgoz
 
+### Output
+
+The output of the release process is twofold:
+- A commit tagged with the version number.
+- A Github release, consisting of a bundle that contains:
+    - The main asset: a tarball named `mattermost-load-test-ng-vx.y.z-linux-amd64.tar.gz`, that contains:
+        - Binaries: `ltagent` and `ltapi`.
+        - Sample configuration files.
+        - License file.
+    - Other assets bundled by Github itself:
+        - Checksums file
+        - A copy of the source code in both `zip` and `tar.gz` formats.
+
+
+An example of this output for version v1.9.1 is:
+- Commit: https://github.com/mattermost/mattermost-load-test-ng/commit/b50140c25d495c4b12fa07d421ad584271b624ed
+- Github release: https://github.com/mattermost/mattermost-load-test-ng/releases/tag/v1.9.1
+    - Main asset: https://github.com/mattermost/mattermost-load-test-ng/releases/download/v1.9.1/mattermost-load-test-ng-v1.9.1-linux-amd64.tar.gz
 
 ## Release
 
-This section outlines the technicalities of the release itself, once the new version is tested and approved.
+This section outlines the technicalities of the release itself, once the new version is tested and approved. These steps will create both the tagged commit and the Github release.
 
 We use a combination of `Makefile` and [Goreleser](https://goreleaser.com/) to automate this process, which consists of two steps:
 1. Prepare the release and get the automatically created PR merged.
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,6 +1,6 @@
 # Release Process
 
-This document outlines the release process of the load-test tool.
+This document outlines the release process of the load-test tool. It is codified in [a Playbook](https://community.mattermost.com/playbooks/playbooks/95pa6a8fcp8p8xsn4frqejcide) that the team follows monthly in the Community server. If you don't have access to the playbook, please ping the team.
 
 ## Requirements
 
--- a/docs/release.md
+++ b/docs/release.md
@@ -17,21 +17,24 @@ As such, we base the timeline of our own release process on the following goal:
 
 > We aim to have a new release ready on T-15.
 
+There may be times when there are no changes in the load-test tool, and that's fine. In those cases, we can skip the release altogether.
+
 ## Process
 
 The goal of this release process is to have a well-tested version where bugs can be surfaced before we use it for the monthly performance comparisons and other analysis. Then, the main part of it will be the testing process, and we will skip any release whose tests did not pass before the testing period ends.
 
-- T-30 (owned by Dev): Cut release branch.
+- T-30 (owned by Dev): Judge whether we need a new release: there are significant changes since last version. If there are, cut the release branch and continue with next steps. If there are not, skip the release.
 - From T-29 to T-16 (owned by QA, assisted by Dev): Testing period.
-    - Run a performance comparison with bounded and unbounded tests, Postgres and MySQL, using the latest stable release of Mattermost as both the base and new builds. The test will pass if and only if:
+    - Run a performance comparison test with bounded and unbounded tests, Postgres and MySQL, using the latest stable release of Mattermost as both the base and new builds. This test will pass if and only if:
         - All four tests finish successfully.
         - There is no significant difference between the base and new builds (they're the same build, and thus should ideally output the same exact results).
-    - If bugs are found:
+    - If bugs are found in the load-test tool:
         - Fix them in master.
         - Cherry-pick them to the release branch.
         - Run the test again.
 - T-16 end of day (owned by Dev): judge whether the tests successfully passed and decide if the release will happen the next day.
 - T-15 (owned by Dev): If the tests passed, release a new version. If not, skip the release for this month.
+- From T-14 (owned by Dev): Run the actual performance comparison with the latest stable version of the load-test tool.
 
 ### Owners
 
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,0 +1,78 @@
+# Release Process
+
+This document outlines the release process of the load-test tool.
+
+## Requirements
+
+This release process follows [the release process of the self-hosted version of Mattermost](https://handbook.mattermost.com/operations/research-and-development/product/release-process/feature-release).
+
+The requirements for this release process are the following:
+- The monthly performance comparisons, where we compare the Mattermost version about to be released with the latest stable one, are based on the RC-1 version of the new release.
+- We need a stable tested version of the load-test tool to run the performance comparisons.
+- The RC-1 version of Mattermost is cut at around T-14[^1].
+
+[^1]: Mattermost release process specifies all the release events (feature freeze, RC-1 branch cut, final build cut...) by using the T-X notation, where X is the number of working days before the release day.
+
+As such, we base the timeline of our own release process on the following goal:
+
+> We aim to have a new release ready on T-15.
+
+## Process
+
+The goal of this release process is to have a well-tested version where bugs can be surfaced before we use it for the monthly performance comparisons and other analysis. Then, the main part of it will be the testing process, and we will skip any release whose tests did not pass before the testing period ends.
+
+- T-30 (owned by Dev): Cut release branch.
+- From T-29 to T-16 (owned by QA, assisted by Dev): Testing period.
+    - Run a performance comparison with bounded and unbounded tests, Postgres and MySQL, using the latest stable release of Mattermost as both the base and new builds. The test will pass if and only if:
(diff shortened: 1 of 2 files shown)


docs/release.md
```diff
@@ -0,0 +1,78 @@
...
+
+### Actual release
+
+With the PR already merged, you can now finish the process with the following command:
+
+```sh
+make release NEXT_VER=v1.1.1
+```
+
+`goreleaser` needs to use a Github token to create the release for you. You can define it inline with the `GITHUB_TOKEN` variable or write it to the default `~/.config/goreleaser/github_token` file. Read [`goreleaser` docs](https://goreleaser.com/scm/github/) for more info.
```
agarciamontoro, line 78: This part is just moved from `developing.md`, it now makes more sense to have it here.

docs/release.md
```diff
@@ -0,0 +1,78 @@
...
+## Requirements
+
+This release process follows [the release process of the self-hosted version of Mattermost](https://handbook.mattermost.com/operations/research-and-development/product/release-process/feature-release).
+
+The requirements for this release process are the following:
+- The monthly performance comparisons, where we compare the Mattermost version about to be released with the latest stable one, are based on the RC-1 version of the new release.
+- We need a stable tested version of the load-test tool to run the performance comparisons.
+- The RC-1 version of Mattermost is cut at around T-14[^1].
+
+[^1]: Mattermost release process specifies all the release events (feature freeze, RC-1 branch cut, final build cut...) by using the T-X notation, where X is the number of working days before the release day.
```
streamer45, line 14: Super unrelated comment but this is great. Didn't know footnotes were supported.

docs/release.md
```diff
@@ -0,0 +1,78 @@
...
+
+As such, we base the timeline of our own release process on the following goal:
+
+> We aim to have a new release ready on T-15.
```
streamer45, line 18: Historically this tool has seen periods of high and low activity hence why we never formalized on a fixed cadence.
I am thinking there's a chance we could end up skipping releases in the future due to nothing changing in the month. Not a huge deal, just something to be aware of, unless we want to bump version for the sake of it but not sure it's needed, unless of course we decide to start mapping versions with the product (e.g. `mattermost-load-test-ng` 9.0, 9.1 etc).

streamer45 (commented):
Thanks, left a few comments.

A meta thought would be whether we have considered using a playbook to guide the various stakeholders through this recurring process.

agnivade (approved):

docs/release.md
```diff
@@ -0,0 +1,78 @@
...
+- Dev: the main owner of the whole process, in charge of cutting the release branch, assisting during the testing period, judging whether the release is ready and releasing the new version.
+- QA: the main owner of the testing period, in charge of running the tests and filing the bugs.
+
+The current people in charge of these roles are:
+- Dev: @agarciamontoro
+- QA: @DHaussermann
+
+In the case that the primary Dev and QA are not available, we define Dev and QA backups:
+- Dev backups: @streamer45, @agnivade, @isacikgoz
+- QA backups: @agarciamontoro, @streamer45, @agnivade, @isacikgoz
```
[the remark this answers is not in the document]
agnivade: Sort by age!

docs/release.md
agarciamontoro: > I am thinking there's a chance we could end up skipping releases in the future due to nothing changing in the month. 

Indeed, I'll clarify that skipping releases should be expected if no changes happened.

> unless we want to bump version for the sake of it but not sure it's needed, unless of course we decide to start mapping versions with the product

I don't think that's needed, no.

docs/release.md
```diff
@@ -0,0 +1,78 @@
...
+
+As such, we base the timeline of our own release process on the following goal:
+
+> We aim to have a new release ready on T-15.
+
+## Process
+
+The goal of this release process is to have a well-tested version where bugs can be surfaced before we use it for the monthly performance comparisons and other analysis. Then, the main part of it will be the testing process, and we will skip any release whose tests did not pass before the testing period ends.
+
+- T-30 (owned by Dev): Cut release branch.
```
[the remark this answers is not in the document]
agarciamontoro: I think we should be strict, yes. Whatever we have at that point, we cut the branch. If there are no changes, we skip the release. So it will be at this point when we decide whether we'll need a release or not.

docs/release.md
```diff
@@ -0,0 +1,78 @@
...
+- From T-29 to T-16 (owned by QA, assisted by Dev): Testing period.
+    - Run a performance comparison with bounded and unbounded tests, Postgres and MySQL, using the latest stable release of Mattermost as both the base and new builds. The test will pass if and only if:
+        - All four tests finish successfully.
+        - There is no significant difference between the base and new builds (they're the same build, and thus should ideally output the same exact results).
+    - If bugs are found:
+        - Fix them in master.
+        - Cherry-pick them to the release branch.
+        - Run the test again.
+- T-16 end of day (owned by Dev): judge whether the tests successfully passed and decide if the release will happen the next day.
+- T-15 (owned by Dev): If the tests passed, release a new version. If not, skip the release for this month.
```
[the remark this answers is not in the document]
agarciamontoro: The real performance comparison will still happen, and they should block the Mattermost release if needed. What I meant here is that if we don't happen to have fixed the release *of the load-test tool* at this point (which we have tested with a dummy performance comparison), we will simply use the latest stable version for the real performance comparison, instead of trying to rush the new one in.

agarciamontoro:
> A meta thought would be whether we have considered using a playbook to guide the various stakeholders through this recurring process.

Sure! I'll create one and link it here.

agarciamontoro:
Done: https://community.mattermost.com/playbooks/playbooks/95pa6a8fcp8p8xsn4frqejcide

amyblais (approved):

docs/release.md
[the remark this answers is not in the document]
streamer45: Yes, but in case of new coverage the two things are sort of coupled. We cannot really use a previous version of the tool to test performance right?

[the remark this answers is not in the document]
agarciamontoro: Indeed. This is the problem we always face: we need to have coverage of new features, and we need to have a stable tested version of the tool to run the performance comparisons. Feature complete for Mattermost is, according [to the handbook](https://handbook.mattermost.com/operations/research-and-development/product/release-process/feature-release), "approximately 1 month prior to the release day", so T-30. If that rule is followed, and if devs add the corresponding coverage to the load-test tool *at the same time* as the feature is being developed, then we should be good.
That's a lot of *if*s, though. We could try to shorten the testing/debugging/fixing period defined here, maybe leave it in one week, so we have an additional week for finishing the coverage after feature complete in Mattermost.
Thoughts?

isacikgoz (approved):

docs/release.md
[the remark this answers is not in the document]
streamer45: I think we shouldn't be able to merge any new complex functionality in monorepo until performance tests have been conducted on the PR which means coverage becomes a requirement. I think that should be the standard process, not the opposite (merging and hoping coverage will get in time). It will inevitably mean that some features will get delayed due to this and that's in the nature of things. What we cannot do going forward (in my opinion) is shipping untested software.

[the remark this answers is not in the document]
agarciamontoro: I agree. This is something that every dev should be aware of. And while we've been seeing more and more coverage added to the tool by other teams, I'm sure we're still missing PRs that may benefit from this.
But let's assume, so we don't increase the scope of this discussion greatly, that this is already happening. For the release process being discussing here, that would mean that we *need* to release a new version (if there are indeed changes) and may end up delaying the product release if we have some unfixed bugs by T-15. I think the step in the release would be something like "communicate to the release manager that the release process will be delayed". I guess at this point we would enter fire-extinguishing mode and try to get the release as soon as possible so that we can run the comparison.
An alternative of course is to be more strict in our own PRs, and thoroughly test all of them (accepting the delay that will cause).
Thoughts?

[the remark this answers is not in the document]
streamer45: Yes, I think that's reasonable. Also if there's a specific regression in the tool itself unrelated to newly added coverage we could always consider reverting to unblock the release/testing process. I think it may depend on case by case basis but overall I am supportive of your proposal.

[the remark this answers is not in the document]
agarciamontoro: > if there's a specific regression in the tool itself unrelated to newly added coverage we could always consider reverting to unblock the release/testing process

Yeah, that was part of the "fire-extinguishing mode" in my mind :) Ok, I'll reword it tomorrow and ask for a review again. Thanks for the discussion!

docs/release.md
```diff
@@ -0,0 +1,81 @@
...
+- QA: @DHaussermann
+
+In the case that the primary Dev and QA are not available, we define Dev and QA backups:
+- Dev backups: @streamer45, @agnivade, @isacikgoz
+- QA backups: @agarciamontoro, @streamer45, @agnivade, @isacikgoz
+
+
+## Release
+
+This section outlines the technicalities of the release itself, once the new version is tested and approved.
```
streamer45, line 57: Maybe we could specify it's the binary release or bundle we are referring to here. Mostly because we are using the word *release* from different levels.

docs/release.md
```diff
@@ -0,0 +1,81 @@
+# Release Process
+
+This document outlines the release process of the load-test tool. It is codified in [a Playbook](https://community.mattermost.com/playbooks/playbooks/95pa6a8fcp8p8xsn4frqejcide) that the team follows monthly in the Community server. If you don't have access to the playbook, please ping the team.
```
streamer45, line 3: Should we specify which team to ping or do you think it's sufficiently covered by the owners below?

docs/release.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Requirements
+
+This release process follows [the release process of the self-hosted version of Mattermost](https://handbook.mattermost.com/operations/research-and-development/product/release-process/feature-release).
+
+The requirements for this release process are the following:
+- The monthly performance comparisons, where we compare the Mattermost version about to be released with the latest stable one, are based on the RC-1 version of the new release.
```
streamer45, line 10: Just some random thought at this point but have we ever considered doing an extra comparison for ESRs to be compared with the previous ESRs?

docs/release.md
```diff
@@ -0,0 +1,81 @@
...
+- We need a stable tested version of the load-test tool to run the performance comparisons.
+- The RC-1 version of Mattermost is cut at around T-14[^1].
```
streamer45, line 12: I assume this is implicitly understood by everyone here but just want to highlight that there's always the case RC-1 won't be enough if we found a performance regression that either needs fixing or reverting.

streamer45 (commented):
Great stuff, left a couple more comments.

One extra piece of feedback that comes out of some of the discussion above is that maybe there's also an opportunity to improve our official release process in the handbook if we can find an agreement between the various stakeholders that a successful load-test comparison step becomes a strict requirement to ship any new MM version. I think it could be a great step towards more awareness on the importance of performance.

docs/release.md
agarciamontoro: Do you mean the ltagent/ltapi vs ltctl? At some point we should release ltctl as well I think, otherwise it'll be confusing.

docs/release.md
agarciamontoro: Ah, yes, I'll specify it!

docs/release.md
agarciamontoro: That's something that's been on my mind for a while, but we've never planned anything related. We probably should at some point.

amyblais: I agree with the idea of adding loadtesting to compare ESRs.

docs/release.md
streamer45: I mean the Github release bundle, whatever that may contain.

docs/release.md
streamer45: > Do you mean that the finally released version may be RC-2 or higher?

Yeah, and it may imply having to do a second comparison as well.

agarciamontoro: I was gonna clarify this in the document, but on second thought, that would be part of the performance comparison process, not part of the release process. Once the load-test version is released, we will use that one to run the performance comparison as many times as needed on RC-1and following. If we ever need to revert a feature from the load-test tool after RC-1, we can plan that ad-hoc.

docs/release.md
agarciamontoro: https://github.com/mattermost/mattermost-load-test-ng/pull/640/commits/666a4a3d75a1a1b55a6345642adbe4d28892545c

docs/release.md
agarciamontoro: https://github.com/mattermost/mattermost-load-test-ng/pull/640/commits/9d18f1b7f1f03a463e4e599dad9c8d23229b98d4

agarciamontoro:
> A note on Claudio's comment for: "maybe there's also an opportunity to improve our official release process in the handbook if we can find an agreement between the various stakeholders that a successful load-test comparison step becomes a strict requirement to ship any new MM version."
> I agree with this as well and I'm happy to update the handbook if there is agreement from other dev teams as well.

I'll bring this up in the dev meeting today and see if there's anything against it.

agarciamontoro:
I'm merging this PR with the hope of trying the release process synchronized with the release of v9.2, which starts tomorrow. The dates will be as follow:

- T-24 (October 13): Decide if a new release is worth happening.
- T-23 to T-12 (October 16 to October 31): Testing period
- T-11 (November 1): Release new version
- T-10 to T-8 (November 2 to November 6): Performance comparison for v9.2

PR #643, opened 2023-08-22
[MM-54213] Add private IP to allowed list only when found

streamer45:
#### Summary

In https://github.com/mattermost/mattermost-load-test-ng/pull/624 I assumed that the instance deploying the cluster always had a private IP. But when ran locally that may not always be the case which can result in an empty IP and resulting in error. So we are adding the private IP to the allowed list only if we can find one.  

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54213


agnivade (approved):

agarciamontoro (approved):
Thank you!

PR #644, opened 2023-08-25
MM-54251: Let the user hardcode the name of the database

agarciamontoro:
#### Summary
When using an existing DB cluster via `ClusterIdentifier`, a database is already present, and its name will be different from the one assumed in the code. The first error encountered because of this happened when trying to alter the database to set the default text search config to `english` instead of `simple` (see the ticket), but we can expect more problems of this broken assumption.

To solve this problem, I've added a new `DBName` parameter that the user can set to hardcode the name of the DB:
- If `ClusterIdentifier` is set, it should be the name of the database in such cluster
- If `ClusterIdentifier` is not set, its value can be whatever, and when the new database is created, instead of calling it `${ClusterName}db`, it'll be simply called `${DBName}`.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54251

### DIFF
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -115,6 +115,10 @@ type TerraformDBSettings struct {
 	DBParameters DBParameters
 	// ClusterIdentifier indicates to point to an existing cluster
 	ClusterIdentifier string `default:""`
+	// DBName specifies the name of the database.
+	// If ClusterIdentifier is not empty, DBName should be set to the name of the database in such cluster.
+	// If ClusterIdentifier is empty, the database created will use DBName as its name.
+	DBName string `default:""`
 }
 
 // ExternalDBSettings contains the necessary data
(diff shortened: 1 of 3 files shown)


agarciamontoro:
@agnivade That's what I thought in the first place (and my initial solution was just to skip this), but I manually checked the value and it was set to `simple`. Aurora uses that as a default, so maybe it doesn't carry it over in the backups :shrug:

agnivade:
Oh that's interesting. :thinking:

agnivade (approved):

deployment/config.go
```diff
@@ -115,6 +115,10 @@ type TerraformDBSettings struct {
 	DBParameters DBParameters
 	// ClusterIdentifier indicates to point to an existing cluster
 	ClusterIdentifier string `default:""`
+	// DBName specifies the name of the database.
+	// If ClusterIdentifier is not empty, DBName should be set to the name of the database in such cluster.
```
streamer45, line 119: Should we enforce this rule during validation somehow to prevent more problems?

agarciamontoro: We could validate that it's not empty, but not that it's set to the correct value. Is that what you mean?

streamer45: Yes, if we require it to be set to something we should at least check that or fail early.

streamer45 (approved):
:100: , appreciate the extra test!

PR #646, opened 2023-08-29
[MM-54315] Allow overriding user prefix on automated deployments

streamer45:
#### Summary

PR introduces a new `UserPrefix` setting to allow overriding the generated username prefix used on automated deployments. If none is provided (empty string), we keep generating it based on the agent's id.

This shouldn't change the existing semantic for the manual use case where we still default to `testuser` and can always override through the `--user-prefix` command line flag.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54315

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -57,7 +57,12 @@ func (a *api) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		})
 		return
 	}
+
+	agentId := r.FormValue("id")
 	ltConfig := data.LoadTestConfig
+	if ltConfig.UsersConfiguration.UserPrefix == "" {
+		ltConfig.UsersConfiguration.UserPrefix = agentId
+	}
 	if err := defaults.Validate(ltConfig); err != nil {
 		writeAgentResponse(w, http.StatusBadRequest, &client.AgentResponse{
 			Error: fmt.Sprintf("could not validate config: %s", err),
@@ -98,7 +103,6 @@ func (a *api) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	agentId := r.FormValue("id")
 	if val, ok := a.getResource(agentId); ok && val != nil {
 		if _, ok := val.(*loadtest.LoadTester); ok {
 			writeAgentResponse(w, http.StatusConflict, &client.AgentResponse{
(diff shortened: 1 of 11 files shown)


agnivade (approved):

agarciamontoro (changes requested):
I've been testing this and saw two problems:
1. I've ran a load-test with 6000 users connected starting with a database that contained 3000 users (3005 counting bots and sysadmin). When the test reached the 6000 users (verified via the number connected devices panel in Grafana), I ran a `SELECT count(*) FROM Users` and still got 3005. I don't know if this is a new problem.
2. Trying to debug the above problem, I set the `UserPrefix` setting in deployer.json to the empty string `""`. Then I got this error:
```
Error: failed to create coordinator: coordinator: coordinator api request error: could not validate load-test config: UserPrefix is empty
```
I think this is caused by [this validation](https://github.com/mattermost/mattermost-load-test-ng/blob/a3e30c0359576f86ce8b041fd41b5999343433e7/api/coordinator.go#L61-L62), where we don't update the agent's config, but I haven't checked.

Problem 2. is definitely introduced by this PR, but I'm not sure if 1. was as well. I'm still on it.

agarciamontoro:
Another note on 1.: Running a new test from current `master` does create new users. Running the following queries when there wer around 900 users already connected (the already existing ones had `testuser` as prefix):
```
agnivaltdb=> select count(*) from users;
 count
-------
  3892
(1 row)

agnivaltdb=> select count(*) from users where username not like '%testuser%';
 count
-------
   894
(1 row)
```
It seems this is a problem introduced by this PR, right?

streamer45:
@agarciamontoro Good catch. I admit I only tested with one agent so didn't notice this. I guess we could just override the cluster name portion and call it a day, does that sound good?

agarciamontoro:
> I guess we could just override the cluster name portion and call it a day, does that sound good?

Do you mean leaving the `agent-%d` part? So that we get `overridenname-agent-0-76` and `overridenname-agent-1-76`? That would work only in the cases where the number of agents in the current cluster equals the number of agents that were used to generate database. However, that wouldn't work with our current database at all, since all existing users are of the form `testuser-%d`.

agarciamontoro:
We could make each agent start with the userId that matches their own agent number, and make them increment the id of the user each time according to the number of agents. Something like:
- agent-0 runs users 0, 3, 6, 9...
- agent-1 runs users 1, 4, 7, 10...
- agent-2 runs users 2, 5, 8, 11...

streamer45:
Alright, I guess we could expose `--user-offset` as well and use that accordingly based on the available number of agents.

streamer45:
Still there's the complicacy of the agent id which in our case is not wanted but in the other case (users generated through a previous load-test) would be required.

streamer45:
Closing in favor of https://github.com/mattermost/mattermost-load-test-ng/pull/647

PR #647, opened 2023-08-30
MM-54315: Expand UsersFilePath usage

agarciamontoro:
#### Summary
This PR tries to fix the problem of the deployments not using the set of users present in the database dump. Instead of trying to change the user prefix so we can adapt it to the users from the dump, we leverage the already existing `UsersFilePath` setting in `config/config.json`, and expand its usage by:
- Adding a new `UsersFilePath` setting to `config/deployer.json` so that it overrides the data in the deployed agents.
- Splitting the contents of such file into $n$ chunks, where $n$ is the number of agents, and uploading a different chunk to each agent.
- Removing the limitation for mixing user credentials defined in `UsersFilePath` with automatic sign-ups of new users, so that the agent simply switches to automatic sign-ups when there are no more credentials to be consumed from the file, creating new users when needed.

This is an alternative to #646, so we should merge only one.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54315

### DIFF
--- a/deployment/terraform/agent.go
+++ b/deployment/terraform/agent.go
@@ -15,6 +15,8 @@ import (
 	"github.com/mattermost/mattermost/server/public/shared/mlog"
 )
 
+const dstUsersFilePath = "/home/ubuntu/users.txt"
+
 func (t *Terraform) generateLoadtestAgentConfig() (*loadtest.Config, error) {
 	cfg, err := loadtest.ReadConfig("")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


agnivade (approved):
This looks good. It's a bit more work, but definitely handles all cases and more flexible.

deployment/terraform/agent.go
```diff
@@ -58,7 +65,29 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent) error {
 		commands = append([]string{"wget -O tmp.tar.gz " + t.config.LoadTestDownloadURL}, commands...)
 	}
 
-	for _, val := range t.output.Agents {
+	// If UsersFilePath is present, split the user credentials among all the agents,
+	// so that the logged in users don't clash
+	splitFiles := make([][]string, 0, len(t.output.Agents))
+	if t.config.UsersFilePath != "" {
```
streamer45, line 71: I wish we didn't have to resort to this. Ideally we upload the same file to all agents and make them aware of their own index (should already be doable with some regex) so that they can simply seek to the needed offset $\frac{nUsers}{nAgents} \cdot agentIdx$.

streamer45 (approved):
Thanks, definitely more than good enough for our current needs!

deployment/terraform/agent.go
agarciamontoro: Yeah, that's a good point. I'm going to merge this one as is, though, as it's already tested, and tackle your suggestion in https://mattermost.atlassian.net/browse/MM-54436

PR #649, opened 2023-09-01
MM-54314: Return the error correctly from config parse failure

agnivade:
We were trying to set the defaults after parsing. But we were ignoring the errors from the parse. To solve this, we set the defaults first and then override it while unmarshaling the JSON. This achieves the same result but also returns errors correctly.

https://mattermost.atlassian.net/browse/MM-54314

### DIFF
--- a/defaults/json.go
+++ b/defaults/json.go
@@ -10,6 +10,10 @@ import (
 // This function will try to read from given path, if it is empty will try
 // fallback path. If it fails on fallback, it will set value to it's defaults
 func ReadFromJSON(path, fallbackPath string, value interface{}) error {
+	if err := Set(value); err != nil {
+		return err
+	}
+
 	if path != "" {
 		if err := readJSON(path, &value); err != nil {
 			return fmt.Errorf("failed to read from path %s: %w", path, err)
@@ -17,9 +21,6 @@ func ReadFromJSON(path, fallbackPath string, value interface{}) error {
 		return nil
 	}
 
-	if err := Set(value); err != nil {
-		return err
-	}
 
 	// If the fallback path doesn't exist, return.
 	if _, err := os.Stat(fallbackPath); err != nil && os.IsNotExist(err) {
--- a/defaults/json.go
+++ b/defaults/json.go
@@ -12,13 +12,13 @@ import (
 func ReadFromJSON(path, fallbackPath string, value interface{}) error {
 	if path != "" {
 		if err := readJSON(path, &value); err != nil {
-			return err
+			return fmt.Errorf("failed to read from path %s: %w", path, err)
 		}
 		return nil
 	}
 
-	if err := readJSON(fallbackPath, value); err == nil {
-		return nil
+	if err := readJSON(fallbackPath, value); err != nil {
+		return fmt.Errorf("failed to read from path %s: %w", fallbackPath, err)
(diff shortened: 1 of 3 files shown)


streamer45 (approved):
Thanks, tested and working nicely :+1:

agarciamontoro (changes requested):
We seem to test only the case where `path` is empty, and the logic for when that argument is not empty is actually different. I'm looking at the code and most (all?) of the calls pass the value of `--config` or `--comparison-config` flags as `path`, whereas `fallbackPath` is the usual config file relative to the working directory, `./config/whatever`. Should we make the behaviour the same for both? I don't know why it's actually different.
If there's a reason to keep the two paths, we should test them both.

agnivade:
ping @agarciamontoro

agarciamontoro:
> And if fallback path exists, then the default values are overriden by the values defined in fallback path.

@agnivade My concern is that we behave differently for `path` and `fallbackpath`, and we're only testing one of the code paths (pun not intended).

agnivade:
From what I know, that's intentional. @streamer45 - any thoughts on the above?

streamer45:
> From what I know, that's intentional. @streamer45 - any thoughts on the above?

I guess you slightly changed the order of operations there. Are we confident that calling `Set()` before attempting to parse from the fallback path isn't going to inadvertently touch fields that may be missing from the JSON file? I'd probably try to keep the sequencing untouched if it's not too complex.

1. Primary path
    - Success: return nil
    - Fail: return error
2. Fallback path
    - Success: return nil
    - Fail: ignore error (Is this case what you weren't totally happy with @agarciamontoro ?)
      - Maybe we could log a global warning just to improve a bit the potential debuggability.
3. Set defaults
    - Success: return nil
    - Fail: return error

agnivade:
> Are we confident that calling Set() before attempting to parse from the fallback path isn't going to inadvertently touch fields that may be missing from the JSON file? 

I can do that, but I think some tests were failing because of this. If we want to change the functionality, I'm fine with that.

agarciamontoro:
> Is this case what you weren't totally happy with @agarciamontoro ?)

I'm simply confused with the fact that we behave differently for each of the paths. So I'd love to either:
1. Behave equally for all paths.
2. Maintain the difference in logic, but test all paths.

agnivade:
Ok, I re-arranged the default setting. Now the default setting happens always, and then all paths are handled equally.

defaults/json.go
```diff
@@ -10,19 +10,24 @@ import (
...
-	if err := readJSON(fallbackPath, value); err == nil {
+	// If the fallback path doesn't exist, return.
+	if _, err := os.Stat(fallbackPath); err != nil && os.IsNotExist(err) {
 		return nil
 	}
 
-	if err := Set(value); err != nil {
-		return err
+	if err := readJSON(fallbackPath, value); err != nil {
+		return fmt.Errorf("failed to read from path %s: %w", fallbackPath, err)
```
agarciamontoro, line 30: Nit to make the error slightly more useful
```suggestion
		return fmt.Errorf("failed to read from fallback path %s: %w", fallbackPath, err)
```

agarciamontoro (approved):
Much simpler now, thanks!

PR #650, opened 2023-09-05
MM-53357: Check MaxUsersPerTeam setting

agnivade:
We verify this setting and fail earlier rather than
suddenly running into this limitation and failing the load test
at a later stage.

https://mattermost.atlassian.net/browse/MM-53357

### DIFF
--- a/coordinator/cluster/config.go
+++ b/coordinator/cluster/config.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/store/memstore"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/user/userentity"
 )
 
 // LoadAgentConfig holds information about the load-test agent instance.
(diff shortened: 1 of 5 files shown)


coordinator/cluster/config.go
```diff
@@ -33,5 +35,46 @@ func (c *LoadAgentClusterConfig) IsValid(ltConfig loadtest.Config) error {
 	if ltConfig.UsersConfiguration.MaxActiveUsers*len(c.Agents) < c.MaxActiveUsers {
 		return errors.New("coordinator: total MaxActiveUsers in loadTest should not be less than clusterConfig.MaxActiveUsers")
 	}
+
+	ok, err := checkMaxUsersPerTeam(ltConfig, *c)
```
agnivade, line 39: This is the natural place where I thought it would fit best. Because this is also where `ltConfig.UsersConfiguration.MaxActiveUsers*len(c.Agents)`  gets checked. Obviously, this check doesn't happen in a simple loadtest via the ltagent, but it seemed cumbersome to pass the `ClusterConfig` there. 

The problem here is that one needs to manually create the user and make the API calls. But even in the loadtest package, there is no admin user, and this would have been needed to be done separately.

LMK what you think.

agnivade:
Mainly for customers who test on an existing env. I believe a certain large customer ran into this. @coltoneshaw can give more context.

coltoneshaw:
@agnivade / @streamer45 - Context as Agniva is right. We were loadtesting that customer and kept hitting an issue whenever they would reset their environment and start a loadtest. It wasn't immediately apparent of the issue, especially when running the loadtest with all its moving parts. So, what we had to do is just take note that once the environment spins up (after the `deployment create`) someone has to log in, adjust the `maxUsersPerTeam` to a much higher value, then kickoff the loadtest.

coltoneshaw:
Additional note: This only seems to be an issue when you are loadtesting with an existing db but letting the loadtest create the app nodes. So, the config already exists in the database but the app nodes are new.

streamer45:
Okay, I now wonder whether we should have a dedicated command/subcommand to validate all these things because we are now coupling remote config validation with the local one and it doesn't seem like the cleanest of approaches.

agarciamontoro (commented):
I have a similar comment to Claudio's. I think the place is a good one to check whether it's valid, but it also feels weird to me to construct the sysadmin using the `userentity` there. Maybe we could have a sysadmin client as part of the `Coordinator`, and pass whatever information we need to `LoadAgentClusterConfig.IsValid`? This is me just throwing ideas here, I don't have a super strong opinion on how to do it, but maybe that's a bit more reusable.

agnivade:
@streamer45 - Do you have any preference on where this code should live?

streamer45:
I think the current location makes sense for this specific check although I am still wondering what happens if in the future we want to extend this server config validation logic to other required settings (e.g. checking that `EnableOpenServer` is on). So far we've deferred this sort of requirements to the documentation.

As far as implementation goes, maybe I'd go with a simple `model.Client4` rather than "abusing" a `userentity` but I am not even sure if that's going to look any prettier so I'll leave it to you.

agnivade:
Thanks, I'll see what I can do.

agnivade:
Closing this. Will revisit later.

PR #653, opened 2023-09-29
MM-46253: Comparison collect command

agarciamontoro:
#### Summary
Just some quality-of-life improvement that I've been wanting to do forever, and today I decided to throw some code at it. I just tested with the current comparison I'm running and it works.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-46253

### DIFF
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/comparison"
+	"github.com/wiggin77/merror"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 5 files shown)


agnivade (approved):
Nice, possible to add any tests for this?

cmd/ltctl/comparison.go
```diff
@@ -172,6 +173,33 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to initialize comparison object: %w", err)
+	}
+
+	merr := merror.New()
+	for _, id := range cmp.GetDeploymentIds() {
+		if err := collect(deployerConfig, id, id+"_"); err != nil {
+			merr.Append(err)
+		}
+	}
```
streamer45, line 198: Is there an opportunity here to use some parallelization?

streamer45 (approved):
Nice improvement :+1:

cmd/ltctl/comparison.go
agarciamontoro: There is, but I'd rather keep this as simple as possible, as the potential gain in this would be in the realm of a few seconds, which doesn't seem worth it at the moment. I'm 1/5 though, if you have a stronger opinion I can parallelize it.

agarciamontoro:
> Nice, possible to add any tests for this?

Not trivial to add tests to the current collection-related functions, since they connect to the remote instances and save the files. Doable, but I think it wouldn't add much value, since we would need to mock a lot of things.

I've added a test of the new function to get the deployment IDs.

agarciamontoro:
Well, it's testing that it is indeed sorted, which is important for consistent results.

cmd/ltctl/comparison.go
streamer45: I asked because I vaguely remember this taking more than a few seconds and multiplied in a big deployment could become several minutes. But it's the sort of improvement we can do at a later point when someone will feel the pain of waiting :p

streamer45 (approved):

PR #654, opened 2023-10-02
[MM-54702] Automatically clear licenses data from DB dumps

streamer45:
#### Summary

PR adds an extra step to the db dump loading command that clears existing license related information. This should be the last piece needed to fix the loading issues.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-54702

### DIFF
--- a/deployment/utils.go
+++ b/deployment/utils.go
@@ -54,25 +54,37 @@ func ProvisionURL(client *ssh.Client, url, filename string) error {
 	return nil
 }
 
-// BuildLoadDBDumpCmds returns a slice of commands that, when piped, feed the
+// BuildLoadDBDumpCmd returns a command string to feed the
 // provided DB dump file into the database. Example:
 //
-//	zcat dbdump.sql
-//	mysql/psql connection_details
-func BuildLoadDBDumpCmds(dumpFilename string, dbInfo DBSettings) ([]string, error) {
-	cmds := []string{
+//	zcat dbdump.sql | mysql/psql connection_details && custom queries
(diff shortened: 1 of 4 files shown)


deployment/utils.go
```diff
@@ -54,25 +54,37 @@ func ProvisionURL(client *ssh.Client, url, filename string) error {
...
+//	zcat dbdump.sql | mysql/psql connection_details && custom queries
+func BuildLoadDBDumpCmd(dumpFilename string, dbInfo DBSettings) (string, error) {
+	loadCmds := []string{
 		fmt.Sprintf("zcat %s", dumpFilename),
 	}
 
-	var dbCmd string
+	var dbConnCmd string
+	var licenseClearCmd string
+	licenseClearQuery := "DELETE FROM Systems WHERE Name = 'ActiveLicenseId'; DELETE FROM Licenses;"
```
agnivade, line 68: Can we not keep this as a separate command after the load is done? This feels like overcomplicating the command structure like `zcat <> | psql <> && psql <>`.

streamer45: You either complicate the command or you complicate the code. But the command generation is automated so it didn't feel that complex to me. If we split them then we need to once more check the DB type and add more clutter. It seemed natural to keep them coupled since they all need to happen in sequence and it makes it easy to add more in the future if needed as there's a single place to take care of this.

agnivade (approved):

agarciamontoro (approved):
Yay, thanks for this!

PR #655, opened 2023-10-16
MM-54959: Run collect commands inside instances as root

agarciamontoro:
#### Summary
The output of `go run ./cmd/ltctl collect` fails when collecting the output of `dmesg`:
```
failed to run cmd "dmesg > /tmp/dmesg.log": Process exited with status 1
```
Manually sshing to the instance and running the command surfaces the permission error that’s preventing the command to successfully finish:

```
ubuntu@ip-172-27-133-234:~$ dmesg > /tmp/dmesg.log
dmesg: read kernel buffer failed: Operation not permitted
```

When running `sudo dmesg > /tmp/dmesg.log`, it finishes successfully.

The current code is a bit convoluted, accommodating both file paths and commands in a single argument of the `addInfo` function. I've taken the quickest solution in this PR to minimize engineering time: adding `sudo` to all commands (which is only `dmesg` as of today), since adding it to the `addInfo` call does not work because it would create the command `sudo dmesg > /tmp/sudo dmesg.log`.

But I'm open to spend a bit more time and refactor that code. Thoughts?

EDIT: I've now implemented the sane solution, see [my comment below.
](https://github.com/mattermost/mattermost-load-test-ng/pull/655#issuecomment-2069114786)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54959


agnivade (changes requested):
Let's spend the time needed to refactor and properly differentiate between commands and files. And then pass the `sudo` bit from commands.

For now, we can just skip the `dmesg` collection. I am not even running this step atm.

agarciamontoro:
Refactored the collection of files and commands into two different functions, and cleaned the code a bit. Tested on a real deployment and it works.

isacikgoz (approved):
Nice, and I liked the way you refactor it.

PR #656, opened 2023-10-17
Fix the dates in the release process

agarciamontoro:
#### Summary
I had somehow understood that RC1 was cut on T-10, but that was wrong: T-10 is code freeze, RC1 is actually cut on T-19.

As the availability of the Mattermost RC1 version was the key date from which the whole timeline was planned, I had to refactor it by:
- Reducing the testing period from 12 days to 5 days.
- Postponing the release to the day *after* RC1 is cut, instead of the day *before*.

I was pretty comfortable with the 12 days we gave to testing (and potential bug fixing), and had considered it maybe too much, since the testing we have planned should be quite short. However, reducing it to 5 days may be too much. This will be informed by the test run of the process we're executing right now, but I want to hear your thoughts (specially @DHaussermann's).

#### Ticket Link
--

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -9,13 +9,13 @@ This release process follows [the release process of the self-hosted version of 
 The requirements for this release process are the following:
 - The monthly performance comparisons, where we compare the Mattermost version about to be released with the latest stable one, are based on the RC-1 version of the new release.
 - We need a stable tested version of the load-test tool to run the performance comparisons.
-- The final release build of Mattermost is cut at around T-10[^1].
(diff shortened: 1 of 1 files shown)


agnivade (approved):
All of this `T-` makes me feel like this is a rocket launch manual :joy:

agarciamontoro:
That's the goal :rocket:

DHaussermann (approved):
@agarciamontoro this makes sense to me. It should give us a large enough buffer before code freeze.

I would proceed 👍

agarciamontoro:
Thanks, Dylan!

PR #658, opened 2023-10-30
MM-54986: Disable app and/or agent profiling

agarciamontoro:
#### Summary
Motivated for the huge amount of memory consumed by profiling 50 agents in the ceiling tests, we decided to allow the user to selectively disable that type of profiling. This PR solves that by making it optional to profile both the app and the agent nodes.

~~Note that the approach to test the generated yaml is a bit esoteric, but parsing the generated yaml seemed even more convoluted than this. I'm open to changing it if your eyes bleed a lot when reading the test, though.~~ Ended up using go-yaml to generate the config, much easier and cleaner.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-54986

### DIFF
--- a/config/deployer.sample.json
+++ b/config/deployer.sample.json
@@ -111,5 +111,9 @@
         "Query": "histogram_quantile(0.99, sum(rate(mattermost_api_time_bucket[1m])) by (le))"
       }
     ]
+  },
+  "PyroscopeSettings": {
+    "EnableAppProfiling": true,
+    "EnableAgentProfiling": true
   }
 }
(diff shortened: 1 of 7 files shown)


streamer45:
One concern is that without tweaking the template we'd still be registering an application but with no targets. I am wondering if that could potentially get confusing or even cause issues.

isacikgoz (approved):
LGTM, feel free to ask for another review if what Claudio mentioned causes any issues.

agarciamontoro:
Ended up using go-yaml not to parse the generated string, but to generate the yaml from scratch. I'm testing it right now in a deployment to make sure that the generated config is valid for Pyroscope.

streamer45 (approved):
Looks so much better, I appreciate the time taken to rework this :1st_place_medal:

agnivade (approved):
Nice one

PR #659, opened 2023-10-31
MM-55019: Optimize storage size and make it configurable

agarciamontoro:
#### Summary
This PR tackles the following issues:
- The type of storage was `gp2`, but `gp3` is more performant and cheaper.
- The size of the storage was not configurable at all, since it was only exposed as a Terraform variable, but not as a user-provided setting.
- Even if one would go and manually change the Terraform variable, the change would affect all instance types, since all of them shared the same variable to set their storage size.

A couple of additional notes:
- The [current AMI](https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#ImageDetails:imageId=ami-003d3d03cfe1b0468) uses an 8GiB snapshot, so that's the minimum size we can set. I rounded it up to 10GiB just in case (we had planned to use 5GiB for app, agents and proxy, but it's too little)
- The old ElasticSearch instance type was not compatible with `gp3`, so I updated it to the one I tested, but we will probably change it again when we investigate ElasticSearch more thoroughly in https://mattermost.atlassian.net/browse/MM-52337.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55019


config/deployer.sample.json
```diff
@@ -111,5 +111,13 @@
         "Query": "histogram_quantile(0.99, sum(rate(mattermost_api_time_bucket[1m])) by (le))"
       }
     ]
-  }
+  },
+  "StorageSize": {
```
streamer45, line 115: Should this be `StorageSizes`? Not a big deal, just a nit.

streamer45 (approved):
Nice improvement :+1:

agnivade (approved):

config/deployer.sample.json
agarciamontoro: Yeah, that makes more sense, thanks! Done in https://github.com/mattermost/mattermost-load-test-ng/pull/659/commits/b931707620510de25362c09c5a1b62f017362d28

PR #662, opened 2023-11-07
MM-55312: Bump MaxNotificationsPerChannel

agnivade:
By default, this number was 1000 which didn't allow
typing events to be sent for channels with larger
member counts.

https://mattermost.atlassian.net/browse/MM-55312


agnivade:
Oh wow, the results quite drastic. I can see a 44% decrease in number of supported users just by unlocking this. :) I guess our calculations are off by a lot.

PR #663, opened 2023-11-08
MM-54455: Fix CORS issue in load test deployment

agnivade:
Since we are using ltserver as the site URL, the ws connection
fails when connecting via browser.

https://mattermost.atlassian.net/browse/MM-54455

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -539,6 +539,8 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
 	cfg.ServiceSettings.EnableLinkPreviews = model.NewBool(true)
 	cfg.ServiceSettings.EnablePermalinkPreviews = model.NewBool(true)
 	cfg.ServiceSettings.PostPriority = model.NewBool(true)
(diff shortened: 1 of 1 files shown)


deployment/terraform/create.go
```diff
@@ -539,6 +539,7 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
 	cfg.ServiceSettings.EnableLinkPreviews = model.NewBool(true)
 	cfg.ServiceSettings.EnablePermalinkPreviews = model.NewBool(true)
 	cfg.ServiceSettings.PostPriority = model.NewBool(true)
+	cfg.ServiceSettings.AllowCorsFrom = model.NewString("*")
```
streamer45, line 542: This obviously works and it's a quick fix. I suppose the more elaborate and security friendly way would be to add a URL using public dns/ip from either the proxy or the first app instance (if running a single app deployment). A potential further improvement I suppose.

For now, would it make sense to leave a comment explaining why we are setting this?

streamer45 (approved):

deployment/terraform/create.go
agnivade: Yep, it's a quick fix of course. A comment sounds fine.

PR #664, opened 2023-11-13
MM-53920: Add new coordinator metrics

agarciamontoro:
#### Summary

This PR adds four more queries to the default coordinator metrics. I thought this was going to be straightforward, but the nuances of each of the `node_*` metrics and the complexity of analyzing when a network is congested or not made this a bit more difficult. So a thorough explanation follows, sorry for the long wall of text.

##### CPU utilization
We compute the percentage of CPU used, by subtracting the percentage of time in idle mode (`node_cpu_seconds_total` is already a percentage) to the total:

```
100 - 100 * (avg(irate(node_cpu_seconds_total{instance=~"app.*",mode="idle"}[5m])))
```

This one is clear, already tested in the first version of the ceiling tests, and it does not need much more discussion.

##### Memory utilization

Similarly, we compute the percentage of memory used, by substracting the percentage of available memory (computed as MemAvailable / MemTotal) to the total:

```
100 - 100 * avg(
    node_memory_MemAvailable_bytes{instance=~"app.*"}
    /
    node_memory_MemTotal_bytes{instance=~\"app.*\"
)
```

The goal for this metric is clear as well, but it needs some clarification, since the `node_memory_*` metrics are a bit confusing. This could be computed as follows as well, computing the available memory as MemFree + Cached + Buffers:

```
100 - 100 * avg(
    (node_memory_MemFree_bytes{instance=~"app.*"} + node_memory_Cached_bytes{instance=~"app.*"} + node_memory_Buffers_bytes{instance=~"app.*"})
    /
    node_memory_MemTotal_bytes{instance=~"app.*"}
)
```

The latter is suggested in many places as the golden rule, but I've chosen the former for two reasons:
1. It seems `MemAvailable` is the correct metric to look at: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773
2. Using `MemAvailable` is more conservative than using `MemFree + Cached + Buffers`, as in: the estimated percentage of used memory is slightly higher when using `MemAvailable`. See the comparison:
image

##### TCP metrics

The two other metrics are designed to keep an eye on the potential congestion of the network. These two may need a bit more discussion.

###### Segments retransmitted

One of the best metrics I've found that may point to unreliable networks is the rate of segments retransmitted versus segments sent out (stolen from Brendan Gregg's Systems Performance book). The query is straightforward:

```
100 * avg(rate(node_netstat_Tcp_RetransSegs{instance=~"app.*"}[1m]) / rate(node_netstat_Tcp_OutSegs{instance=~"app.*"}[1m]))
```

This percentage is very very low in the environments I've tested it, unless there is a weird spike, as shown in the two following examples of two different tests (~15K and ~18K users, respectively):

image
image

Looking at these two examples, I've set the threshold to 0.1%

###### Listen drops

As discussed in the thread in Community, we were also interested in the total number of listen drops (which includes listen overflows as well). Again, the query is straightforward:

```
avg(rate(node_netstat_TcpExt_ListenDrops{instance=~"app.*"}[1m]))
```

And looking at its values in the same environments as before, I've decide to set the threshold at 20. I'm not too convinced this is right, or if it may be too low, so I'm open to suggestions on how to decide its value:

image
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-53920

### DIFF
--- a/config/coordinator.sample.json
+++ b/config/coordinator.sample.json
@@ -46,6 +46,34 @@
         "Threshold": 0.025,
         "MinIntervalSec": 60,
         "Alert": true
+      },
+      {
+        "Description": "CPU utilization",
+        "Query": "100 - 100 * (avg(irate(node_cpu_seconds_total{instance=~\"app.*\",mode=\"idle\"}[5m])))",
+        "Threshold": 85,
+        "MinIntervalSec": 60,
+        "Alert": true
+      },
+      {
+        "Description": "Memory utilization",
+        "Query": "100 - 100 * avg(node_memory_MemAvailable_bytes{instance=~\"app.*\"} / node_memory_MemTotal_bytes{instance=~\"app.*\"})",
+        "Threshold": 85,
+        "MinIntervalSec": 60,
+        "Alert": true
+      },
+      {
+        "Description": "Percentage of TCP segments retransmitted",
+        "Query": "100 * avg(rate(node_netstat_Tcp_RetransSegs{instance=~\"app.*\"}[1m]) / rate(node_netstat_Tcp_OutSegs{instance=~\"app.*\"}[1m]))",
+        "Threshold": 0.1,
+        "MinIntervalSec": 60,
+        "Alert": true
+      },
+      {
+        "Description": "Total TCP listen drops",
+        "Query": "avg(rate(node_netstat_TcpExt_ListenDrops{instance=~\"app.*\"}[1m]))",
+        "Threshold": 20,
+        "MinIntervalSec": 60,
+        "Alert": true
       }
     ]
   },
--- a/config/coordinator.sample.json
+++ b/config/coordinator.sample.json
@@ -62,16 +62,17 @@
         "Alert": true
       },
       {
-        "Description": "Percentage of TCP segments retransmitted",
-        "Query": "100 * avg(rate(node_netstat_Tcp_RetransSegs{instance=~\"app.*\"}[1m]) / rate(node_netstat_Tcp_OutSegs{instance=~\"app.*\"}[1m]))",
-        "Threshold": 0.1,
+        "Description": "Percentage of TCP timeouts in the proxy node",
+        "Query": "100 * rate(node_netstat_TcpExt_TCPTimeouts{instance=~\"proxy:9100\"}[1m]) / (rate(node_netstat_TcpExt_TCPTimeouts{instance=~\"proxy:9100\"}[1m]) + node_netstat_Tcp_CurrEstab{instance=~\"proxy:9100\"})",
+        "Threshold": 0.2,
         "MinIntervalSec": 60,
         "Alert": true
       },
+
       {
-        "Description": "Total TCP listen drops",
-        "Query": "avg(rate(node_netstat_TcpExt_ListenDrops{instance=~\"app.*\"}[1m]))",
-        "Threshold": 20,
+        "Description": "Percentage of TCP timeouts in the app nodes",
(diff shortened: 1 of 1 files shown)


agnivade:
Really good work. Yes, calculating memory utilization in Linux is a trap. Nice work doing an exhaustive analysis on that. Might be worth sharing in the dev meeting ;)

Just one final question: can we run an unbounded test with these new metrics on, say 2.1a, and see if we get reduced numbers? And if so, by how much. Maybe we should first get a baseline number with the typing event fix included.

agarciamontoro:
> Just one final question: can we run an unbounded test with these new metrics on, say 2.1a, and see if we get reduced numbers? And if so, by how much. Maybe we should first get a baseline number with the typing event fix included.

Is the following plan what you mean?
1. Run 2.1.a with the typing event fix included and the old metrics
2. Run 2.1.a with the typing event fix included and these new metrics

agnivade:
Si!

config/coordinator.sample.json
```diff
@@ -46,6 +46,34 @@
...
+        "Threshold": 0.1,
+        "MinIntervalSec": 60,
+        "Alert": true
+      },
+      {
+        "Description": "Total TCP listen drops",
+        "Query": "avg(rate(node_netstat_TcpExt_ListenDrops{instance=~\"app.*\"}[1m]))",
+        "Threshold": 20,
+        "MinIntervalSec": 60,
+        "Alert": true
```
streamer45, line 76: I wonder if any of these should be targeting the proxy instance as well given it's a key part of the system in most cases.

streamer45 (commented):
Nice work! 

Any reason for not including a threshold on `node_netstat_TcpExt_TCPTimeouts`?

Also agreed on running an unbounded test as sanity check.

config/coordinator.sample.json
agarciamontoro: Oh, darn, that's true, that's definitely what we need to target! I'll run the baseline test (with old metrics) and measure this against the proxy as well.

agnivade: oooh great catch! yes they should be targeted for the proxy.

streamer45 (approved):
Thanks :100:

agarciamontoro:
After running some unbounded tests, we decided to use the percentage of TCP timeouts in both the proxy and app nodes as the indicator of the connection health. Please refer to the discussion in [this thread](https://community.mattermost.com/core/pl/qt31ezyz4tygijx6639co1x4ar) for more info and specific details. The summary is:
- We discarded the metrics for the TCP segments retransmitted and TCP listen dropouts
- We added a couple of metrics for the TCP timeouts, as a percentage of the currently established connections, with different thresholds for proxy (0.2%) and app nodes (3%).
See the metrics in place for each of the four tests we ran:
#### Baseline
Baseline test with only the old metrics.
image
####  Baseline + memory + segments retransmitted
The new memory metric did not get used at all. The segments retransmitted were cut at 2% for both the proxy and the app nodes. This made the test way longer, and the number of users was not affected at all.
image
####  Baseline + memory + timeouts proxy
The percentage of TCP timeouts for the proxy was set at 0.2%. The number of users was slightly reduced (from 42K to 41.3K), but the time of the test remained the same as the baseline.
image
#### Baseline + memory + timeouts proxy + timeouts app
The percentage of TCP timeouts for the proxy was set at 0.2%, and the percentage for the app was set at 3%. The number of users was slightly reduced (from 41.3K to 41.0K), but the time of the test remained the same.
image

agarciamontoro:
Friendly ping on this, @agnivade

agnivade (approved):
Apologies for the delay.

PR #665, opened 2023-11-14
Disable sentry in load tests

agnivade:
If not disabled, crashes from the tests start
polluting the sentry dashboard


agarciamontoro (approved):
Makes sense, thanks!

PR #666, opened 2023-11-15
MM-54715: Set service environment automatically

agnivade:
We set the service environment to production
to avoid unexpected license validation failures.

https://mattermost.atlassian.net/browse/MM-54715


agarciamontoro (approved):
Ah, didn't know about those `Persistent{Pre,Post}Run` fields, quite useful!

streamer45 (approved):

PR #668, opened 2023-11-20
Bump MySQL version to 8.0

agnivade:
https://mattermost.atlassian.net/browse/MM-55589


agnivade:
Strangely, this still creates a 5.7 instance. Need to investigate some more.

agnivade:
Woops my bad, I changed the param group, not the engine version. :facepalm: Monday morning.

streamer45 (approved):
Thanks :+1:

agnivade:
@agarciamontoro - Wondering if @DHaussermann should give this a quick smoke test just to ensure there is no regression?

deployment/terraform/assets/cluster.tf
```diff
@@ -318,7 +318,7 @@ resource "aws_rds_cluster_instance" "cluster_instances" {
 
 resource "aws_db_parameter_group" "db_params_group" {
   name   = "${var.cluster_name}-db-pg"
-  family = var.db_instance_engine == "aurora-mysql" ? "aurora-mysql5.7" : "aurora-postgresql12"
+  family = var.db_instance_engine == "aurora-mysql" ? "aurora-mysql8.0" : "aurora-postgresql12"
```
agarciamontoro, line 321: Oh, huh, the one for postgres should have been `aurora-postgresql14` :thinking:

agarciamontoro (approved):
Woohoo! Thanks, Agniva :tada: And yes, we should test this similarly to how we tested the Postgres bump in https://github.com/mattermost/mattermost-load-test-ng/pull/590

agnivade:
Closing for now. To be reopened on Feb 2024.

agnivade:
Reopening this now! FYI @agarciamontoro @streamer45

agarciamontoro:
@agnivade, next step here would be for @DHaussermann to test it, or is there anything else we need to do before that?

agnivade:
Yep, nothing else needed.

DHaussermann (approved):
@agnivade everything seem normal with the Load testing process.
- The deployments have changes as expected. 
image
- The deployed instance seem to function normally and I could see the regular activity from the load test as expected
image
- The data generated in Graphana for both bounded and unbounded seems typical
image

No issues found LGTM!

I couldn't think of a simple way to benchmark the MySql versions against each other. As this is configured at the Terraform level, both base and new will always use the same version of MySql. @agarciamontoro, am I mistaken? I feel like I did this for Postgres once a few months ago but, can't recall how.
Also, I understand that comparing performance was not the goal here but I was curious on if there are appreciable performance improvements.

agnivade:
Thanks Dylan!

agarciamontoro:
> I couldn't think of a simple way to benchmark the MySql versions against each other. As this is configured at the Terraform level, both base and new will always use the same version of MySql. @agarciamontoro, am I mistaken? I feel like I did this for Postgres once a few months ago but, can't recall how.

A late answer, sorry, but here it goes: the only way is to run both tests independently and then manually generate a report for each of them (with `report generate`) and compare them (with `report compare`). It's what the `comparison` sub-command makes automatically, but you'd need to do it manually.

PR #670, opened 2023-11-27
Revert "MM-55312: Bump MaxNotificationsPerChannel"

agnivade:
Reverts mattermost/mattermost-load-test-ng#662

After a discussion with @jwilander, he suggested that it's okay for this setting to be 1000 for large installs.


streamer45 (approved):

agarciamontoro (changes requested):
What's the motivation to keep it as 1000? Also, now that we're here, can we actually be more explicit and keep the line with 1000 even if that's the default? It sorts of documents our decision.

jwilander:
I was chatting with Agniva about this in a DM. To me it's ok to require certain configurations to scale to the highest limits. Ideally we'd be able to scale without having to disable part of the feature set, and that's something worth striving for, but until the we can support it it's entirely ok to tell a customer they need to disable typing events for channels over 1000 users if they want to scale as high as possible.

agarciamontoro (approved):
Thanks!

PR #675, opened 2023-12-07
[MM-55728] Add user reporting load tests

devinbinnie:
#### Summary
This PR adds a new load test to generate a report of all users using the new reporting query. It should run relatively infrequently since we don't expect the report to be generated all that often.

The test will loop through the entire list of users (unbounded) until it reaches the end and then finish.

Relevant Server PR: https://github.com/mattermost/mattermost/pull/25499

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55728


loadtest/control/simulcontroller/actions.go
```diff
@@ -2062,3 +2062,37 @@ func (c *SimulController) openPermalink(u user.User) control.UserActionResponse
...
+
+func (c *SimulController) generateUserReport(u user.User) control.UserActionResponse {
+	// Simulate scrolling through the entire list of users
+	// (should be similar to generating the complete report and exporting it)
+
+	lastColumnValue := ""
+	lastId := ""
+	totalUsers := 0
+
+	for {
```
devinbinnie, line 2074: Is there a way to throttle this a little bit? Or does it make sense to hit the API as fast as possible?

loadtest/control/simulcontroller/actions.go
```diff
@@ -2062,3 +2062,37 @@ func (c *SimulController) openPermalink(u user.User) control.UserActionResponse
 
 	return control.UserActionResponse{Info: fmt.Sprintf("clicked permalink on post %s", postID)}
 }
+
+func (c *SimulController) generateUserReport(u user.User) control.UserActionResponse {
```
devinbinnie, line 2066: Would we want to test a bunch of scenarios here? To me likely the biggest load on the system would be to return as many users as possible, but are we wanting to test different sorting/filtering here too?

loadtest/control/simulcontroller/controller.go
```diff
@@ -202,6 +202,11 @@ func getActionList(c *SimulController) []userAction {
 			run:       c.reconnectWebSocket,
 			frequency: 0.144,
 		},
+		{
+			name:      "GenerateUserReport",
+			run:       c.generateUserReport,
+			frequency: 0.0001,
```
devinbinnie, line 208: This looks like the most infrequent value we use, so I went with that.

devinbinnie:
@agnivade @streamer45 @agarciamontoro This is now blocking further work on the Admin Reporting piece, is there anything I can do to help expedite the review?

agnivade:
Apologies Devin, I was mainly waiting for Alejandro. But I'll look into this first thing in the morning.

agarciamontoro:
Sorry for the delay here! Will review today :)

loadtest/control/simulcontroller/actions.go
agarciamontoro: > Would we want to test a bunch of scenarios here? To me likely the biggest load on the system would be to return as many users as possible, but are we wanting to test different sorting/filtering here too?


Ideally, we'd have a behaviour as similar as possible as that of a real server (on average during a high traffic time window). That's not possible yet, since the feature is not in production, so I'm happy covering the use case with the biggest load. Once the webapp side of the feature is done and we know how the interaction will look like, we can always iterate here.

loadtest/control/simulcontroller/actions.go
agarciamontoro: > Is there a way to throttle this a little bit? Or does it make sense to hit the API as fast as possible?

What we want is to simulate what real user would do. Again, we don't have data on that yet, but I guess we can do an educated guess and try to throttle it to make it more like a human-triggered action.

loadtest/control/simulcontroller/actions.go
```diff
@@ -2062,3 +2062,37 @@ func (c *SimulController) openPermalink(u user.User) control.UserActionResponse
...
+
+	lastColumnValue := ""
+	lastId := ""
+	totalUsers := 0
+
+	for {
+		report, err := u.GetUsersForReporting(&model.UserReportOptionsAPI{
+			UserReportOptionsWithoutDateRange: model.UserReportOptionsWithoutDateRange{
+				SortColumn:          "Username",
+				PageSize:            50,
```
agarciamontoro, line 2078: Nit: could we define a variable with the page size and use it below as well?

loadtest/control/simulcontroller/actions.go
```diff
@@ -2062,3 +2062,37 @@ func (c *SimulController) openPermalink(u user.User) control.UserActionResponse
...
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+
+		totalUsers += len(report)
+		if len(report) < 50 {
+			break
+		}
+
+		lastColumnValue = report[len(report)].Username
+		lastId = report[len(report)].Id
```
agarciamontoro, line 2094: Should this be one less?
```suggestion
		lastColumnValue = report[len(report) - 1].Username
		lastId = report[len(report) - 1].Id
```

agarciamontoro (changes requested):
Thanks for this, @devinbinnie! I have left some comments below, but wanted to discuss first the big picture of this. I guess my first question is maybe a bit dumb: how will the user trigger this endpoint? The PR that you linked has only the server implementation, so I don't know how it will look like.

agarciamontoro:
You can merge https://github.com/mattermost/mattermost-load-test-ng/pull/677 into this branch, it should resolve the CI failures you're seeing here.

agnivade (changes requested):
@devinbinnie - I am getting this permission error while running it:

```
 control/simulcontroller.(*SimulController).generateUserReport loadtest/control/simulcontroller/actions.go:2086 You do not have the appropriate permissions., userId=gy5s3k5p9pyyzqgr3hod8obkpa, permission=sysconsole_read_user_management_users caller="loadtest/loadtest.go:60" controller_id=5 user_id=gy5s3k5p9pyyzqgr3hod8obkpa
```

Can you tell me what is required to add the necessary permissions?

devinbinnie:
> @devinbinnie - I am getting this permission error while running it:
> 
> ```
>  control/simulcontroller.(*SimulController).generateUserReport loadtest/control/simulcontroller/actions.go:2086 You do not have the appropriate permissions., userId=gy5s3k5p9pyyzqgr3hod8obkpa, permission=sysconsole_read_user_management_users caller="loadtest/loadtest.go:60" controller_id=5 user_id=gy5s3k5p9pyyzqgr3hod8obkpa
> ```
> 
> Can you tell me what is required to add the necessary permissions?

Ahh, you need to be an administrator. I should probably add that check in there.

loadtest/control/simulcontroller/actions.go
```diff
@@ -2073,3 +2073,42 @@ func (c *SimulController) openPermalink(u user.User) control.UserActionResponse
 
 	return control.UserActionResponse{Info: fmt.Sprintf("clicked permalink on post %s", postID)}
 }
+
+func (c *SimulController) generateUserReport(u user.User) control.UserActionResponse {
+	// Simulate scrolling through the entire list of users
+	// (should be similar to generating the complete report and exporting it)
+
+	if isAdmin, err := u.IsSysAdmin(); err != nil || !isAdmin {
```
devinbinnie, line 2081: @agnivade @agarciamontoro Is this the correct way to check for sys admin? I feel like given the frequency of running this function, is it likely that a user would be a sysadmin? Not sure if I'm understanding this correctly or not.

agnivade: Hmm .. this is a problem. This is actually exposed only in the system console right? Then I think we are doing this the wrong way. We need to simulate a full system console page load, and then navigate to the user reporting section.

Moreover, in the simulcontroller, all the users are normal users, not sysadmins. 

We can ignore the system console load for the time being. But to actually make this call work, we need to randomly bump some users to sysadmins. To achieve that, we need to modify this code: https://github.com/mattermost/mattermost-load-test-ng/blob/9fe059e92a159617b2787004c3e47ae6633e4569/api/agent.go#L387-L389 to simply take https://github.com/mattermost/mattermost-load-test-ng/blob/9fe059e92a159617b2787004c3e47ae6633e4569/api/agent.go#L433-L435 instead. 

And we can do this once every 50 users or so. So something like `if id %50 == 0 { // use sysadmin credentials} ` would work.

I think this should be acceptable for now. cc @streamer45 as well for thoughts.

streamer45: I agree with the above. I'd probably expect a new configuration setting under [`UsersConfiguration`](https://github.com/mattermost/mattermost-load-test-ng/blob/9fe059e92a159617b2787004c3e47ae6633e4569/loadtest/config.go#L126) to control how many admins to run, either as an absolute number or percentage of total users.

One important thing to note is that having an action that results in a no-op for most users will skew other frequencies so I'd probably queue the action into the [list](https://github.com/mattermost/mattermost-load-test-ng/blob/9fe059e92a159617b2787004c3e47ae6633e4569/loadtest/control/simulcontroller/controller.go#L18) only if the controller is an admin vs having it in there for all users and returning early if not admin.

Also, we should create a ticket for the system console load which I agree to be out of scope for now, something our team can probably take care of.

@devinbinnie Please let me know if you need help with the above. I realize some of this boilerplate should have probably been there already, you just happen to be a bit of a pioneer in this instance :p

devinbinnie: @agnivade @streamer45 Have a look :)

loadtest/config.go
```diff
@@ -135,6 +135,8 @@ type UsersConfiguration struct {
 	MaxActiveUsers int `default:"2000" validate:"range:(0,]"`
 	// The average number of sessions per user.
 	AvgSessionsPerUser int `default:"1" validate:"range:[1,]"`
+	// The percentage of users generated that will be system admins
+	PercentOfUsersAreAdmin float64 `default:"0.02" validate:"range:[0,1]"`
```
streamer45, line 139: A value of `0.02` seems a bit high as it will result in logging 40 admins if running with defaults. I think running one or two would be more reasonable.

We'll need to update a couple of related things as well:

- https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/config/config.md#usersconfiguration
- https://github.com/mattermost/mattermost-load-test-ng/blob/9fe059e92a159617b2787004c3e47ae6633e4569/config/config.sample.json#L48-L52

api/agent.go
```diff
@@ -365,17 +365,26 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
 
 		username := fmt.Sprintf("%s-%d", namePrefix, id)
 		email := fmt.Sprintf("%s-%d@example.com", namePrefix, id)
 		password := "testPass123$"
 
-		// If UsersFilePath was set, and we haven't yet consumed all of the credentials
-		// provided there, ovewrite this user's credentials with the next available
-		// user in that file
-		if len(creds) > 0 && id < len(creds) {
+		if modAdmins > 0 && id%modAdmins == 0 {
```
streamer45, line 380: One caveat with this approach is that we'd always be running admins first and always using the same user. Not a big deal I suppose but something to note.

streamer45 (commented):
Looks good @devinbinnie, left a couple of minor comments.

loadtest/config.go
devinbinnie: So we have a bit of a balancing act to do here, because the frequency in which the report is run is also the absolute lowest, so I don't know how often that would be run vs. the number of admins. If the plan here is a load test though, then maybe more than 1 or 2 admins is warranted.

I would expect we go for worst-case scenario here, so I think more than 1 or 2 is better, but maybe also less than 40. @agnivade @agarciamontoro Thoughts?

I'll add the sample config.

streamer45: Fair enough, generally though I'd rather bump the frequency of the action than skewing the simulation with more users. Maybe we could refine this by gathering some data on percentage of active admins from our reference customer. Something to think about @agarciamontoro.

streamer45 (approved):
Thanks!

loadtest/user/userentity/helper_test.go
```diff
@@ -46,10 +46,11 @@ type config struct {
 		PercentGroupChannels        float64 `default:"0.1" validate:"range:[0,1]"`
 	}
 	UsersConfiguration struct {
```
devinbinnie, line 48: Just curious why do we have these structs duplicated for this test?

agnivade: There was an import loop somewhere or some other problem due to which I had to duplicate it.

loadtest/control/simulcontroller/controller.go
```diff
@@ -203,6 +203,16 @@ func getActionList(c *SimulController) []userAction {
 			frequency: 0.144,
 		},
 	}
+
+	if isAdmin, err := c.user.IsSysAdmin(); err != nil && isAdmin {
```
agnivade, line 207: Unfortunately, this doesn't work here because the `roles` property of a user isn't initialized until the user logs in. So I'm afraid we have to fall back to checking this in the action itself. Also there's a mistake here, it should be `err == nil`

agnivade (changes requested):

agnivade (approved):
Sent a fix myself.

loadtest/config.go
agarciamontoro: Will review the PR again tomorrow, but asked about this percentage [here](https://community.mattermost.com/private-core/pl/b39wqaat4bg1p8euzcuzo5r3ia):
Community: 0.09%
Reference customer: 0.05%

devinbinnie:
@agarciamontoro Once you get a chance to review this, if there's no changes feel free to merge as I'm going to be off until the new year :)

agarciamontoro (changes requested):
I am slightly concerned about the skewing of the frequencies with this change, since we're now adding two sets of different users sharing the same set of actions, so the frequencies will differ from what we define in the first place.
What if we have some admin client floating around (maybe a global admin client per agent) and, when a controller picks an admin action, it uses that client instead? That way we don't even need to compute a percentage of admin actions (although we could still do that), just have an admin per agent asleep until it's needed.We've discussed this approach in the past and there were some concerns as well (e.g. it breaks the controller isolation, the code will be slightly more complex), but I'd rather have this than adding a new layer of indirection to the computation of the frequencies :thinking: 

We can discuss when I'm back from holidays if you want. Sorry for the delay on the review!

agnivade:
> I am slightly concerned about the skewing of the frequencies with this change, since we're now adding two sets of different users sharing the same set of actions, so the frequencies will differ from what we define in the first place.

Just to clarify here, when you say "skewing of the frequencies", all that's happening here is that for a set of users, one action is going to return faster than the rest. One idea that we can do is introduce an artificial random delay for normal users to counter-balance the effect. 

> What if we have some admin client floating around (maybe a global admin client per agent) and, when a controller picks an admin action, it uses that client instead? That way we don't even need to compute a percentage of admin actions (although we could still do that), just have an admin per agent asleep until it's needed.

Are you suggesting every user become an admin? That could be an option, although slightly different from real life. 

Overall, I'd suggest that we merge this PR in this state, and let @devinbinnie move on to other things, while we fix this in a better way ourselves. Fixing internal load-test behavior is totally out of scope for Devin, and we shouldn't be blocking this PR any further. cc @streamer45 @agarciamontoro

streamer45:
Yeah, I specifically [asked](https://github.com/mattermost/mattermost-load-test-ng/pull/675#discussion_r1425685059) to add the action to the controlled admins only so that we wouldn't skew any normal user. And now we can configure the percentage of admins so I don't see any blocking issue with this approach.

loadtest/config.go
streamer45: Just to confirm, in the config we use the [0,1] range for percentage, so I'd expect those values to be 100 times lower when set.

agnivade:
I reverted that: https://github.com/mattermost/mattermost-load-test-ng/pull/675#discussion_r1426146516.

streamer45:
> I reverted that: [#675 (comment)](https://github.com/mattermost/mattermost-load-test-ng/pull/675#discussion_r1426146516).

I see. This is why we have init actions though. We should be able to generate the full actions list after those have completed.

streamer45:
Also, if it makes it simpler, we could just pass a `isAdmin` flag to `simulcontroller.New()` for the time being. It wouldn't be too bad.

agnivade:
Sounds good. I'll defer to Alejandro on this when he comes back. There are some good solutions here.

agarciamontoro:
### TL;DR
All of my following digression is definitely out of scope for this PR, and I'm ok with unblocking it as is and fixing later ourselves.

### Digression

> Just to clarify here, when you say "skewing of the frequencies", all that's happening here is that for a set of users, one action is going to return faster than the rest. One idea that we can do is introduce an artificial random delay for normal users to counter-balance the effect.

What I mean with "skewing of the frequencies" is that the numbers we use for action frequencies will not be informative. Let me explain it with an example:

- Let's say that we have 10 users: 1 of them is an admin and the other 9 are normal users.
- And let's say that we have 2 actions: one that can be executed by all 10 users (a non-admin action) and other that can be executed only by the one admin user (an admin action)
- The action frequencies happen to be equal (this is informed by the metrics in the production servers, where they happen roughly the same amount of times), so we set both frequencies to 1.
- Let's also assume that both actions take the same amount of time for the users to execute them.

The problem here is that only the admin user is able to execute both actions. That user, in the long run, will execute each action roughly the same amount of times, given the same frequencies we had set: that's good. But the other 9 users won't: they will only execute the non-admin action. If each user runs 1K actions, we'll have (more or less):
- 9 users * 1K non-admin actions = 9K non-admin actions --- All the actions from the non-admin users
- 1 user * 0.5K non-admin actions = 0.5K non-admin actions --- Half the actions for the admin user
- 1 user * 0.5K admin actions = 0.5K admin actions --- Half the actions for the admin user

Since users executed 10K actions and both frequencies were the same, we would have expected to have 5K normal actions and 5K admin actions executed during this test, but we ended up with 9.5K normal actions and 0.5K admin actions.

Even if skipping the admin action takes as much time as executing the non-admin action, and thus the non-admin users only execute half the actions they would otherwise, staying idle for the other half of time, we would have 5K (4.5K from the 9 non-admin users + 0.5K from the admin user) non-admin actions and 0.5K admin actions: still far from the expected 50/50 split among the two actions.

I also don't see how Claudio's solution would solve this problem: having a no-op is effectively the same as not having the action, right? There will be a slight delay when having the no-op, but even if we don't have the admin actions for non-admin users, the frequencies are still not well represented.

> Are you suggesting every user become an admin? That could be an option, although slightly different from real life.

Not exactly, I agree that having all users logged in as admins is too much of a change from real life. What I was thinking was for every `UserEntity` to still use their normal `client`, but apart from that, have an `adminClient` already logged in and shared among *all users* in an agent. If a user happens to run an admin action, they use the global `adminClient` instead of their own `client`. This is not ideal, and there may be better solutions, but it does fix the skewing of the frequencies, since all users have the opportunity to execute all actions.

> But Alejandro, the frequency we chose for this action is completely made up!

Yes, it is, so all my arguments above become kind of moot as of today, that's why I think it's ok to just merge and fix later. But when we collect real data and choose a real frequency, I just want to keep it as simple as possible. If the real data says this admin action is executed 1% of times compared to the `createPost` action, then we'll set 0.01 as its frequency and call it a day. That's my goal with all this discussion.

streamer45:
> Since users executed 10K actions and both frequencies were the same, we would have expected to have 5K normal actions and 5K admin actions executed during this test, but we ended up with 9.5K normal actions and 0.5K admin actions.

Not sure I fully understand this. If we only have 1 admin user I'd expect the results to be exactly as you showed as the frequencies are relative.

Also, I'd probably consider admin users on their own. We could have them perform admin actions only if it helps to simplify things versus mixing them together.

> I also don't see how Claudio's solution would solve this problem: having a no-op is effectively the same as not having the action, right? There will be a slight delay when having the no-op, but even if we don't have the admin actions for non-admin users, the frequencies are still not well represented.

Agreed not a solution to your problem statement above but I'd argue that the "slight delay" is key. No-op means we still wait idle for some time which skews everything in the long run because that time is gone instead of used for the next action. It ultimately alters the rate of the controlled users.

agarciamontoro:
> Not sure I fully understand this. If we only have 1 admin user I'd expect the results to be exactly as you showed as the frequencies are relative.

I guess it's a matter of expectations. My goal is to simplify how we compute the frequencies (maybe I'm just traumatized by our latest efforts :joy:), so if we measure in a real server that the relative frequency of an action is 0.1, I expect to use that exact number in the action frequency. But if we have this additional layer of indirection of only a percentage of users being able to run certain actions, we need to take that into account to compute the number we set.

> Also, I'd probably consider admin users on their own. We could have them perform admin actions only if it helps to simplify things versus mixing them together.

I'm ok with this, and ideally it would be the solution: if we can model admins separately, as in getting frequencies for them independently, then we can consider just two separate sets of users. I don't think that's possible with the data we currently have, though.

> Agreed not a solution to your problem statement above but I'd argue that the "slight delay" is key. No-op means we still wait idle for some time which skews everything in the long run because that time is gone instead of used for the next action. It ultimately alters the rate of the controlled users.

Yes, I agree with this, we definitely need that.

agnivade:
I am completely supportive of merging this as-is and we can own any further improvements.

agarciamontoro (approved):

PR #676, opened 2023-12-11
[MM-56069] Add check for emoji limit

devinbinnie:
#### Summary
This PR checks for the reaction emoji limit when adding a reaction.

Related server PR: https://github.com/mattermost/mattermost/pull/25331

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56069

PR #677, opened 2023-12-12
Update Mattermost dependency

agarciamontoro:
#### Summary
Fixing the errors seen in #675 and #676: we need to update not only `v8` (on which we still rely for a couple of things), but also `public`.

This is turns updated `mlog`, that recently introduced generic methods (`Int[T]` instead of `Int64`).

#### Ticket Link
--


agarciamontoro:
@devinbinnie: this fixes your problems with CI. Feel free to merge master into your branches when this PR is merged.

streamer45 (approved):
Makes sense :+1:

cpoile (approved):
👍

PR #678, opened 2023-12-13
MM-56188: Add presence indicator to load test

agnivade:
We send the message via the websocket whenever
the client state changes.

https://mattermost.atlassian.net/browse/MM-56188

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -294,8 +294,10 @@ func (c *SimulController) switchTeam(u user.User) control.UserActionResponse {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
-	if err := u.UpdateActiveTeam(team.Id); err != nil {
-		mlog.Warn("Failed to update active team", mlog.String("team_id", team.Id))
+	if c.user.Store().FeatureFlags()["WebSocketEventScope"] {
+		if err := u.UpdateActiveTeam(team.Id); err != nil {
+			mlog.Warn("Failed to update active team", mlog.String("team_id", team.Id))
+		}
 	}
 
 	// We should probably keep track of the last channel viewed in the team but
@@ -535,8 +537,10 @@ func switchChannel(u user.User) control.UserActionResponse {
 		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
 
-	if err := u.UpdateActiveChannel(channel.Id); err != nil {
-		mlog.Warn("Failed to update active channel", mlog.String("channel_id", channel.Id))
+	if u.Store().FeatureFlags()["WebSocketEventScope"] {
+		if err := u.UpdateActiveChannel(channel.Id); err != nil {
+			mlog.Warn("Failed to update active channel", mlog.String("channel_id", channel.Id))
+		}
 	}
 
 	return control.UserActionResponse{Info: fmt.Sprintf("switched to channel %s", channel.Id)}
@@ -1419,6 +1423,15 @@ func (c *SimulController) initialJoinTeam(u user.User) control.UserActionRespons
 		return c.joinTeam(c.user)
 	}
 
+	if c.user.Store().FeatureFlags()["WebSocketEventScope"] {
+		// Setting the active thread to empty to allow the optimization to kick in early.
+		// We don't do this in the webapp to keep the code simple, but it's okay to do this in load-test
+		// to magnify the boost.
+		if err := u.UpdateActiveThread(""); err != nil {
+			mlog.Warn("Failed to update active thread", mlog.String("channel_id", ""))
+		}
+	}
+
 	return resp
 }
 
@@ -1738,8 +1751,10 @@ func (c *SimulController) viewThread(u user.User) control.UserActionResponse {
 		}
 	}
 
-	if err := u.UpdateActiveThread(thread.Post.ChannelId); err != nil {
-		mlog.Warn("Failed to update active thread", mlog.String("channel_id", thread.Post.ChannelId))
+	if c.user.Store().FeatureFlags()["WebSocketEventScope"] {
+		if err := u.UpdateActiveThread(thread.Post.ChannelId); err != nil {
+			mlog.Warn("Failed to update active thread", mlog.String("channel_id", thread.Post.ChannelId))
+		}
 	}
 
 	return control.UserActionResponse{Info: fmt.Sprintf("viewedthread %s", thread.PostId)}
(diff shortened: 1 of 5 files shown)


loadtest/user/userentity/websocket.go
```diff
@@ -139,9 +139,10 @@ func (ue *UserEntity) wsEventHandler(ev *model.WebSocketEvent) error {
 // Only on calling Disconnect explicitly, it will return.
 func (ue *UserEntity) listen(errChan chan error) {
 	connectionFailCount := 0
+	var err error
 start:
 	for {
-		client, err := websocket.NewClient4(&websocket.ClientParams{
+		ue.wsClient, err = websocket.NewClient4(&websocket.ClientParams{
```
agnivade, line 145: I realize this is breaking from our established norm. However, it's been a long time since anyone touched this code and I don't think there is anything inherently bad by storing the client in the struct. I know the actual client is pretty resilient against failures, so I don't foresee any issue calling its methods from outside the main goroutine.

To push all activity inside the main for-select loop would require further channels and more coordination which didn't sit right with me. The only edge-case is that the call will fail when the client is closing, which I think is fine. LMK.

agarciamontoro (commented):
I looked at the related server PR ([this one](https://github.com/mattermost/mattermost/pull/25694), right?) to confirm that `UpdateActiveThread` does indeed need the channel ID. Why is that? Why don't we need the root post ID? It was quite confusing to see the header of the function being `UpdateActiveThread(channelId string)` and thought something was wrong.

agnivade:
I am trying to keep the scope to the high level objects like - teamID, channelID. The rootID is embedded as a metadata to the typing event and is not a property in the `WebSocketBroadcast` struct. So yes it'll be slightly more wasteful but it keeps the overall logic simple.

We can move to tracking the actual thread maybe as a future perf improvement.

agarciamontoro (approved):
Thanks! The only ask I have would be to clarify with a comment in the code why `UpdateActiveThread receives a channel ID instead of a thread ID.

streamer45:
@agnivade In case you are wondering, I am mostly waiting for the server and client logic to be finalized before reviewing this.

agnivade:
@streamer45 @agarciamontoro - This can now be reviewed.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1410,6 +1423,15 @@ func (c *SimulController) initialJoinTeam(u user.User) control.UserActionRespons
 		return c.joinTeam(c.user)
 	}
 
+	if c.user.Store().FeatureFlags()["WebSocketEventScope"] {
+		// Setting the active thread to empty to allow the optimization to kick in early.
+		// We don't do this in the webapp to keep the code simple, but it's okay to do this in load-test
+		// to magnify the boost.
```
streamer45, line 1429: When in doubt I'd probably want the load-test simulation to perform worse than webapp to be on the safe side. I'd be okay with keeping this to prove the ideal improvements though.

loadtest/user/userentity/websocket.go
streamer45: The reason we never did that is because it's inherently racy and we didn't want the complexity of adding a mutex to `UserEntity`. Now any of the `UpdateXXX` methods has a race condition in accessing the client object.

One decent way to solve this would be to repurpose the `wsTyping` channel to be a generic event channel and then switch on event type to call the appropriate method.

streamer45: No other concerns other than the race accessing the client object which is the reason we kept it internal to `listen` but I see you took care of that so all good :+1:

loadtest/control/simulcontroller/actions.go
agnivade: The problem is our simulation is not close enough to the webapp. We don't differentiate between loading a thread from RHS or from global thread view. It's all just a loadThread action. Until we differentiate that, we'll have to approximate somehow.

loadtest/user/userentity/websocket.go
```diff
@@ -239,9 +254,47 @@ func (ue *UserEntity) SendTypingEvent(channelId, parentId string) error {
...
+	// We either load a channel or load a thread. For now, we just set `is_thread_view`
+	// as both true and false to set the scope.
+	ue.dataChan <- threadPresenceMsg{
+		channelId:  channelId,
+		threadView: true,
+	}
+	ue.dataChan <- threadPresenceMsg{
+		channelId:  channelId,
+		threadView: false,
+	}
```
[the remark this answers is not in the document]
agnivade: The webapp has 2 different ways of loading a thread. One from RHS and one from global thread viewer. That means one can have 2 threads open at the same time. Since we don't simulate that, we are approximating it here.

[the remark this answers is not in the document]
agnivade: See https://github.com/mattermost/mattermost/pull/25794 for more details. And also the comment at the end of this section: https://mattermost.atlassian.net/wiki/spaces/~5de4a29d752c1b0d114f23c9/pages/2556788794/Rearchitecting+WebSocket+Events#Add-concept-of-%E2%80%9Cactive_channel%E2%80%9C-and-%E2%80%9Cactive_team%E2%80%9C-to-a-connection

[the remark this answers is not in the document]
agarciamontoro: Ah, I get it now, we're simply opening both threads, right? That makes sense, thanks!

loadtest/control/simulcontroller/actions.go
agarciamontoro: But are we making things more performant here with this? Or am I misunderstanding the goal?

loadtest/user/userentity/websocket.go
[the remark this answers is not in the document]
agnivade: Basically. Actually we are making the API call to open just a single thread. But we are setting the scope such that both threads are open.

loadtest/control/simulcontroller/actions.go
agnivade: Correct, we are making it more performant here.

agarciamontoro: That's what worries me (and what I think concerned Claudio as well). Is the performance improvement in here really needed?

agnivade: I talked about this earlier in this comment: https://github.com/mattermost/mattermost-load-test-ng/pull/678#discussion_r1459290297

Until we do that, we have to approximate somehow. And if we choose not to set the active thread at all, then there will be no improvements at all, so it will be impossible to measure the impact of the change.

agarciamontoro: Ok, let's go with it then! Thank you for the patience :)

agarciamontoro (approved):
:rocket:

agnivade:
The feature PR is merged. Time to get this in!

PR #681, opened 2023-12-28
Bump server ping timeout

agnivade:
In the latest data dump, it seems like the migrations take a while to run. It takes more than a minute for the server to be up.

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -691,7 +691,7 @@ func pingServer(addr string) error {
 	mlog.Info("Checking server status:", mlog.String("host", addr))
 	client := model.NewAPIv4Client(addr)
 	client.HTTPClient.Timeout = 10 * time.Second
-	dur := 60 * time.Second
+	dur := 240 * time.Second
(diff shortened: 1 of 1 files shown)


deployment/terraform/create.go
```diff
@@ -691,7 +691,7 @@ func pingServer(addr string) error {
 	mlog.Info("Checking server status:", mlog.String("host", addr))
 	client := model.NewAPIv4Client(addr)
 	client.HTTPClient.Timeout = 10 * time.Second
-	dur := 60 * time.Second
+	dur := 240 * time.Second
```
streamer45, line 694: Maybe we could make these package constants to keep them all in the same place.

streamer45 (approved):

agarciamontoro (approved):
Looks good to me, but I'm curious: what data dump do you mean? And the migrations from which to which version?

agnivade:
> but I'm curious: what data dump do you mean? And the migrations from which to which version?

https://community.mattermost.com/core/pl/xggk7tpyppbmdj5c5fcsctd9te

PR #682, opened 2024-01-10
Exit on failure of any command inside remote-exec

agarciamontoro:
#### Summary
According to the documentation on the [remote-exec provisioner](https://developer.hashicorp.com/terraform/language/resources/provisioners/remote-exec):

> Since `inline` is implemented by concatenating commands into a script, `on_failure` applies only to the final command in the list. In particular, with `on_failure = fail` (the default behaviour) earlier commands will be allowed to fail, and later commands will also execute. If this behaviour is not desired, consider using `"set -o errexit"` as the first command.

Gonna test this now.

#### Ticket Link
--


streamer45 (approved):
Makes sense :+1:

agnivade (approved):
Nice one!

agarciamontoro:
I've run several deployments with this change, and it seems to be working. Merging now.

PR #685, opened 2024-01-17
MM-56288: Fix panic while getting config

agnivade:
The entire Config struct isn't set until the API call is made.
Clients are expected to use the ClientConfig which is guaranteed to be
always set before actions are called. So we use that instead.

https://mattermost.atlassian.net/browse/MM-56288

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -825,13 +825,18 @@ func (c *SimulController) addReaction(u user.User) control.UserActionResponse {
 		}
 	}
 
-	if reactionLimit := u.Store().Config().ServiceSettings.UniqueEmojiReactionLimitPerPost; reactionLimit != nil {
+	reactionLimit, err := strconv.ParseInt(u.Store().ClientConfig()["UniqueEmojiReactionLimitPerPost"], 10, 64)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


agnivade:
I couldn't see a way to write a test for this. LMK.

loadtest/control/simulcontroller/actions.go
```diff
@@ -825,13 +825,18 @@ func (c *SimulController) addReaction(u user.User) control.UserActionResponse {
 		}
 	}
 
-	if reactionLimit := u.Store().Config().ServiceSettings.UniqueEmojiReactionLimitPerPost; reactionLimit != nil {
```
streamer45, line 828: I think there's a similar line in `gencontroller`. Wondering if it's worth fixing that to avoid hitting it again.

streamer45 (approved):
Looks good, left one small ask.

loadtest/control/simulcontroller/actions.go
agnivade: Done

agarciamontoro (approved):
Thanks for the fix, @agnivade! It seems this is what was affecting the latest ceiling test and the testing of v1.13-rc1. Let's test that hypothesis now :)

PR #691, opened 2024-01-31
MM-54860: Automatically vacuum tables after deployment

agnivade:
https://mattermost.atlassian.net/browse/MM-54860

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -495,7 +495,7 @@ func (t *Terraform) updatePostgresSettings(extAgent *ssh.ExtAgent) error {
 	sqlCmd = "vacuum analyze channels, sidebarchannels, sidebarcategories, posts, threads, threadmemberships, channelmembers;"
 	cmd = fmt.Sprintf("psql '%s' -c '%s'", dns, sqlCmd)
 
-	mlog.Info(fmt.Sprintf("Vacuuming the tables"), mlog.String("cmd", cmd))
+	mlog.Info("Vacuuming the tables", mlog.String("cmd", cmd))
 	if out, err := sshc.RunCommand(cmd); err != nil {
 		return fmt.Errorf("error running ssh command: %s, output: %s, error: %w", cmd, out, err)
 	}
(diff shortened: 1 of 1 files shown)


isacikgoz (approved):
LGTM

deployment/terraform/create.go
```diff
@@ -230,7 +220,19 @@ func (t *Terraform) Create(initData bool) error {
...
+	// have been created by then.
+	if t.output.HasDB() {
+		// Aurora sets the default_text_search_config parameter to 'simple';
+		// a more sane default for a generic deployment is 'english'
+		if t.config.TerraformDBSettings.InstanceEngine == "aurora-postgresql" {
+			if err := t.updatePostgresSettings(extAgent); err != nil {
+				return fmt.Errorf("could not modify default_search_text_config: %w", err)
+			}
+		}
 	}
```
agarciamontoro, line 236: We need to update the comment and the error message here, since we do more in `updatePostgresSettings` now.

deployment/terraform/create.go
```diff
@@ -465,30 +467,35 @@ func (t *Terraform) createAdminUser(extAgent *ssh.ExtAgent) error {
...
+func (t *Terraform) updatePostgresSettings(extAgent *ssh.ExtAgent) error {
 	dns := fmt.Sprintf("postgres://%s:%s@%s/%s?sslmode=disable",
 		t.config.TerraformDBSettings.UserName,
 		t.config.TerraformDBSettings.Password,
 		t.output.DBWriter(),
 		t.config.DBName(),
 	)
 
-	sqlCmd := fmt.Sprintf("ALTER DATABASE %s SET default_text_search_config TO \"%s\"",
+	sshc, err := extAgent.NewClient(t.output.Instances[0].PublicIP)
```
agarciamontoro, line 478: I know this line was here before, but should we double-check that `t.ouput.Instances` has at least one instance and fail otherwise?

agarciamontoro (approved):
Looks great! Thank you for this, it will save us a loooot of time. Have you tested it? I haven't, but I can give it a try if you need it.

agnivade:
> Have you tested it? I haven't, but I can give it a try if you need it.

Yep I have tested it.

PR #694, opened 2024-02-13
MM-55630: Generate a new dashboard with enabled coordinator metrics

agarciamontoro:
#### Summary
A new Grafana dashboard is dynamically generated when there is at least one enabled query in the coordinator. I had to add a new optional `Legend` field to the queries configuration so that we can provide a string to use as legend in the panel. The following screenshot shows the generated dashboard with the sample metrics (not very interesting graphs, since it was a dummy test): 

image

I didn't add the first panel with the number of connected users as shown in the ticket's screenshot, since that's already in the other dashboard.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-55630


config/coordinator.sample.json
```diff
@@ -14,55 +14,63 @@
     "Queries": [
       {
         "Description": "Percentage of HTTP 5xx server errors",
+        "Legend": "Percentage of requests responded from the server with a 5xx status code",
```
agnivade, line 17: The legends are very close to the description. In any case, they should be one or two words at max. I'll suggest values like "percent" or "duration".

deployment/terraform/metrics.go
```diff
@@ -80,6 +82,27 @@ func (t *Terraform) UploadDashboard(dashboard string) (string, error) {
...
+	Width     int
+	PosX      int
+	PosY      int
+	Query     string
+	Threshold float64
+}
+
+const (
+	panelWidth  = 12
+	panelHeight = 9
```
agnivade, line 99: Maybe some comments to explain how you got these numbers?

config/coordinator.sample.json
agarciamontoro: > The legends are very close to the description.

Yeah, the thing is that we're using `Description` like a title or name. I thought about changing Description to Title, thoughts?

> In any case, they should be one or two words at max. I'll suggest values like "percent" or "duration".

Do you feel strongly about them being so short? I think I like a very descriptive legend, but maybe that's just me.

deployment/terraform/metrics.go
agarciamontoro: Sure!

config/coordinator.sample.json
agnivade: > Do you feel strongly about them being so short? 

I do, but there is a functional reason for it. Labels should be short because if there are tens of them, then others become hard to read. In this case, there's just one, but I'd like to maintain the same convention.

I will defer to you for a final decision :)

agarciamontoro: Indeed, my reasoning here was that in this case there is only one, but I'll buy the consistency argument :)

agarciamontoro: Done in https://github.com/mattermost/mattermost-load-test-ng/pull/694/commits/29dc25fe01884c77454003761beb828e3b2f9fec

config/coordinator.sample.json
```diff
@@ -14,63 +14,63 @@
...
+        "Legend": "Percent",
         "Query": "(sum(rate(mattermost_api_time_count{status_code=~\"5..\"}[1m]))/sum(rate(mattermost_api_time_count[1m])))*100",
         "Threshold": 0.025,
         "MinIntervalSec": 60,
         "Alert": true
       },
       {
         "Description": "Average client request duration",
-        "Legend": "Average duration of all requests as measured from the client side",
+        "Legend": "Duration in ms (avg)",
```
agnivade, line 25: You don't even need the unit because it's there in the y axis. But ok to keep this :)

agnivade (approved):

config/coordinator.sample.json
agarciamontoro: It's not, right? And it's seconds, actually! I'll fix it now
image

agnivade: That's because you haven't mentioned the unit in the display section. If you set that to time in seconds, then it automatically adjusts to ms when it's too low. It's far better that way.

agarciamontoro: Yeah, I thought about that. But then we'd have to add yet another field to the coordinator metrics for users to specify the unit. I'd start with this and we can iterate later if needed.

agnivade: Why do you need user input for that? It's always in seconds right?

agarciamontoro: This particular query is, but the queries are arbitrary and user-defined. Most of the ones we use now are plain percentages.

agnivade: Ahhh that's true. Forgot that you are templating the dashboard. All good then :+1:

isacikgoz (approved):

PR #695, opened 2024-02-19
MM-56918: Ingest the DB dump from inside Terraform.Create

agarciamontoro:
#### Summary
When using `config.DBDumpURI`, the generated DB did not get the settings applied from `updatePostgresSettings`, since the latter was executed before ingesting the dump. We cannot simply move the call to `IngestDump` before the call to `Create`, though, because we need the DB cluster up and running to ingest the dump. Thus, this PR moves the call to `IngestDump` inside `Create`, after the DB cluster creation and right before updating the settings.

I'm testing a deployment with these changes as we speak.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56918

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -27,20 +27,12 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
 
-	// If DBDumpURI is not set, we seed data. Otherwise,
-	// we just load the dump.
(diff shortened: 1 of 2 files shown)


agnivade (approved):
Aha good catch! I have to remember why I called `IngestDump` at a high level rather than from inside `Create`. But nothing is coming to mind. If the testing passes, then it's good to go from my side.

agarciamontoro:
Tested with both `DBDumpURI` set and unset, and everything seems to be working as expected. Merging now!

PR #700, opened 2024-02-28
Update dependencies

agarciamontoro:
#### Summary
Ran `make update-dependencies`.

#### Ticket Link
--


agarciamontoro:
PR #700, let's celebrate! :tada:

agnivade (approved):

agarciamontoro:
For some reason, using go 1.19 made golangci-lint fail with this error:
```
WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package model: could not load export data: no export data for "github.com/prometheus/common/model"
ERRO Running error: 1 error occurred:
	* can't run linter goanalysis_metalinter: buildir: failed to load package model: could not load export data: no export data for "github.com/prometheus/common/model"
```

I didn't catch this locally because my local version was 1.20. I'm not sure why 1.19 triggered this error, but an update has been pending for a while in any case, so here it is.

Re-requesting review for the new changes.

agnivade (approved):

.github/workflows/ci.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
         with:
-          go-version: "1.19.5"
+          go-version: "1.20.14"
```
streamer45, line 23: Is the dot release actually needed here?

streamer45 (approved):

.github/workflows/ci.yml
agarciamontoro: Not sure, to be honest. It was a quick fix so I just copied the same pattern. Let me check.

agarciamontoro: It seems it's not needed: https://github.com/marketplace/actions/setup-go-environment#supported-version-syntax Let's see in practice

agarciamontoro: It works :+1:

PR #705, opened 2024-03-11
Bump version to v1.15.0-rc1

agarciamontoro:
#### Summary
Preparing release v1.15.0-RC1

#### Ticket Link
--

### DIFF
--- a/config/deployer.sample.json
+++ b/config/deployer.sample.json
@@ -60,7 +60,7 @@
   "AdminEmail": "sysadmin@sample.mattermost.com",
   "AdminUsername": "sysadmin",
   "AdminPassword": "Sys@dmin-sample1",
-  "LoadTestDownloadURL": "https://github.com/mattermost/mattermost-load-test-ng/releases/download/v1.9.1/mattermost-load-test-ng-v1.9.1-linux-amd64.tar.gz",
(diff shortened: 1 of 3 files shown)


config/deployer.sample.json
```diff
@@ -60,7 +60,7 @@
   "AdminEmail": "sysadmin@sample.mattermost.com",
   "AdminUsername": "sysadmin",
   "AdminPassword": "Sys@dmin-sample1",
-  "LoadTestDownloadURL": "https://github.com/mattermost/mattermost-load-test-ng/releases/download/v1.9.1/mattermost-load-test-ng-v1.9.1-linux-amd64.tar.gz",
```
streamer45, line 63: Is this out of date somehow? Noticing the bump from v1.9 to v1.15.

agarciamontoro: Yeah, this is something we need to fix process-wise. We only bump the version in the release branches, but never on master. We may want to merge into master and then cut the branch? :thinking:

agarciamontoro: > Yeah, that would be my expectation to be honest.

Let's test this now

streamer45: That could also work, at which point the pinning would make sense in the release branch only.

agarciamontoro:
Let's modify the process so that we merge this into master and *then* we cut the branch. Merging into master now.

streamer45 (approved):

PR #708, opened 2024-03-14
MM-57214: Integrate Redis into Terraform deployment

agnivade:
https://mattermost.atlassian.net/browse/MM-57214


agnivade:
Just for review. The code is actually pointing to a branch in the server repo, which needs to be updated once it's merged.

deployment/terraform/metrics.go
```diff
@@ -157,6 +159,31 @@ func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent) error {
...
+		redisTargets = append(redisTargets, "metrics:9121")
+
+		mlog.Info("Enabling Redis exporter", mlog.String("host", t.output.MetricsServer.PublicIP))
+
+		// TODO: Pass username/pass later if we ever start using them internally.
+		// It's possible to configure them on the server, but there is no need to set them up for internal load tests.
+		redisExporterService := fmt.Sprintf(redisExporterServiceFile, redisEndpoint)
+		rdr := strings.NewReader(redisExporterService)
+		if out, err := sshc.Upload(rdr, "/lib/systemd/system/redis-exporter.service", true); err != nil {
+			return fmt.Errorf("error upload redis exporter service file: output: %s, error: %w", out, err)
```
agarciamontoro, line 173:
```suggestion
			return fmt.Errorf("error uploading redis exporter service file: output: %s, error: %w", out, err)
```

deployment/terraform/output.go
```diff
@@ -189,6 +209,11 @@ func (o *Output) HasElasticSearch() bool {
 	return o.ElasticSearchServer.Endpoint != ""
 }
 
+// HasRedis returns whether a deployment has ElasticSaearch installed in it or not.
```
agarciamontoro, line 212:
```suggestion
// HasRedis returns whether a deployment has Redis installed in it or not.
```

deployment/terraform/assets/cluster.tf
```diff
@@ -85,6 +85,7 @@ resource "aws_instance" "app_server" {
       "sudo apt-get install -y mysql-client-8.0",
       "sudo apt-get install -y postgresql-client-11",
       "sudo apt-get install -y prometheus-node-exporter",
+      "sudo NEEDRESTART_MODE=a apt-get install -y redis-tools",
```
agarciamontoro, line 88: What does `NEEDRESTART_MODE=a` mean?

deployment/terraform/metrics.go
```diff
@@ -157,6 +159,31 @@ func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent) error {
 		}
 	}
 
+	if t.output.HasRedis() {
+		redisEndpoint := fmt.Sprintf("redis://%s", net.JoinHostPort(t.output.RedisServer.Address, strconv.Itoa(t.output.RedisServer.Port)))
+		redisTargets = append(redisTargets, "metrics:9121")
```
agarciamontoro, line 164: Nit:
This is the port of the redis exporter, right? I know the rest of definitions here are similar to this one, but they seem a bit cryptic if one doesn't know what they're defining, so maybe we can start slightly improving things  with something like:
```suggestion
		redisExporterPort := 9121
		redisTargets = append(redisTargets, "metrics:"+redisExporterPort)
```

agarciamontoro (commented):
This looks great! A couple of comments:
- Do we want to support multiple Redis nodes at some point? I see everything hardcoded to one node. 0/5 on this at this point, just curious.
- I understand this is still a WIP, but writing it here so that we don't foget: we need to document the newly added settings.

deployment/terraform/assets/cluster.tf
agnivade: Haha this was a fun one. So if you try to install it manually, a screen comes up and it asks you what services you want to restart post installation. This `NEEDRESTART_MODE=a` is the way to automatically say ok to that prompt. `-y` does not take care of this case.

agnivade:
> Do we want to support multiple Redis nodes at some point? I see everything hardcoded to one node. 0/5 on this at this point, just curious.

Nope we don't. Because we don't want to get into handling consistency issues for Redis writer-reader again. One Redis instance is good for now.

streamer45:
Please add me back when ready for final review.

agarciamontoro:
Now that the main PR is to be merged, should we finalize this?

agnivade:
Yep, this is next in my queue once the PR gets merged.

agarciamontoro:
Neat, thanks!

deployment/terraform/metrics.go
agnivade: Isn't the same in the elasticsearch exporter as well?

agnivade:
@agarciamontoro, @streamer45 - Ready for review now. I tested and verified that everything works fine.

deployment/terraform/assets/cluster.tf
```diff
@@ -84,6 +84,7 @@ resource "aws_instance" "app_server" {
       "sudo apt-get install -y mysql-client-8.0",
       "sudo apt-get install -y postgresql-client-14",
       "sudo apt-get install -y prometheus-node-exporter",
+      # "sudo NEEDRESTART_MODE=a apt-get install -y redis-tools",
```
streamer45, line 87: What's the deal with this one?

deployment/terraform/create.go
```diff
@@ -1031,6 +1032,13 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
 		cfg.ElasticsearchSettings.UserIndexReplicas = model.NewInt(numReplicas)
 	}
 
+	if t.output.HasRedis() {
+		cfg.CacheSettings.CacheType = model.NewString(model.CacheTypeRedis)
+		redisEndpoint := net.JoinHostPort(t.output.RedisServer.Address, strconv.Itoa(t.output.RedisServer.Port))
+		cfg.CacheSettings.RedisAddress = model.NewString(redisEndpoint)
+		cfg.CacheSettings.RedisDB = model.NewInt(0)
```
streamer45, line 1039: How about `model.NewPointer`? :p We probably need a ticket to review the whole thing, eventually.

streamer45 (approved):
Nice!

deployment/terraform/assets/cluster.tf
agnivade: This is needed to install the `redis-cli`. But the thing is there is an UI intervention while installing, which is supposedly fixed with `NEEDRESTART_MODE=a`. But that still somehow didn't fix things. So just ignoring for now. If someone needs to connect to Redis, they can install manually.

streamer45: FWIW, the command worked fine on my end (EC2 running Ubuntu `22.04`).

agnivade: Hmm .. weird. I will test it again later.

deployment/terraform/metrics.go
agarciamontoro: Most probably, yes, but I noticed it here because it wasn't me the one writing the code :joy: I'm 1/5 here, though,just a  nit :)

agarciamontoro (approved):
Whoops, missed this one, sorry! All looks good, thank you!

The only thing is the `"sudo NEEDRESTART_MODE=a apt-get install -y redis-tools"` line. What are we doing with this one? I'm ok simply removing it and adding it as an improvement later.

agnivade:
> The only thing is the "sudo NEEDRESTART_MODE=a apt-get install -y redis-tools" line. What are we doing with this one? I'm ok simply removing it and adding it as an improvement later.

Removed it.

PR #715, opened 2024-03-20
MM-57354: Fetch static assets

agnivade:
https://mattermost.atlassian.net/browse/MM-57354


streamer45 (approved):
Nice to see this finally used. 

We may need a review of its logic as I don't think it accounts for some potentially heavy assets such as plugin bundles, although that may require more work.

agnivade:
Yeah, if any JS files are fetched from within JS then they won't be fetched. But the basic react bundle and our entry.js should be fetched.

agarciamontoro (approved):
I swear I reviewed this very early in my day, I guess I never pushed the Submit button. Well, here it goes! Thank you, this is great :)

PR #720, opened 2024-03-27
Add OpenID authentication support using a KeyCloak server

fmartingr:
#### Summary
- Allow users to authenticate with Mattermost by using an OpenID provider
- Adds provisioning for Keycloak server in in development mode with a postgresql database bundled in the same server
- A default realm bundled into the load test with some default credentials.
  - An optional import file can be provided, if none is set, the default one is used.
  - Also a database dump can be provided in the form of an URI and the data will be imported into the SQL database.
- An optional parameter to generate users in the Keycloak instance
- Automatic configuration of the Mattermost instance by using the generated Keycloak server with the provided `KeycloakClientID`, `KeycloakClientSecret` and `KeycloakRealmName` (only needed when a custom import realm file is provided if the default values do not match).
- Documentation


deployment/terraform/create.go
```diff
@@ -220,6 +227,37 @@ func (t *Terraform) Create(initData bool) error {
 		if err := pingServer("http://" + pingURL); err != nil {
 			return fmt.Errorf("error whiling pinging server: %w", err)
 		}
+
+		if t.output.HasKeycloak() {
+			keycloakCmdFmt := "/opt/mattermost/bin/mmctl config set %s '%s' --local"
```
agnivade, line 232: All of this is actually done by setting the model.Config struct. See `(t *Terraform) updateAppConfig`. This code should exist inside that method.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,205 @@
...
+		command = "start"
+	}
+
+	sshc, err := extAgent.NewClient(t.output.KeycloakServer.PublicIP)
+	if err != nil {
+		return err
+	}
+
+	// Create keycloak.env file
+	var keycloakEnvFileContents []string
```
agarciamontoro, line 35: Can we move this declaration closer to where it's used (below the Install realm file if-else block)?

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+func (t *Terraform) populateKeycloakUsers(sshc *ssh.Client) error {
+	keycloakBinPath := "/opt/keycloak/keycloak-" + t.config.ExternalAuthProviderSettings.KeycloakVersion + "/bin"
+	usersTxtPath := t.getAsset("keycloak-users.txt")
+
+	// Check if users file exists. Prevents from creating users multiple times.
+	if _, err := os.Stat(usersTxtPath); err == nil || os.IsExist(err) {
+		return nil
+	}
```
agarciamontoro, line 175: Not sure if this is a dumb question, but: could it happen that the users file exists but that it doesn't contain the users we want? For example, we create a deployment with 10 users and then recreate it with 50.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,205 @@
...
+	// Open users.txt file
+	handler, err := os.OpenFile(usersTxtPath, os.O_CREATE|os.O_WRONLY, 0666)
+	if err != nil {
+		return fmt.Errorf("failed to open keycloak users file: %w", err)
+	}
+	defer handler.Close()
+
+	// Create users
+	for i := 1; i <= t.config.ExternalAuthProviderSettings.GenerateUsersCount; i++ {
+		username := fmt.Sprintf("keycloak-auto-%06d", i) // this is the password too
```
agarciamontoro, line 188: Nit to make the code below a bit more clear
```suggestion
		username := fmt.Sprintf("keycloak-auto-%06d", i)
		password := username // we just use the same string for both
```

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,205 @@
...
+		email := username + "@test.mattermost.com"
+
+		_, err := sshc.RunCommand(fmt.Sprintf("%s/kcadm.sh create users -r %s -s username=%s -s enabled=true -s email=%s", keycloakBinPath, t.config.ExternalAuthProviderSettings.KeycloakRealmName, username, email))
```
agarciamontoro, line 191: Oh wow there is no way to create more than one user at a time? :sob:

deployment/config.go
```diff
@@ -179,6 +183,50 @@ type ExternalBucketSettings struct {
 	AmazonS3SSE             bool   `default:"false"`
 }
 
+// ExternalAuthProviderSettings contains the necessary data
+// to configure an external auth provider.
+type ExternalAuthProviderSettings struct {
+	// InstanceCount is the number of keycloak instances to deploy. It should be 0 or 1.
+	// If set to 0, the deployment will not include a keycloak instance.
+	InstanceCount int `default:"0" validate:"range:[0,1]"`
```
agarciamontoro, line 191: Should we use an `Enable bool` instead of `InstanceCount int` if we're only allowing one instance?

loadtest/user/userentity/actions.go
```diff
@@ -11,28 +11,106 @@ import (
...
+	switch ue.config.AuthenticationType {
+	case AuthenticationTypeOpenID:
+		if err := ue.authOpenID(authOpenIDSignup); err != nil {
+			return fmt.Errorf("error while signing up using OpenID: %w", err)
+		}
+
+		newUser, _, err = ue.client.GetUserByUsername(context.Background(), username, "")
+		if err != nil {
+			return fmt.Errorf("error while getting user by username: %w", err)
+		}
```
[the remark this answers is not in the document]
agarciamontoro: I wrote this before reading authOpenID, when I understood that no, that's not possible. So we'll have to live with this one, no problem.

agnivade:
Is this ready to review now?

fmartingr:
> Is this ready to review now?

I have to address Alejandro's comments and make some changes from the work in the LT environment. I will assign you again when that's done.

deployment/terraform/assets/cluster.tf
```diff
@@ -695,3 +695,139 @@ resource "null_resource" "s3_dump" {
     command = "aws --profile ${var.aws_profile} s3 cp ${var.s3_bucket_dump_uri} s3://${aws_s3_bucket.s3bucket[0].id} --recursive"
   }
 }
+
```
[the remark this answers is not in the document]
fmartingr: I agree 😅

deployment/terraform/assets/variables.tf
```diff
@@ -113,6 +113,48 @@ variable "block_device_sizes_elasticsearch" {
...
 
+# Keycloak variables
+variable "block_device_sizes_keycloak" {
+  type    = number
+  default = 10
+}
+
+variable "keycloak_instance_count" {
+  type    = number
+  default = 1
```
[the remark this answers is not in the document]
fmartingr: Yeah, this is a leftover. Defaults are set in `config.go` to `0`.

deployment/config.go
```diff
@@ -179,6 +183,50 @@ type ExternalBucketSettings struct {
...
+	// GenerateUsersCount is the number of users to generate in the keycloak instance.
+	GenerateUsersCount int `default:"0" validate:"range:[0,)"`
+	// KeycloakRealmName is the name of the realm to be used in Mattermost. Must exist in the keycloak instance.
+	KeycloakRealmName string `default:"mattermost"`
+	// KeycloakClientID is the client id to be used in Mattermost from the above realm.
+	// Must exist in the keycloak instance
+	KeycloakClientID string `default:"mattermost-openid"`
+	// KeycloakClientSecret is the client secret from the above realm to be used in Mattermost.
+	// Must exist in the keycloak instance
+	KeycloakClientSecret string `default:"qbdUj4dacwfa5sIARIiXZxbsBFoopTyf"`
```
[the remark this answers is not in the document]
fmartingr: Yes, this is the data for the realm embedded with the load test if no custom realm is provided.

api/agent.go
```diff
@@ -373,7 +373,7 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 
 	err = createCustomEmoji(config)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("error creating custom emoji from config: %w", err)
```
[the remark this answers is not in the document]
fmartingr: I was getting crazy because I got an error and I didn't knew where it came from... 😓

loadtest/user/userentity/actions.go
```diff
@@ -11,28 +11,106 @@ import (
...
+	if loginResponse.StatusCode != http.StatusOK {
+		return fmt.Errorf("login failed with status code %d", loginResponse.StatusCode)
+	}
+
+	cookies := jar.Cookies(loginResponse.Request.URL)
+	for _, cookie := range cookies {
+		if cookie.Name == "MMAUTHTOKEN" {
+			ue.client.SetToken(cookie.Value)
+		}
+	}
```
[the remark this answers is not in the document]
fmartingr: Oh _shirt_. I keep forgetting basic things...

fmartingr:
> * Do we need to expose the settings for Keycloak's DB? I'd lean towards being more opinionated whenever possible so that we expose less settings. Given that we don't want to monitor/load-test this server, can we use a one-size-fits-all instance type? Maybe not only for the DB but also for the Keycloak server? Or do the specs differ a lot when using more users? In any case, I think we can at least hardcode the usage of mysql or postgres here.

> * Why do the docs say that production mode is not implemented? Is there anything missing? Also: do we really need the distinction? I guess development mode is just cheaper to deploy because of the lack of DB, right?

Keycloak provides different commands for development/production servers. `start-dev` starts the server into development mode using a local [h2](https://www.h2database.com/html/main.html) database and it's less picky with the server configuration, while production mode requires https and a proper database. If h2 is anything like sqlite my guess is that we should be able to use it to log in all the users in dev-mode so all code for production mode can be removed, but I'd rather do a test run first in case the database I/O is a bottleneck.

> * Could you add some docs on the logic around the prefix for the usernames? The overwriting of the users file path was a bit confusing for me at first.

Yeah, I'm unsure about that part. I wanted to do it more automagic for the user but it breaks basic flows and I don't like it. I will remove the user generation part because is very slow. It's better and faster to manually generate an import file.

deployment/terraform/assets/cluster.tf
```diff
@@ -695,3 +695,139 @@ resource "null_resource" "s3_dump" {
...
+  engine                       = var.keycloak_db_instance_engine
+  apply_immediately            = true
+  auto_minor_version_upgrade   = false
+  performance_insights_enabled = false
+  db_parameter_group_name      = length(var.keycloak_db_parameters) > 0 ? "${var.cluster_name}-keycloak-db-pg" : ""
+}
+
+resource "aws_db_parameter_group" "keycloak_db_params_group" {
+  name   = "${var.cluster_name}-keycloak-db-pg"
+  family = var.keycloak_db_instance_engine == "aurora-mysql" ? "aurora-mysql8.0" : "aurora-postgresql14"
```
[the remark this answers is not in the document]
fmartingr: I'm going to fix it to psql for now, until we decide if remove it or not.

deployment/terraform/keycloak.go
fmartingr: We could call the API I guess, but there's no guarantee it would be faster.

deployment/terraform/keycloak.go
fmartingr: Yeah, I see that hapenning if that's an usual way of using the load test. Alternatively you could remove the file and the keycloak database to start from scratch.

loadtest/user/userentity/actions.go
```diff
@@ -11,28 +11,120 @@ import (
...
 	}
 
-	newUser.Password = password
-	return ue.store.SetUser(newUser)
+	client := &http.Client{
+		Jar: jar,
+	}
+
+	// make a request to the openid login page of mattermost
+	resp, err := client.Get(ue.client.URL + "/oauth/openid/" + string(action))
```
agarciamontoro, line 84: Should this be login/signup?
```suggestion
	// make a request to the openid login/signup page of mattermost
	resp, err := client.Get(ue.client.URL + "/oauth/openid/" + string(action))
```

agarciamontoro (changes requested):
Again, incredible work, very thorough and well documented, thanks! I left some more comments below (99% of them nits, sorry)

I also wanted to ask you for a summary of what assets we'll need once this gets in. If I'm not mistaken:
- DB dump of Keycloak (where is this? are we planning to upload it to S3 as the MM DB dump?)
- Realm file (is this really needed? or is it ok using the default one?)
- Anything else?

Thanks for all your work!

api/agent.go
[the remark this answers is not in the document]
agnivade: Guilty as charged :)

api/agent.go
```diff
@@ -395,15 +396,30 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 			username = creds[id].username
 			email = creds[id].email
 			password = creds[id].password
+
+			// Check if the user has a custom authentication type. Custom authentication types are
+			// specified by prepending the username with the authentication type followed by a colon.
+			// Example: "openid:user1@test.mattermost.com user1password"
+			if strings.Contains(username, ":") {
+				usernameParts := strings.Split(username, ":")
```
agnivade, line 404: Can be combined to `if parts := strings.Split(username, ":"); len(parts) > 1`

api/agent.go
```diff
@@ -395,15 +396,30 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
+				authenticationType = usernameParts[0]
+				username = usernameParts[1]
+
+				// Fix the email as well
+				emailParts := strings.Split(email, ":")
+				email = emailParts[1]
```
agnivade, line 410: Why is this needed? I thought just the username would become like "openid:user1"? The email should be untouched right? Also, we need to do length validation of the slice here to prevent panics.

deployment/config.go
```diff
@@ -179,6 +183,44 @@ type ExternalBucketSettings struct {
 	AmazonS3SSE             bool   `default:"false"`
 }
 
+// ExternalAuthProviderSettings contains the necessary data
+// to configure an external auth provider.
+type ExternalAuthProviderSettings struct {
+	// Enabled is set to true if the external auth provider should be enabled.
+	Enabled bool `default:"false"`
+	// DevelopmentMode is set to true if the keycloak instance should be started in development mode.
+	DevelopmentMode bool `default:"true"`
```
agnivade, line 192: Should it not be "false" by default? The default config is always aimed towards production usage.

deployment/config.go
```diff
@@ -179,6 +183,44 @@ type ExternalBucketSettings struct {
...
+	// An optional URI to a keycloak database dump file to be uploaded on environment
+	// creation.
+	// The file is expected to be the h2 folder gzip compressed.
+	// This can also point to a local file if prefixed with "file://".
+	// In such case, the dump file will be uploaded to the app servers.
+	KeycloakDBDumpURI string `default:""`
+	// GenerateUsersCount is the number of users to generate in the keycloak instance.
+	GenerateUsersCount int `default:"0" validate:"range:[0,)"`
+	// KeycloakRealmName is the name of the realm to be used in Mattermost. Must exist in the keycloak instance.
+	// Is used when creating users and to properly set the OpenID configuration in Mattermost.
```
agnivade, line 214: `It is used .. `. Common translation mistake :)

deployment/terraform/agent.go
```diff
@@ -129,6 +129,27 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent) error {
...
+			ip := output.Instances[0].PrivateIP
+			if output.HasProxy() {
+				ip = output.Proxy.PrivateIP
+			}
+
+			proxyHost := fmt.Sprintf("%s %s\n", ip, t.config.SiteURL)
+			appHostsFile := fmt.Sprintf(appHosts, proxyHost)
+
+			batch = append(batch, uploadInfo{srcData: appHostsFile, dstPath: "/etc/hosts", msg: "Updating /etc/hosts to point to the correct IP"})
+		}
```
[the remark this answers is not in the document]
agnivade: I think we've had a discussion here: https://community.mattermost.com/core/pl/1944im15z3f1pn9thb647kdi5y. I am 0/5 as long as it doesn't break permalinks.

deployment/terraform/create.go
```diff
@@ -678,6 +685,19 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
...
 
+	if t.output.HasKeycloak() {
+		keycloakScheme := "http"
+		if !t.config.ExternalAuthProviderSettings.DevelopmentMode {
+			keycloakScheme = "https"
+		}
+
+		cfg.OpenIdSettings.Enable = model.NewBool(true)
+		cfg.OpenIdSettings.ButtonText = model.NewString("Keycloak Login")
+		cfg.OpenIdSettings.DiscoveryEndpoint = model.NewString(keycloakScheme + "://" + t.output.KeycloakServer.PublicDNS + ":8080/realms/" + t.config.ExternalAuthProviderSettings.KeycloakRealmName + "/.well-known/openid-configuration")
```
agnivade, line 696: I see this `8080` being duplicated in a lot of places. Not for now, but later we'd want to move this to a config and use that value instead.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+		command = "start"
+	}
+
+	sshc, err := extAgent.NewClient(t.output.KeycloakServer.PublicIP)
+	if err != nil {
+		return err
+	}
+
+	// Check if the keycloak database exists
+	result, _ := sshc.RunCommand(`sudo -iu postgres psql -l | grep keycloak 2> /dev/null`)
```
agnivade, line 38: Why this `2> /dev/null`? Why not simply pipe to `wc -l`? Also, please check the error here.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+	command := "start-dev"
+	extraArguments := []string{}
+
+	if !t.config.ExternalAuthProviderSettings.DevelopmentMode {
+		command = "start"
+	}
+
+	sshc, err := extAgent.NewClient(t.output.KeycloakServer.PublicIP)
+	if err != nil {
+		return err
```
agnivade, line 34: No need to add error context here? :P

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	// Check if the keycloak database exists
+	result, _ := sshc.RunCommand(`sudo -iu postgres psql -l | grep keycloak 2> /dev/null`)
+	if len(result) == 0 {
+		mlog.Info("keycloak database not found, creating it and associated role")
+		_, err = sshc.RunCommand(`sudo -iu postgres psql <<EOSQL
+		CREATE USER keycloak WITH PASSWORD 'mmpass';
+		CREATE DATABASE keycloak OWNER keycloak;
+		GRANT ALL PRIVILEGES ON DATABASE keycloak TO keycloak;
```
agnivade, line 44: Please move this to a .sql file under assets, copy it to the keycloak node via provisioner like:

```
provisioner "file" {
    source      = var.mattermost_license_file
    destination = "/home/ubuntu/mattermost.mattermost-license"
  }
```

And use `psql -f` to execute from the file.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+			err := t.IngestKeycloakDump()
+			if err != nil {
+				return fmt.Errorf("failed to ingest keycloak dump: %w", err)
+			}
+		}
+	}
+
+	// Check if we should use a custom dump, a custom realm file or the default one
+	if t.config.ExternalAuthProviderSettings.KeycloakRealmFilePath != "" {
+		_, err := sshc.UploadFile(t.config.ExternalAuthProviderSettings.KeycloakRealmFilePath, "/opt/keycloak/keycloak-"+t.config.ExternalAuthProviderSettings.KeycloakVersion+"/data/import/mattermost-realm.json", true)
```
agnivade, line 61: The destination realm file is copied in two places. Let's consolidate into a single variable.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+		"KEYCLOAK_ADMIN_PASSWORD=" + t.config.ExternalAuthProviderSettings.KeycloakAdminPassword,
+		// Ensure Java JVM has enough memory for large imports
+		"JAVA_OPTS=-Xms1024m -Xmx2048m",
+		// Logging
+		"KC_LOG_FILE=/opt/keycloak/keycloak-" + t.config.ExternalAuthProviderSettings.KeycloakVersion + "/data/log/keycloak.log",
+		"KC_LOG_FILE_OUTPUT=json",
+		// Database
+		"KC_DB_POOL_MIN_SIZE=20",
+		"KC_DB_POOL_INITIAL_SIZE=20",
+		"KC_DB_POOL_MAX_SIZE=200",
```
agnivade, line 96: Question: would we need to modify this to support 100K users? If so, would it make sense to move it to config? (Out of scope for this PR obviously).

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	// Upload keycloak.env file
+	_, err = sshc.Upload(strings.NewReader(strings.Join(keycloakEnvFileContents, "\n")), "/etc/systemd/system/keycloak.env", true)
+	if err != nil {
+		return fmt.Errorf("failed to upload keycloak env file: %w", err)
+	}
+
+	// Parse keycloak service file template
+	tmpl, err := template.New("keycloakServiceFile").Parse(string(keycloakServiceFileContents))
```
agnivade, line 111: Is the `string()` really needed? I don't see it in other places where templates are used.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	// Install systemd service
+	_, err = sshc.Upload(&buf, "/etc/systemd/system/keycloak.service", true)
+	if err != nil {
+		return fmt.Errorf("failed to upload keycloak service file: %w", err)
+	}
+
+	// Enable and start service
+	_, err = sshc.RunCommand("sudo systemctl enable keycloak")
```
agnivade, line 134: Don't you need to do a `sudo systemctl daemon-reload` before enabling? Usually we do that for other services in the test.

deployment/terraform/keycloak.go
agnivade: +1 to this. We should do a quick check to ensure that # of users in keycloak file is equal to # users in users.txt file. And if possible, have a way to add new users, or fail clearly in the logs. But importantly, there should never be a case of some users partially logging in via keycloak and some via email/pass.

deployment/terraform/metrics.go
```diff
@@ -157,6 +157,12 @@ func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent) error {
 		}
 	}
 
+	if t.output.HasKeycloak() {
+		host := "keycloak"
+		keycloakTargets = append(keycloakTargets, fmt.Sprintf("%s:8080", host))
+		hosts += fmt.Sprintf("%s %s\n", t.output.KeycloakServer.PrivateIP, host)
+	}
+
```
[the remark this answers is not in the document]
agnivade: This seems wrong. 8080 is the port of the keycloak server. Is it also the same port where it exports metrics? That'd be weird.

deployment/terraform/strings.go
```diff
@@ -48,6 +48,9 @@ scrape_configs:
   - job_name: loadtest
     static_configs:
         - targets: [%s]
+  - job_name: keylcoak
```
agnivade, line 51: typo

loadtest/user/userentity/user.go
```diff
@@ -33,12 +33,19 @@ type UserEntity struct {
 	wsServerSeq int64
 }
 
+const (
+	AuthenticationTypeMattermost = "mattermost"
```
agnivade, line 37: Should this be `AuthenticationTypeEmailPass`? 0/5

agnivade (commented):
Gave it a first pass. Looks to be in good shape. Will give it a deeper dive after that.

api/agent.go
fmartingr: Since the `users.txt` file works with `email password` lines, what this does is parse if theres a `:` in the email to check for the auth service, hence parsing: `authservice:email password`. But in order to properly set the email I need to discard the `authservice:` part.

deployment/config.go
fmartingr: I left it as `false` because the difference development vs production means. If I use the production option I have to disable some of the features it enables manually (like enforced hostnames, SSL, local cache). I could go that route but in the end we would have the same result but with more work. That is, unless we want to really run it in production mode with all security features enforced as they should be.

deployment/terraform/agent.go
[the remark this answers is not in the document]
fmartingr: Is there an straightforward way for me to check that permalinks are working?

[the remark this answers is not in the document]
agarciamontoro: Changing the agents host will not break permalinks: the problem was with the app nodes trying to retrieve the data from the permalink. As long as we don't touch that, permalinks will keep working.

deployment/terraform/keycloak.go
fmartingr: I was using that to avoid errors here, but piping to `wc -l` is way more elegant. :thumbsup:

deployment/terraform/keycloak.go
fmartingr: You are right, though I have created environments from scratch that started the service correctly, but a `daemon-reload` should be required. Mhmmm. Maybe is only required for changes but not for new units? I need to check this.

deployment/terraform/metrics.go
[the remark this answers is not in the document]
fmartingr: I forgot the path, but yeah, [they expose metrics in the same port](https://www.keycloak.org/server/configuration-metrics).

api/agent.go
agnivade: I thought the authservice is just a prefix for the username right? Or is it for both?

deployment/terraform/keycloak.go
agnivade: Ok yeah that's a good point. I guess we will see in our big test.

agnivade (approved):
Fantastic work! 50 commits! :scream:

agnivade:
Blocking merge because I want to test this out manually by myself.

deployment/terraform/keycloak.go
fmartingr: Changed it to create the number of users setup in the configuration. If the number is higher it will create more users accordingly, using the number of lines in the `users.txt` file as the source. It will skip user creation if the number of lines is the same or higher than the expected number of users.

api/agent.go
fmartingr: It's used in the users txt file only, but ignored for login.

Unsure how the `:` would affect login in, and I selected that symbol so we easily know in code if there's a custom authservice setup. This could be changed to use `openid-username-00001` and check if the first part split by `-` matches an authservice but it seemed more complicated in my head. Would love your fresh eyes input on this.

deployment/terraform/keycloak.go
fmartingr: Mhmmm, how do I keep this in assets and then assign it to a variable if it's bundled in the binary?

deployment/config.go
```diff
@@ -179,6 +188,43 @@ type ExternalBucketSettings struct {
...
+// to configure an external auth provider.
+type ExternalAuthProviderSettings struct {
+	// Enabled is set to true if the external auth provider should be enabled.
+	Enabled bool `default:"false"`
+	// DevelopmentMode is set to true if the keycloak instance should be started in development mode.
+	DevelopmentMode bool `default:"true"`
+	// KeycloakVersion is the version of keycloak to deploy.
+	KeycloakVersion string `default:"24.0.2"`
+	// KeycloakInstanceType is the type of the EC2 instance for keycloak.
+	InstanceType string `default:"c5.xlarge"`
```
agnivade, line 201: We are moving everything to the c7 series. Let's do it here as well.

deployment/terraform/agent.go
```diff
@@ -24,7 +24,9 @@ func (t *Terraform) generateLoadtestAgentConfig() (*loadtest.Config, error) {
 		return nil, err
 	}
 	url := t.output.Instances[0].PrivateIP + ":8065"
-	if t.output.HasProxy() {
+	if t.config.SiteURL != "" {
```
agnivade, line 27:
```go
	if t.config.SiteURL != "" {
		if t.output.HasProxy() {
			url = t.config.SiteURL
		} else {
			url = t.config.SiteURL + ":8065"
		}
	} else {
		if t.output.HasProxy() {
			url = t.output.Proxy.PrivateIP
		} else {
			url = t.output.Instances[0].PrivateIP + ":8065"
		}
	}
```

agnivade (approved):
New commits look good to me. The only small thing is that let's change the default keycloak instance spec to `c7i` series. Thanks again for all the hard work!

api/agent.go
```diff
@@ -395,15 +396,30 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
...
+			// Example: "openid:user1@test.mattermost.com user1password"
+			if usernameParts := strings.Split(username, ":"); len(usernameParts) > 1 {
+				authenticationType = usernameParts[0]
+				username = usernameParts[1]
+
+				// Fix the email as well
+				if emailParts := strings.Split(email, ":"); len(emailParts) > 1 {
+					email = emailParts[1]
+				}
+			}
```
streamer45, line 411: I think this could become a util function.

loadtest/user/userentity/user.go
streamer45: +1

loadtest/user/userentity/actions.go
```diff
@@ -11,28 +11,120 @@ import (
...
+			ue.client.SetToken(cookie.Value)
+
+			me, _, err := ue.client.GetMe(context.Background(), "")
+			if err != nil {
+				return fmt.Errorf("error while getting user: %w", err)
+			}
+
+			if err := ue.store.SetUser(me); err != nil {
+				return fmt.Errorf("error while setting user: %w", err)
+			}
```
streamer45, line 121: Do we need to do this here? I'd expect this method to purely authenticate (e.g. set the token) and then deal with everything else in `SignUp/Login`.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,356 @@
...
+		)
+		if err != nil {
+			return fmt.Errorf("failed to upload default keycloak realm file: %w", err)
+		}
+	}
+
+	// Values for the keycloak.env file
+	keycloakEnvFileContents := []string{
+		// Enable health endpoints
+		"KC_HEALTH_ENABLED=true",
```
streamer45, line 100: Maybe we could use a template file as we do with other plaintext assets.

deployment/terraform/keycloak.go
```diff
@@ -0,0 +1,356 @@
...
+// number of lines in the file.
+// If the file has more users than expected, it will log and skip the user creation.
+func (t *Terraform) populateKeycloakUsers(sshc *ssh.Client) error {
+	keycloakBinPath := "/opt/keycloak/keycloak-" + t.config.ExternalAuthProviderSettings.KeycloakVersion + "/bin"
+	usersTxtPath := t.getAsset("keycloak-users.txt")
+	startNumber := 1
+
+	// Check if users file exists and has the expected number of users. If the file has less users than expected,
+	// we will start creating users from the next number, otherwise we will skip user creation.
+	if _, err := os.Stat(usersTxtPath); err == nil || os.IsExist(err) {
```
streamer45, line 235: Why would `Stat` return an error if the file exists?

streamer45 (commented):
Very quickly skimmed through this. Looks great overall. Left some comments.

agarciamontoro (approved):
Agian, amazing work, @fmartingr. Nothing else to add. Thank you!

api/agent.go
fmartingr: I'm going to make a note of this but refrain from modifying too much logic until this is merged now that we have this working. I can address the rest of the comments in individual PRs later.

deployment/terraform/dump.go
```diff
@@ -151,13 +140,126 @@ func (t *Terraform) IngestDump() error {
...
+
+	appClients := make([]*ssh.Client, len(output.Instances))
+	for i, instance := range output.Instances {
+		client, err := extAgent.NewClient(instance.PublicIP)
+		if err != nil {
+			return fmt.Errorf("error in getting ssh connection %w", err)
+		}
+		defer client.Close()
+		appClients[i] = client
+	}
```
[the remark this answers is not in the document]
fmartingr: Good catch.

deployment/terraform/keycloak.go
fmartingr: I think this was supposed to be `os.IsNotExist` in a previous apporach or something else with a different logic and I just inherited it here, this doesn't make any sense at the moment.

agnivade:
Merge it at your convenience. :smile:

loadtest/user/userentity/actions.go
fmartingr: There's a mix of logic here, is either I get the user and set it here, or I get the user by username later and set it in the other method. I've set a TODO to change the logic in a separate PR since I don't want to compromise or delay the test we're going to run today.

fmartingr:
> Merge it at your convenience. 😄

@streamer45 I've put comments on heavy logic changes that you mentioned so I can address those on separate PRs so I can merge this for now since it's getting too big and not easy to review. Thank you for reviewing!

PR #722, opened 2024-04-03
Acknowledge websocket POSTED events when they should be

devinbinnie:
#### Summary
Adding to the load tester the changes needed to test [this PR](https://github.com/mattermost/mattermost/pull/26604).
Marking as Draft for now since I think we might want to revisit how we want to queue the ACKs to send back.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57006
https://mattermost.atlassian.net/browse/MM-57067

### DIFF
--- a/loadtest/control/simulcontroller/websocket.go
+++ b/loadtest/control/simulcontroller/websocket.go
@@ -36,10 +36,9 @@ func (c *SimulController) wsEventHandler(wg *sync.WaitGroup) {
 				break
 			}
 
-			if err := c.user.PostedAck(post.Id, "success", "", ""); err != nil {
-				c.status <- c.newErrorStatus(fmt.Errorf("PostedAck failed: %w", err))
-				break
-			}
+			go func() {
+				c.user.PostedAck(post.Id, "success", "", "")
+			}()
 
 			cm, _ := c.user.Store().ChannelMember(post.ChannelId, c.user.Store().Id())
 			if cm.UserId != "" {
--- a/loadtest/control/simulcontroller/websocket.go
+++ b/loadtest/control/simulcontroller/websocket.go
@@ -36,9 +36,11 @@ func (c *SimulController) wsEventHandler(wg *sync.WaitGroup) {
 				break
 			}
 
-			go func() {
-				c.user.PostedAck(post.Id, "success", "", "")
-			}()
+			if c.user.Store().Id() != post.UserId {
+				go func() {
+					c.user.PostedAck(post.Id, "success", "", "")
+				}()
+			}
 
 			cm, _ := c.user.Store().ChannelMember(post.ChannelId, c.user.Store().Id())
 			if cm.UserId != "" {
--- a/loadtest/control/simulcontroller/websocket.go
+++ b/loadtest/control/simulcontroller/websocket.go
@@ -36,7 +36,7 @@ func (c *SimulController) wsEventHandler(wg *sync.WaitGroup) {
 				break
 			}
 
-			if c.user.Store().Id() != post.UserId {
+			if ack, ok := ev.GetData()["should_ack"]; ok && ack.(bool) {
 				go func() {
 					c.user.PostedAck(post.Id, "success", "", "")
 				}()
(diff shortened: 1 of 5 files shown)


loadtest/control/simulcontroller/websocket.go
```diff
@@ -36,6 +36,12 @@ func (c *SimulController) wsEventHandler(wg *sync.WaitGroup) {
 				break
 			}
 
+			if ack, ok := ev.GetData()["should_ack"]; ok && ack.(bool) {
```
agarciamontoro, line 39: Double-checking: What should we do if there's no `"should_ack"` key in the event data? Just skip?

loadtest/control/simulcontroller/websocket.go
```diff
@@ -36,6 +36,12 @@ func (c *SimulController) wsEventHandler(wg *sync.WaitGroup) {
...
+				go func() {
+					c.user.PostedAck(post.Id, "success", "", "")
```
agarciamontoro, line 41: We should log the potential error returned by `PostedAck`

agarciamontoro (changes requested):
Looks great! Just a couple of nits

loadtest/control/simulcontroller/websocket.go
devinbinnie: Correct. we only want to actually send the ACK back if the posted event has this flag, if not we shouldn't do anything.

loadtest/control/simulcontroller/websocket.go
devinbinnie: Good call. Do I need to just pass the error to the `c.status` channel if there is one?

loadtest/control/simulcontroller/websocket.go
```diff
@@ -36,6 +36,12 @@ func (c *SimulController) wsEventHandler(wg *sync.WaitGroup) {
 				break
 			}
 
+			if ack, ok := ev.GetData()["should_ack"]; ok && ack.(bool) {
+				go func() {
```
streamer45, line 40: As I mentioned in channel I see this as a decent workaround but I'd prefer if we fixed the root issue rather than starting a pattern of avoiding the deadlock. My suggestion would be to make `ue.dataChan` a buffered channel and do a non-blocking write in `UserEntity.PostedAck()` with the default case (channel full) returning an error (to be logged).

loadtest/control/simulcontroller/websocket.go
agarciamontoro: Yup!

loadtest/control/simulcontroller/websocket.go
agarciamontoro: Got it!

loadtest/control/simulcontroller/websocket.go
devinbinnie: Couple questions:
1) How big of a buffer do we want?
2) If the buffer is full, will it just lose events? If so, this might need that tracking of ACKs might not be super accurate, which isn't necessarily a problem for load testing, but isn't a great simulation.

devinbinnie: So one thing to note, I think it's already sending it through the channel:
https://github.com/mattermost/mattermost-load-test-ng/pull/722/files#diff-d66de5fdcf416b8ffb207a36b55fca5b16554291b2a06a94e45d7372cc3272b3R307

Is the buffering something that would fix this? Given that it was dead-locking before, is calling this method causing a blocking call or something?

streamer45: Problem is that sending to `dataChan` is blocking right now and `dataChan` is unbuffered which causes the deadlock. We need to address both things.

streamer45: If we only make it non-blocking it will fail to send (chan full) as soon as there's a bit of extra concurrency. If we only make it buffered it could deadlock again if channel becomes full. I'd prefer to error than deadlock.

streamer45 (approved):
Thanks Devin, looks good :+1:

agarciamontoro (approved):
Thank you, Devin, this looks great!

PR #723, opened 2024-04-04
Bumping instance specs to new series

agnivade:


agarciamontoro (commented):
Neat! Can you change the `default` tag in the struct fields as well?

agarciamontoro (approved):
Neat, thanks!

streamer45 (approved):
Nice, and some great improvement overall :rocket:

config/deployer.sample.json
```diff
@@ -6,9 +6,9 @@
   "ClusterVpcID": "",
   "ClusterSubnetID": "",
   "AppInstanceCount": 1,
-  "AppInstanceType": "c5.xlarge",
+  "AppInstanceType": "c7i.xlarge",
   "AgentInstanceCount": 2,
-  "AgentInstanceType": "c5.xlarge",
+  "AgentInstanceType": "c7i.xlarge",
```
streamer45, line 11: @agarciamontoro We may be able to review the long standing `2000` cap per instance. Not urgent of course but something that may be valuable because that value was decided a long way back when the load profile was very different.

agnivade: It's still very much that number. Quite recently I just used a larger number on a whim and it started to fail. :P

agnivade: Yeah I think 2000 is still the right number.

PR #724, opened 2024-04-08
MM-57355: Moving to a test service env

agnivade:
https://mattermost.atlassian.net/browse/MM-57355

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform"
 	"github.com/mattermost/mattermost-load-test-ng/logger"
+	"github.com/mattermost/mattermost/server/public/model"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


agnivade:
Apparently, this is all that was needed :man_shrugging: The only pending thing is that now we need a test license rather than a prod license going forward.

agarciamontoro (commented):
Really? This was all? :sob: Jesse will be very happy when he comes back.

What MM versions have you tested this on? I guess we should at least test with v8.1

agnivade:
I tested on 9.7, but I don't think anything really changed after the original PR. Do you think it's necessary to run with 8.1 again? (Especially given 9.5 is the latest ESR)

streamer45 (commented):
Changes look good but I have a question. How would customers use this going forward? I'd expect most of them to have real licenses so maybe we should consider an easy (and documented) way to override this if needed?

agnivade:
Hmm oh yeah good point. Are you thinking of exposing this whole thing from config.json? Or just document this to change the code as needed?

streamer45:
I think we'd want to prevent them from changing the code directly if possible so either a config knob or a way to override it through a CLI flag, whatever is easiest to implement. And the most important part would be a documentation update since it's a bit of a breaking change.

streamer45 (approved):
Thanks, looks good :+1:

agarciamontoro (changes requested):
We're technically inside the v8.1 ESR window still, so i'd love to get it tested there as well. It will probably work, but better make sure before merging.

agnivade:
Tested with 8.1. Works fine as expected.

agarciamontoro (approved):
Thank you for the additional testing!

PR #726, opened 2024-04-12
MM-57747: Upgrade nginx and optimize cache config

agnivade:
https://mattermost.atlassian.net/browse/MM-57747


deployment/terraform/create.go
```diff
@@ -426,16 +427,41 @@ func (t *Terraform) setupProxyServer(extAgent *ssh.ExtAgent) {
...
+		parts := strings.Split(t.config.ProxyInstanceType, ".")
+		if len(parts) > 1 {
+			// Hack alert. There can be other higher variants like 12xlarge or even metal. But we are keeping it simple for now.
+			switch parts[1] {
+			case "4", "8":
+				cacheObjects = "50m"
+				cacheSize = "16g" // Ideally we'd like half of the total server mem. But the mem consumption rarely exceeds 10G
+				// from my tests. So there's no point stretching it further.
+			}
+		}
```
streamer45, line 443: Would it make sense to make this a util? Feels a bit out of place in the whole terraform logic.

deployment/terraform/create.go
```diff
@@ -426,16 +427,41 @@ func (t *Terraform) setupProxyServer(extAgent *ssh.ExtAgent) {
 			backends += "server " + addr.PrivateIP + ":8065 max_fails=0;\n"
 		}
 
+		cacheObjects := "10m"
+		cacheSize := "3g"
+		// Extracting the instance class from the type.
+		// Usually they are of the form (m7/c7).(large/2xlarge/4xlarge/..)
+		parts := strings.Split(t.config.ProxyInstanceType, ".")
+		if len(parts) > 1 {
+			// Hack alert. There can be other higher variants like 12xlarge or even metal. But we are keeping it simple for now.
```
streamer45, line 436: I'd rather we failed, logged an error or something if we find anything we don't understand/expect. Mostly to avoid subtle bugs as we'll forget about this eventually.

deployment/terraform/strings.go
```diff
@@ -194,13 +193,13 @@ proxy_cache_use_stale timeout;
...
-const nginxSiteConfig = `
+const nginxSiteConfigTmpl = `
 upstream backend {
-%s
+{{.backends}}
   keepalive 256;
 }
 
-proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=mattermost_cache:10m max_size=3g inactive=120m use_temp_path=off;
+proxy_cache_path /var/cache/nginx levels=1:2 keys_zone=mattermost_cache:{{.cacheObjects}} max_size={{.cacheSize}} inactive=60m use_temp_path=off;
```
streamer45, line 202: @agarciamontoro An additional note to include `nginx` recommendations as part of the reference architecture.

streamer45 (approved):
Nice work, nothing blocking :+1:

deployment/terraform/create.go
agnivade: Hmm .. maybe?

deployment/terraform/create.go
agnivade: You mean you want to limit to only 2x/4x/8x instances? That sounds a bit overly restrictive.. :thinking:

agarciamontoro (approved):
Great work, thanks for the thorough investigation! Just left a comment below for a potential improvement, but we don't need to tackle it now.

deployment/terraform/create.go
agnivade: Yes indeed, I was thinking of doing something like this. I think after your ES PR goes in, which starts to use the aws client, I can easily do this as an OSF project.

deployment/terraform/strings.go
agnivade: I think it should be nginx + postgres config optimizations as well.

deployment/terraform/create.go
streamer45: Great idea!

agarciamontoro: Neat, thanks! :)

PR #727, opened 2024-04-13
Bump TCP timeout thresholds

agnivade:
A look at production instances show that 0.2 is the norm there.
This is with very low traffic <1Gbps. So it should be safe to
triple the limits when we expect traffic to be close to 8-9Gbps
at peak usage.

### DIFF
--- a/config/coordinator.sample.json
+++ b/config/coordinator.sample.json
@@ -72,7 +72,7 @@
         "Description": "Percentage of TCP timeouts in the proxy node",
         "Legend": "Percent",
         "Query": "100 * rate(node_netstat_TcpExt_TCPTimeouts{instance=~\"proxy:9100\"}[1m]) / (rate(node_netstat_TcpExt_TCPTimeouts{instance=~\"proxy:9100\"}[1m]) + node_netstat_Tcp_CurrEstab{instance=~\"proxy:9100\"})",
-        "Threshold": 0.2,
+        "Threshold": 0.6,
         "MinIntervalSec": 60,
         "Alert": true
       },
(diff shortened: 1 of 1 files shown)


agnivade:
From my observations in the tests, it keeps on increasing.

streamer45 (approved):
A couple of non-blocking comments from my side:

- Something to keep in mind is that using production values for networking details such as this one could be a bit misleading since our tests are designed to run in a controlled environment where latency is minimal (a few milliseconds) as client requests are all coming from the same region/cluster. Using real-world bounds could make these queries much less meaningful.
- I think we should open a follow-up ticket to understand further what's behind the timeouts. I am especially confused by the app nodes needing higher thresholds than the proxy node which on paper should be handling significantly more load both in terms of connections and throughput.

agnivade:
In that case, let's close this. The whole idea was to increase this based on real world data because we didn't really know what exactly should be a good threshold. I agree that we should open a follow-up ticket to understand further. But then this also means that with the current limits, 88K is a hard bound irrespective of however much we improve our app performance.

streamer45:
> In that case, let's close this. The whole idea was to increase this based on real world data because we didn't really know what exactly should be a good threshold. I agree that we should open a follow-up ticket to understand further. But then this also means that with the current limits, 88K is a hard bound irrespective of however much we improve our app performance.

I think it's reasonable to bump the threshold as long as the measured performance, especially on the client side is still acceptable (no noticeable issues). That said, this could also be a good opportunity to start exploring horizontal scaling at the proxy level while we figure out more on this (follow up investigation).

agnivade:
I haven't really tested the client side of things at peak scale. So I am not 100% sure. So perhaps we can defer this until we run future tests?

streamer45:
> I haven't really tested the client side of things at peak scale. So I am not 100% sure. So perhaps we can defer this until we run future tests?

Sure, that works :+1:

agnivade:
Re-opening this since after the latest 100K test, we can confirm that without the TCP thresholds, the app doesn't have any sort of disconnections or otherwise network issues. On the side, we will also be doing other app and load-test agent optimizations that should help the situation further.

agarciamontoro:
In the latest 100K test, both the proxy and the app nodes are above the proposed thresholds:
image
image

Do we want to maintain them like this or increase them further?

agnivade:
Oh good point there. I think let's start with 3x for now. The test was actually bounded, so it's also possible that even with the 3x thresholds, it would have slowly reached the target. But since we have more tunings planned, we can bump them higher later as well.

agarciamontoro (approved):
Good, let's go with these. Thanks!

PR #729, opened 2024-04-16
MM-56605: Create results file and print to file

DHaussermann:
#### Summary
This will create a the `results.txt` file and write the results blocks to it eliminating the need to manually copy from the console. 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56605

### DIFF
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -175,7 +175,6 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
 		// Call writeResults and handle any errors it returns
 		err := writeResults(output.Results, multiWriter)
 		if err != nil {
-			fmt.Println("Error:", err)
 			return fmt.Errorf("failed to write results: %w", err)
 		}
 
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -63,7 +63,7 @@ func getReportFilename(id int, res comparison.Result) string {
 	return fmt.Sprintf("report_%s.md", name)
 }
 
-func printResults(results []comparison.Result) {
+func writeToFile() {
 	var resultsFile *os.File
 
 	// Create or open the file for writing
@@ -73,6 +73,16 @@ func printResults(results []comparison.Result) {
 		return
 	}
 	defer resultsFile.Close()
+
+}
+
+func printResults(results []comparison.Result, resultsFile *os.File) {
+	_, err := fmt.Println("Error with file:")
+
+	if err != nil {
+		fmt.Println("Error:", err)
+		return
+	}
 
 	for i, res := range results {
 		_, err = fmt.Fprintf(resultsFile, "==================================================")
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -101,7 +101,7 @@ func printResults(results []comparison.Result, writer io.Writer) {
 			} else if ltRes.Config.Type == comparison.LoadTestTypeUnbounded {
 				fmt.Fprintf(writer, "  Supported Users: %d\n", ltRes.Status.SupportedUsers)
 			}
-			fmt.Printf("  Errors: %d\n", ltRes.Status.NumErrors)
+			fmt.Fprintf(writer, "  Errors: %d\n", ltRes.Status.NumErrors)
 		}
 		fmt.Fprintf(writer, "==================================================\n\n")
 	}
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -67,8 +67,8 @@ func writeResults(results []comparison.Result, writer io.Writer) error {
 	var content string
 
 	for i, res := range results {
-		content += ("==================================================")
-		content += ("Comparison result:")
+		content += "=================================================="
+		content += "Comparison result:"
 		content += fmt.Sprintf("Report: %s\n", getReportFilename(i, res))
 		content += fmt.Sprintf("Grafana Dashboard: %s\n", res.DashboardURL)
 		for _, ltRes := range res.LoadTests {
@@ -83,7 +83,7 @@ func writeResults(results []comparison.Result, writer io.Writer) error {
 			}
 			content += fmt.Sprintf("  Errors: %d\n", ltRes.Status.NumErrors)
 		}
-		content += ("==================================================\n\n")
+		content += "==================================================\n\n"
 	}
 	_, err := fmt.Fprintf(writer, content)
 	return err
@@ -162,7 +162,7 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("failed to write reports: %w", err)
 		}
 
-		//write to file here
+		//create the file
 		resultsFile, errResult := os.Create(filepath.Join(outputPath, "results.txt"))
 		if errResult != nil {
 			fmt.Println("Error:", errResult)
@@ -172,7 +172,12 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
 
 		multiWriter := io.MultiWriter(resultsFile, os.Stdout)
 
-		writeResults(output.Results, multiWriter)
+		// Call writeResults and handle any errors it returns
+		err := writeResults(output.Results, multiWriter)
+		if err != nil {
+			fmt.Println("Error:", err)
+			return fmt.Errorf("failed to write results: %w", err)
+		}
 
 		if archive {
 			if err := createArchive(outputPath, archivePath); err != nil {
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -64,28 +64,48 @@ func getReportFilename(id int, res comparison.Result) string {
 }
 
 func printResults(results []comparison.Result) {
+	var resultsFile *os.File
+
+	// Create or open the file for writing
+	resultsFile, err := os.Create("results.txt")
+	if err != nil {
+		fmt.Println("Error:", err)
+		return
+	}
+	defer resultsFile.Close()
+
 	for i, res := range results {
-		fmt.Println("==================================================")
-		fmt.Println("Comparison result:")
+		_, err = fmt.Fprintf(resultsFile, "==================================================")
+		if err != nil {
+			fmt.Println("Error:", err)
+			return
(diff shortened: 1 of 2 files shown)


DHaussermann:
@streamer45 The unit test on `comparison_test.go` now runs but there are 2 parts of the string compare that are failing and I'm not sure how to fix them. 

1. Line 43 fails in 2 places. It seems that and actual line return and `/n` cannot be used interchangeably but I have tried this both ways and cannot make the expected match
2. The end of the string compare is getting `:\n  Type: \n  DB Engine:` instead of ` Errors: 2\` I'm trying to pass into the buffer and I don't know why. 

Please advise if you have time to take a look.

streamer45:
@DHaussermann Some problems I see:

1. (bug) We are missing a `fmt.Fprintf` call in https://github.com/mattermost/mattermost-load-test-ng/blob/d65c7ec3aa1b852de24ea7c439602af60b2edc23/cmd/ltctl/comparison.go#L104 as it's still using `fmt.Printf`.
2. You should provide a second result because we always expect that array to have 2 elements (remember the `[2]`).
3. There's some indentation expected in the output, which we need to match.

DHaussermann:
@streamer45 I believe my last commit solves 1. and 3. from the feedback above.

For 2. I'm not entirely sure what type of element is expected here, Would there be a second element following this one? A test2 for example? 
https://github.com/mattermost/mattermost-load-test-ng/blob/a38abb562eed996f5c1a75672b07037a18552733/cmd/ltctl/comparison_test.go#L22-L34

Would solving this solve the extra content in the expected that I cannot account for?  I still see the following i the expected `n:\n  Type: \n  DB Engine: \n  Errors: 0\n` but I'm not clear on why. 

When ever you have some time please review again.

streamer45:
@DHaussermann I think we are very close. So we need to add the additional `comparison.LoadTestResult` to the array since we are always comparing 2 test runs, not just one. Something like this:

```go
	// Prepare test data
	results := []comparison.Result{
		{
			Report:       "Sample Report",
			DashboardURL: "http://example.com/dashboard",
			LoadTests: [2]comparison.LoadTestResult{
				{
					Label: "Test1",
					Config: comparison.LoadTestConfig{
						Type:     "bounded",
						DBEngine: "postgres",
						NumUsers: 100,
						Duration: "10m",
					},
					Status: coordinator.Status{
						SupportedUsers: 80,
						NumErrors:      2,
					},
				},
				{
					Label: "Test2",
					Config: comparison.LoadTestConfig{
						Type:     "bounded",
						DBEngine: "postgres",
						NumUsers: 100,
						Duration: "10m",
					},
					Status: coordinator.Status{
						SupportedUsers: 80,
						NumErrors:      2,
					},
				},
			},
		},
	}
```

 Also, this looks off as the test name should go before the colon:

https://github.com/mattermost/mattermost-load-test-ng/blob/a38abb562eed996f5c1a75672b07037a18552733/cmd/ltctl/comparison_test.go#L44

cmd/ltctl/comparison_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"bytes"
+	"testing"
+
+	"github.com/mattermost/mattermost-load-test-ng/comparison"
+	"github.com/mattermost/mattermost-load-test-ng/coordinator"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPrintResults(t *testing.T) {
```
agarciamontoro, line 12: Great test!

agarciamontoro (changes requested):
Great work! I've added some comments below for a couple of improvements I think we can make. Let me summarize them here:
1. Rename the function so that its name doesn't say `print` but `write`, since we're now writing to a file.
2. Make the function return an error.
3. Regarding the errors returned by `fmt.Fprintf`, we should check them every time we call that function. But that would become quite verbose very easily, since we call it a lot of times. I think in this case we can do something else: as the content we want to write is not large, we can build a string with `fmt.Sprintf` and then call `fmt.Fprintf` just once, using the string created. It would be something like:
```go
content := fmt.Sprintf("===\n")
content += fmt.Sprintf("thing1\n")
content += fmt.Sprintf("thing2\n")

// ... keep adding stuff to content

// At the very end of the function, call Fprintf once and return its error
_, err := fmt.Fprintf(writer, content)
return err
```
4. Instead of creating the file as `results.txt`, we need to prepend the `outputPath` setting, so we should do something like `os.Create(filepath.Join(outpath, "results.txt"))` instead.

Again, great first pass, thank you for working on this!

cmd/ltctl/comparison.go
```diff
@@ -161,16 +179,27 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
...
-		printResults(output.Results)
-	}
 
-	if archive {
-		if err := createArchive(outputPath, archivePath); err != nil {
-			return fmt.Errorf("failed to create archive: %w", err)
+		//write to file here
+		resultsFile, errResult := os.Create("results.txt")
+		if errResult != nil {
+			fmt.Println("Error:", errResult)
```
agnivade, line 186: Return from here: `return fmt.Errorf("failed to create file: %w", err)`

agnivade (commented):
Same as Alejandro's other comments.

- Remove unnecessary code.
- Return error from `printResults` rather than print the error. Remember: logging should always be done at a higher level. Internal functions should simply return the error. This has 2 benefits: you can test your function easily, and you can add more context when you print the error.

agnivade:
Feel free to ask for a review when it's ready.

DHaussermann:
@agnivade and @agarciamontoro Thanks for your help with this! Please review again when you have a moment. I *think* I have addressed all the review feedback. 

Note, I solved the linter error to the best my ability. it did not like the use of `fmt.Sprintf` on strings that require no formatting and `fmt.Sprintf`could no longer be used because it returns too many values as there 1 extra for the error that I don't account for anymore. I don't know how better to solve this than to use neither.

cmd/ltctl/comparison.go
```diff
@@ -161,16 +161,28 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
...
+		resultsFile, errResult := os.Create(filepath.Join(outputPath, "results.txt"))
+		if errResult != nil {
+			fmt.Println("Error:", errResult)
+			return fmt.Errorf("failed to create file: %w", err)
+		}
+		defer resultsFile.Close()
+
+		multiWriter := io.MultiWriter(resultsFile, os.Stdout)
+
+		writeResults(output.Results, multiWriter)
```
agnivade, line 175: You have to handle the error here Dylan. Otherwise, there's not much benefit of returning the error :)

cmd/ltctl/comparison.go
```diff
@@ -63,30 +63,30 @@ func getReportFilename(id int, res comparison.Result) string {
...
 	for i, res := range results {
-		fmt.Println("==================================================")
-		fmt.Println("Comparison result:")
-		if res.Report != "" {
-			fmt.Printf("Report: %s\n", getReportFilename(i, res))
-		}
-		if res.DashboardURL != "" {
-			fmt.Printf("Grafana Dashboard: %s\n", res.DashboardURL)
-		}
+		content += ("==================================================")
```
agnivade, line 70: You can just remove the `(`, `)` and treat this like a normal string.

DHaussermann:
Hi @agnivade I have added error handling for `writeResults`. When you have a moment, please review and advised if this was done the way you would expect.

cmd/ltctl/comparison.go
```diff
@@ -161,16 +161,33 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
...
+			return fmt.Errorf("failed to create file: %w", err)
+		}
+		defer resultsFile.Close()
+
+		multiWriter := io.MultiWriter(resultsFile, os.Stdout)
+
+		// Call writeResults and handle any errors it returns
+		err := writeResults(output.Results, multiWriter)
+		if err != nil {
+			fmt.Println("Error:", err)
```
agnivade, line 178:
```suggestion
```

agnivade (approved):
The `fmt.Println` line can be removed. This is because when you are returning the error, it will anyways get logged at the higher level. So there is no benefit in both logging and returning an error.

cmd/ltctl/comparison.go
```diff
@@ -161,16 +161,32 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
...
 
-	if archive {
-		if err := createArchive(outputPath, archivePath); err != nil {
-			return fmt.Errorf("failed to create archive: %w", err)
+		//create the file
+		resultsFile, errResult := os.Create(filepath.Join(outputPath, "results.txt"))
+		if errResult != nil {
+			fmt.Println("Error:", errResult)
+			return fmt.Errorf("failed to create file: %w", err)
+		}
```
agarciamontoro, line 170: We can remove this `fmt.Println` line as well. Also, it seems we were not using the correct error in the `fmt.Errorf` call:
```suggestion
		//create the file
		resultsFile, errResult := os.Create(filepath.Join(outputPath, "results.txt"))
		if errResult != nil {
			return fmt.Errorf("failed to create file: %w", errResult)
		}
```

agarciamontoro (changes requested):
Great work! There's just one more `fmt.Println` we can remove and I think we're done!

cmd/ltctl/comparison.go
```diff
@@ -161,16 +161,31 @@ func RunComparisonCmdF(cmd *cobra.Command, args []string) error {
...
-	}
 
-	if archive {
-		if err := createArchive(outputPath, archivePath); err != nil {
-			return fmt.Errorf("failed to create archive: %w", err)
+		//create the file
+		resultsFile, errResult := os.Create(filepath.Join(outputPath, "results.txt"))
+		if errResult != nil {
+			return fmt.Errorf("failed to create file: %w", err)
+		}
```
agarciamontoro, line 169: I think my other suggestion to use the correct `errResult` was missed before!
```suggestion
		if errResult != nil {
			return fmt.Errorf("failed to create file: %w", errResult)
		}
```

DHaussermann:
Thanks for your patience with this @agarciamontoro. Yes there was silly error I think  I have now corrected.

agarciamontoro (approved):
Thanks, @DHaussermann! This was great, and thank you for your patience as well :)

PR #730, opened 2024-04-17
Use "status" as the posted ACK result instead of "result"

devinbinnie:
#### Summary
I had changed the server code to use `status` but forgot to change it here, see: https://github.com/mattermost/mattermost/pull/26799


streamer45 (approved):

agarciamontoro (approved):
Thanks for having this in mind! :)

PR #731, opened 2024-04-19
MM-57915: Clear licenses data when using ClusterIdentifier

agarciamontoro:
#### Summary
The SQL query to clear old licenses data from a DB dump was only executed when using the `DBDumpURI` setting, but not with `ClusterIdentifier`, which is the one we use for larger DB dumps.

I've tested it a couple of times and I didn't get the license error, which I was consistently getting before with the same deployment. Let's see if we can forget about this forever :crossed_fingers: 

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-57915


agnivade (approved):
Nice! Let's hope that the exorcism works this time!

streamer45 (approved):
Looks great!

PR #739, opened 2024-05-02
Allow multiple instances in ElasticSearch domains

agarciamontoro:
#### Summary
There was an artificial limitation on the number of instances allowed in the OpenSearch Service domain, which this PR removes. 
This, in turn, surfaced an error in the way we measure the progress when restoring snapshots, since we were considering all types of shards, while the ones we're interested in are only of type `SNAPSHOT` (see [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-recovery.html)). So we change that logic and the naming of the corresponding function and struct to better reflect what they do.

This introduced some weird behaviour regarding the load-balancing in AWS side, which we are tackling in [another ticket](https://mattermost.atlassian.net/browse/MM-58093).

#### Ticket Link
--


deployment/elasticsearch/elasticsearch.go
```diff
@@ -339,12 +339,15 @@ func (c *Client) IndicesRecovery(indices []string) ([]IndexShardRecovery, error)
 		return nil, fmt.Errorf("unable to perform IndicesRecovery request: %w", err)
 	}
 
-	recovery := []IndexShardRecovery{}
+	recovery := []SnapshotIndexShardRecovery{}
 	for _, resp := range indicesRecovery {
 		for _, shard := range resp.Shards {
-			recovery = append(recovery, IndexShardRecovery{
+			// Add only the shards correspondign to the snapshot restoration
```
agnivade, line 345: typo at corresponding

agnivade (commented):
Unit tests need to be fixed.

agnivade:
Can we try to get this merged by today? I want to start a new test with a 2 node cluster.

deployment/elasticsearch/elasticsearch.go
agarciamontoro: Thanks! Fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/739/commits/f8275844cc4dfaf306389e6943c7214463545f6c

agnivade (approved):

isacikgoz (approved):

agarciamontoro:
Thank you, Ibrahim!

PR #740, opened 2024-05-09
MM-58178: Apply server side TCP buffer settings to client

agnivade:
In our investigation, we could see more packet drops on the client
side than server side. This would indicate that the TCP buffer
sizes on the client side need some bumping as well.

https://mattermost.atlassian.net/browse/MM-58178

### DIFF
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -247,6 +247,16 @@ net.ipv4.tcp_fin_timeout = 30
 
 # Reuse TIME-WAIT sockets for new outgoing connections.
 net.ipv4.tcp_tw_reuse = 1
+
+# TCP buffer sizes are tuned for 10Gbit/s bandwidth and 0.5ms RTT (as measured intra EC2 cluster).
+# This gives a BDP (bandwidth-delay-product) of 625000 bytes.
+net.ipv4.tcp_rmem = 4096 156250 625000
+net.ipv4.tcp_wmem = 4096 156250 625000
+net.core.rmem_max = 312500
+net.core.wmem_max = 312500
+net.core.rmem_default = 312500
+net.core.wmem_default = 312500
+net.ipv4.tcp_mem = 1638400 1638400 1638400
 `
 
 const serverSysctlConfig = `


agnivade:
@agarciamontoro - This test failed:

```
=== RUN   TestRoundTrip
    roundtripper_test.go:40: 
        	Error Trace:	/home/runner/work/mattermost-load-test-ng/mattermost-load-test-ng/deployment/elasticsearch/roundtripper_test.go:101
        	            				/home/runner/work/mattermost-load-test-ng/mattermost-load-test-ng/deployment/elasticsearch/roundtripper_test.go:40
        	Error:      	Received unexpected error:
        	            	dial tcp 127.0.0.1:2222: connect: connection refused
        	Test:       	TestRoundTrip
--- FAIL: TestRoundTrip (0.00s)
```

Perhaps our first flaky test of the repo?

agarciamontoro:
Oh! That's bad but also kind of a milestone, huh :joy: Will take a look, I don't get why that would fail :thinking:

deployment/terraform/strings.go
```diff
@@ -247,6 +247,16 @@ net.ipv4.tcp_fin_timeout = 30
 
 # Reuse TIME-WAIT sockets for new outgoing connections.
 net.ipv4.tcp_tw_reuse = 1
+
+# TCP buffer sizes are tuned for 10Gbit/s bandwidth and 0.5ms RTT (as measured intra EC2 cluster).
+# This gives a BDP (bandwidth-delay-product) of 625000 bytes.
```
agarciamontoro, line 252: I know this is copied from the app nodes, but it's a good opportunity for me to understand it better. The BDP computation I do understand it:
BDP = bandwidth * delay  = 10Gbit/s * 0.5ms = (10 * 10^9) * (0.5 * 10^-3) = 5000000 bits = 625000 bytes

But how are the other numbers computed? Are they all defined in terms of the BDP?

agarciamontoro (approved):
Looks good, let's check if things get better! I added a comment just for me to understand things better :)

deployment/terraform/strings.go
agarciamontoro: Also, is the RTT the same intra-cluster than between agents and proxy? Should we double-check that?

agnivade: @streamer45 wrote this. I'll defer to him.

streamer45 (approved):
Thanks, it would be nice to run a comparison at some point but not opposed to using the next monthly load-test. Just want to make sure we check for the results of this.

PR #741, opened 2024-05-10
feat: Add TOML support for config files

fmartingr:
#### Summary
This pull request adds TOML support for config files.

The `ReadFromJSON` function has been refactored into a unified `ReadFrom` with the same signature that determines the filetype by the extension the file has, if the extension is `.toml` a TOML decoder is used, following with JSON by default.

#### Code chagues

- Created a `Decoder` interface that we can use to simplify the unmarshalling logic in the code (try to decode, search for fallback, etc). Now we just pass the `Decoder` (or `DecoderFactory`) around and the rest of the logic is the same for all decoders.
- Implemented `JSONDecoder` and `TOMLDecoder` (from `json.Decoder` and `toml.Decoder`) since their signatures are different and also to allow TOML to show up better errors on decoding failures. 
- Main point is the "new" `ReadFrom` method that should take care of everything.
- It defaults to JSON so all regular usage should keep working as usual

#### Potential improvements

I focused on the Decoding part only, but if we plan to use it also as a encoder (to generate config files) we need to improve the data structure, interfaces and names around. It shouldn't be much but something to take into consideration.

I like to have "`Codec`" interfaces that take care of encoding/decoding around and create implementations for whatever thing I want to support, to I would go in that direction. Maybe in another PR if this one gets in.

#### Why

TOML files are way more readable for the end user, and we can put comments in the file for explanation of fields, better sort keys, etc.

Even better could be using the TOML's library capabilites to [add comments on TOML files](https://pkg.go.dev/github.com/pelletier/go-toml/v2#example-Marshal-Commented). I potentially see this in a way that:
- The data structure is defined with all necessary fields and comments
- Sample files are rendered directly from the structs
  - Creating the file automatically means we would lose manual organization, like in the deployer.sample.toml file I commited to this branch where I grouped settings. But that can be solved by nesting structs by topic.
- A CI step (lint) can create sample files from the configuration structs and check if the ones match with the code, ensuring that they are always up to date. This would lead to the data structures being the source of truth for the configuration while providing documentation right in the TOML files. 

#### Test run

Performed a trial deployment with `config` and `deployer` TOML files. Didn't test all commands manually.

#### Ticket Link
--


defaults/json.go
```diff
@@ -2,51 +2,15 @@ package defaults
...
-
+type JSONDecoder struct {
+	*json.Decoder
 }
 
-func readJSON(path string, value interface{}) error {
-	file, err := os.Open(path)
-	if err != nil {
-		return fmt.Errorf("could not open file: %w", err)
+func NewJSONDecoder(r io.Reader) Decoder {
```
agnivade, line 12: It is an anti-pattern to return an interface. Instead you can return the struct, but just keep accepting interfaces like you are doing now. https://developers.mattermost.com/contribute/more-info/server/style-guide/#hahahugoshortcode-s4-hbhb

coordinator/config.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/logger"
 )
 
+const DefaultConfigLocation = "./config/coordinator.json"
```
agnivade, line 13: Can be unexported.

defaults/main.go
```diff
@@ -0,0 +1,70 @@
...
+type DecoderFactory func(r io.Reader) Decoder
+
+// ReadFrom reads a file from the given path and path and decodes it into the given value,
+// based on the file extension.
+func ReadFrom(path, fallbackPath string, value any) error {
+	if strings.HasSuffix(path, ".toml") || strings.HasSuffix(fallbackPath, ".toml") {
+		return readFromFile(NewTOMLDecoder, path, fallbackPath, value)
+	}
+
+	return readFromFile(NewJSONDecoder, path, fallbackPath, value)
```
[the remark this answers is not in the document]
fmartingr: I've gone further and removed the factory altogether, putting that same logic directly in the `read` method to:
1. Make it more readable
2. Focusing the decoder logic in the place it's actually used
3. Allows for different file types in `path` and `fallbackPack`.

https://github.com/mattermost/mattermost-load-test-ng/pull/741/commits/b3233353cee7e77513a74e3a5ffeafeb8e838fb5

Let me know what you think.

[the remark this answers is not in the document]
agnivade: Looks good!

agnivade (approved):

config/deployer.sample.toml
```diff
@@ -0,0 +1,138 @@
+# AWS Configuration
```
agarciamontoro, line 1: TOML having comments is already a big advantage.

defaults/main.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type Decoder interface {
+	Decode(value any) error
+	DisallowUnknownFields()
+}
+
+// readFromFile reads a configuration file to the value.
+// This function will try to read from given path, if it is empty will try
+// fallback path. If it fails on fallback, it will set value to it's defaults
```
agarciamontoro, line 16: A couple of typos
```suggestion
// ReadFrom reads a configuration file to the value.
// This function will try to read from given path, if it is empty will try
// fallback path. If it fails on fallback, it will set value to its defaults
```

defaults/main.go
```diff
@@ -0,0 +1,63 @@
...
+// read reads a file to the value using the provided decoder
+func read(path string, value any) error {
+	file, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("could not open file: %w", err)
+	}
+	defer file.Close()
+
+	var dec Decoder
+	if strings.Contains(path, ".toml") {
```
agarciamontoro, line 50: Should this be `HasSuffix`?
```suggestion
	if strings.HasSuffix(path, ".toml") {
```

agarciamontoro (changes requested):
This looks great! Thank you so much for putting the effort into implementing this <3 Just a  couple of nits before approving.

PR #742, opened 2024-05-10
refactor: keycloak PR remaining TODOs

fmartingr:
#### Summary

- Moved the parsing of the authentication type from an user from an users file to a separate function and added tests
- Use a template file instead of code to populate the `keycloak.env` file
- Set the user in the store directly in the `Login`/`SignUp` methods instead of `authOpenID` method.


api/utils.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"strings"
+
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/user/userentity"
+)
+
+// user represents a user in the users.txt file
+// The format of each line in the file is:
+// (authentication_type:)?email password
+func parseUserFromLine(u user) (username, email, password, authenticationType string) {
```
streamer45, line 12: The more I look at this the more I am wondering whether this logic should be part of `getUserCredentials`. Would that make sense or are there further complexities?

loadtest/user/userentity/actions.go
```diff
@@ -134,23 +123,29 @@ func (ue *UserEntity) Login() error {
...
 	}
+	var loggedUser *model.User
 
 	switch ue.config.AuthenticationType {
 	case AuthenticationTypeOpenID:
 		if err := ue.authOpenID(authOpenIDLogin); err != nil {
 			return fmt.Errorf("error while logging in using OpenID: %w", err)
 		}
+
+		loggedUser, _, err = ue.client.GetUserByUsername(context.Background(), user.Username, "")
```
agnivade, line 134: Ah okay, so `Signup` already had the right code. We just hadn't applied it to `Login`? That feels a bit weird.

agnivade (approved):
Looks good, but could you please give this a test with a fresh deployment?

loadtest/user/userentity/actions.go
fmartingr: Yeah, in my changes (thinking it wasn't working at that particular time because of a Keycloak issue) I forgot that `Login` should store the user the same way as `SignUp` does.

api/utils.go
fmartingr: Yeah that's probably a better place for it: 
- Logic refactor: https://github.com/mattermost/mattermost-load-test-ng/pull/742/commits/a6798195fc51d26b9dc11b024c2b25ee5ffd9064
- Tests: https://github.com/mattermost/mattermost-load-test-ng/pull/742/commits/0fc2bc12c5b7253372965237b46c1f0c2446b217

streamer45 (approved):
Thanks, looks great!

fmartingr:
> but could you please give this a test with a fresh deployment?

I tested this locally and can confirm that users can login correctly using openID

PR #743, opened 2024-05-10
[WIP] Add "wizard createconfig" command to easily populate configuration files

fmartingr:
#### Summary
In order to make iteration, first usage and potentially development faster I see myself editing, copying and pasting files manually around so it occurred to me that having a command to easily populate the basics for configuration files would be particularily useful.

I envisioned this as a question-answer prompt, but having a command with flags seems more appropriate here and we can always evolve into something like that.

This Pull Request contains the first steps for that, allowing to run a `ltctl wizard createconfig` command with several flags to populate the `config.json` and `deployment.json` files with predefined values for an environment with an specific amount of active concurrent users (determines by the `--active-users` flag).

I added some draft structures around to define this architecture list borrowing the information from the references architectures work, but we would also need to define number of agents and all other load-test specific configuration.

This is just a first step/draft idea, let me know what you think @agnivade @agarciamontoro

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -420,6 +420,31 @@ func main() {
 	comparisonCmd.AddCommand(runComparisonCmd, destroyComparisonCmd, collectComparisonCmd)
 	rootCmd.AddCommand(comparisonCmd)
 
+	wizardCmd := &cobra.Command{
+		Use:               "wizard",
+		Short:             "Wizard utilities to setup the",
+		PersistentPreRun:  func(cmd *cobra.Command, _ []string) { setServiceEnv(cmd) },
+		PersistentPostRun: func(_ *cobra.Command, _ []string) { os.Unsetenv("MM_SERVICEENVIRONMENT") },
+	}
+
+	wizardCreateConfigCmd := &cobra.Command{
+		Use:   "createconfig",
+		Short: "Create configuration files",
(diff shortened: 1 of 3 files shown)


agarciamontoro (commented):
I like this idea! In particular, I like the `deployment/infrastructure.go` file: I feel there's great potential there to manage templates like you propose, and we can follow the results of the ceiling tests as a first step. Thank you!

I also like the idea of a question-answer prompt, and I think we can directly work on that. Do you want to keep working on this or is there a lot on your plate right now?

Btw, feel free to ask for a review in these PRs even if they're in a WIP state. Otherwise I'll probably forget about them, as it happened with this :nerd_face: Sorry!

agnivade:
I personally feel more comfortable directly editing config files. But I don't mind this. Feel free to ask a review once it's ready.

fmartingr:
> I personally feel more comfortable directly editing config files. But I don't mind this. Feel free to ask a review once it's ready.

My objective with this is having a initial setup to create the config files, and then those should be edited manually. But for example if you want to have a load-test for a 50k envionrment, instead of modifying the files manually you could just execute the CLI command to setup those values in the files with one command, and then proceed to tune other things that you may want to change.

agnivade:
I usually copy the template from *.sample.json and modify them manually. But it's just my personal preference.

PR #746, opened 2024-05-22
Check DB status before destroy

agnivade:
We check if the DB is available or not before destroying.
Because going ahead with the destroy if the DB is stopped
messes up with the Terraform state.

https://mattermost.atlassian.net/browse/MM-58236

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -48,14 +48,18 @@ func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
 
-	status, err := t.DBStatus()
-	if err != nil {
-		return fmt.Errorf("failed to get DB status: %w", err)
+	// If we are using a Terraform managed DB.
+	if config.ExternalDBSettings.DataSource == "" {
+		status, err := t.DBStatus()
+		if err != nil {
+			return fmt.Errorf("failed to get DB status: %w", err)
+		}
+
+		if status != dbAvailable {
+			return fmt.Errorf("The database isn't available at the moment. Its status is %q. Please check this, and then try again", status)
+		}
 	}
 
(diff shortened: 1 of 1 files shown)


streamer45 (approved):

cmd/ltctl/main.go
```diff
@@ -47,6 +47,16 @@ func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
...
+
+	status, err := t.DBStatus()
+	if err != nil {
+		return fmt.Errorf("failed to get DB status: %w", err)
+	}
+
+	if status != dbAvailable {
+		return fmt.Errorf("The database isn't available at the moment. Its status is %q. Please check this, and then try again", status)
+	}
+
```
agarciamontoro, line 59: We should probably run this only if there is a deployed DB. If not, we're using an external one and the destroy will not affect it in any case.

agarciamontoro (changes requested):

cmd/ltctl/main.go
```diff
@@ -47,6 +47,19 @@ func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
+
+	// If we are using a Terraform managed DB.
+	if config.ExternalDBSettings.DataSource == "" {
```
agarciamontoro, line 52: We could also use `t.output.HasDB()`, which seems more idiomatic and we'd avoid the comment, but then you'd need to run `t.loadOutput()` before that (or do it inside `t.Destroy()`, which already loads the output)

agarciamontoro (approved):
Looks good to me! I've left a comment with a potential minor improvement, but up to you whether to follow it.

cmd/ltctl/main.go
agnivade: Yeah, I think it's okay.

PR #749, opened 2024-05-27
MM-58347: Ensure all Elasticsearch data nodes have shard replicas

agarciamontoro:
#### Summary
We [recently found out](https://mattermost.atlassian.net/wiki/spaces/XYZ/pages/2827059205/ElasticSearch+load-balancing+investigation+report) that the cause of the unbalanced Elasticsearch queries was the lack of shard replicas in most of the nodes, since we were using the default value of 1 replica per shard, which was allocated to a random data node. The solution to get all data nodes to handle queries at the same rate, then, is to have one replica of the indices' shards in each of them. We only have one shard per index, so this means that every node will contain exactly one copy of each index.

That's what this PR does in [the first commit](https://github.com/mattermost/mattermost-load-test-ng/commit/98b4194c7af8165fa209d5b5a12f0507343c700f): fortunately, the Elasticsearch API allows overwriting index settings when restoring a snapshot, so I just had to modify that request to also specify the number of replicas we wanted per shard.

Then I noticed another problem: even once the snapshot is fully restored, the replicas of shards may not be assigned or initialized, making the [cluster status](https://www.elastic.co/guide/en/elasticsearch/reference/7.17/cluster-health.html) yellow instead of green. We do need to wait for the cluster status to get green again to get the load-balancing of queries perfectly right, so [the second commit](https://github.com/mattermost/mattermost-load-test-ng/commit/a4c11a1b6439a7f9582973b9007003339ee3442b) adds a new `ClusterHealth` method to the Elasticsearch client to check the cluster status. We use it to wait for the green status, while informing the user of the remaining unassigned or initializing shards.

Even after all this, with all shards assigned and initialized, and the cluster green, I was getting an index with only two replicas, regardless of the number of data nodes I deployed :angry: I later realized that it happened only with the new indices, not the ones we restored from the snapshot (the server creates a new posts index as soon as it starts to track the new posts), so I had to configure the number of replicas in the server config as well, something we should have done before, honestly. That's what [the third commit](https://github.com/mattermost/mattermost-load-test-ng/commit/1205f7a220392875b574d9f972362978a7538b67) does. (I [first thought](https://community.mattermost.com/core/pl/en4xcbawjb8tbptpps6rpj9z7y) I needed to create a new index template, but that's not actually needed, since the server already takes care of that configuration for us).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-58347

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -454,6 +454,8 @@ func (t *Terraform) setupLoadtestAgents(extAgent *ssh.ExtAgent, initData bool) e
 }
 
 func (t *Terraform) setupElasticSearchServer(extAgent *ssh.ExtAgent, ip string) error {
+	esSettings := t.config.ElasticSearchSettings
+
 	output, err := t.Output()
 	if err != nil {
 		return fmt.Errorf("unable to get Terraform output to setup ElasticSearch: %w", err)
@@ -482,7 +484,7 @@ func (t *Terraform) setupElasticSearchServer(extAgent *ssh.ExtAgent, ip string) 
 	}
 	mlog.Debug("Repositories registered", mlog.Array("repositories", repositories))
 
-	repo := t.config.ElasticSearchSettings.SnapshotRepository
+	repo := esSettings.SnapshotRepository
 
 	// Check if the registered repositories already include the one configured
 	repoFound := false
@@ -511,7 +513,7 @@ func (t *Terraform) setupElasticSearchServer(extAgent *ssh.ExtAgent, ip string) 
 
 	// Look for the configured snapshot
 	var snapshot elasticsearch.Snapshot
-	snapshotName := t.config.ElasticSearchSettings.SnapshotName
+	snapshotName := esSettings.SnapshotName
 	for _, s := range snapshots {
 		if s.Name == snapshotName {
 			snapshot = s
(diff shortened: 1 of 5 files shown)


agnivade (approved):
Really high quality commit message!

deployment/terraform/create.go
```diff
@@ -557,15 +557,40 @@ func (t *Terraform) setupElasticSearchServer(extAgent *ssh.ExtAgent, ip string)
...
-		WithIndices: snapshotIndices,
+		WithIndices:      snapshotIndices,
+		NumberOfReplicas: t.config.ElasticSearchSettings.InstanceCount - 1,
 	}
 	if err := es.RestoreSnapshot(repo, snapshotName, opts); err != nil {
 		return fmt.Errorf("unable to restore snapshot: %w", err)
 	}
 
-	// Wait until the snapshot is completely restored, or 30 minutes have
+	// Wait until the snapshot is completely restored, or 45 minutes have
```
streamer45, line 567: So, this was 30, now it's 45. Should we make it configurable or at least overridible through an environment variable?

deployment/terraform/create.go
```diff
@@ -613,6 +638,34 @@ func (t *Terraform) setupElasticSearchServer(extAgent *ssh.ExtAgent, ip string)
...
+// waitForGreenCluster blocks until the cluster's status is green or the
+// provided timeout is reached
+func waitForGreenCluster(dur time.Duration, es *elasticsearch.Client) error {
+	timeout := time.After(dur)
+	for {
+		select {
+		case <-timeout:
+			return fmt.Errorf("timeout after %s, cluster's status did not become green", dur.String())
+		case <-time.After(30 * time.Second):
+			clusterHealth, err := es.ClusterHealth()
```
streamer45, line 650: Is there a timeout set for this request? Just want to make sure this couldn't block indefinitely or something.

streamer45 (approved):
Great work, nothing blocking.

deployment/terraform/create.go
agarciamontoro: It defaults to 30 seconds, yes: https://www.elastic.co/guide/en/elasticsearch/reference/7.10/cluster-health.html

deployment/terraform/create.go
agarciamontoro: Good point. Fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/749/commits/8060c61abacc2aebe30a49740728aa0db1df524b. Testing it now.

PR #755, opened 2024-06-12
Update postgres client

streamer45:
#### Summary

I got a weird error describing a table on a deployment I am testing on and realized our server is v14, but our client is still v11, which is causing some incompatibilities. Updating the client to match the server. Let me know if there are concerns.


agarciamontoro (approved):
Oh, yes, I've seen this before, but I had some workaround. Thanks for changing it!

agnivade (approved):

PR #758, opened 2024-06-27
ltkeycloak and SAML improvements

fmartingr:
#### Summary
- Keycloak helper command to ease out the load-test environment configuration by syncing the mattermost users into the keycloak server.
- Added defaults for automatic SAML configuration in the load-test environment
- Fixed some issues


agarciamontoro:
Just curious: how's this going?

fmartingr:
Stalled at the moment.

agarciamontoro:
Got it, thanks!

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+			_, err := usersTxtFile.Write([]byte(model.UserAuthServiceSaml + ":" + email + " " + userPassword + "\n"))
+			if err != nil {
+				mlog.Error("failed to write to users.txt", mlog.String("err", err.Error()))
+			}
+
+		case <-doneChan:
+			return
+		}
+	}
+}
```
agarciamontoro, line 39: If opening the file or writing to it fails, we should return.

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+	keycloakClient   *gocloak.GoCloak
+	keycloakToken    *gocloak.JWT
+	keycloakRealm    string
+	userPassword     string
+	deploymentConfig *deployment.Config
+	doneChan         chan struct{}
+}
+
+func migrateMattermostUsersToKeycloak(worker *workerConfig) {
+	refreshTokenTicker := time.NewTicker(30 * time.Second)
```
agarciamontoro, line 55: Wow, do we need to refresh the token every 30 seconds?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+				wg.Add(1)
+				usersChan <- user
+			} else {
+				mlog.Info("dry-run: would migrate user", mlog.String("username", user.Username))
+			}
+		}
+
+		if len(users) == 0 {
+			break
+		}
```
agnivade, line 260: I think this bug is still there right? It should be `len(users) < perPage`.

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+			if err != nil {
+				cancel()
+				// Ignore already existing users
+				if apiErr, ok := err.(*gocloak.APIError); ok && apiErr.Code == 409 {
+					mlog.Debug("user already exists in keycloak", mlog.String("username", user.Username))
+					continue
+				}
+
+				mlog.Error("failed to create user in keycloak", mlog.String("err", err.Error()))
+				continue
```
agarciamontoro, line 85: Should we fail early here if a user couldn't be created?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+			}
+
+			user.AuthData = model.NewString(kcUserID)
```
agarciamontoro, line 88: We can use `NewPointer` now!
```suggestion
			user.AuthData = model.NewPointer(kcUserID)
```

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+			user.AuthService = model.UserAuthServiceSaml
+			user.Password = ""
+
+			_, _, err = worker.mmClient.UpdateUser(ctx, user)
+			if err != nil {
+				cancel()
+				mlog.Error("failed to update user in mattermost", mlog.String("err", err.Error()))
+				continue
+			}
```
agarciamontoro, line 97: Again: should we fail early in this case? Or is it safe to continue?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+				Credentials: &[]gocloak.CredentialRepresentation{
+					{
+						Temporary: gocloak.BoolP(false),
+						Type:      gocloak.StringP("password"),
+						Value:     gocloak.StringP(worker.userPassword),
+					},
+				},
+			})
+			if err != nil {
+				cancel()
```
agarciamontoro, line 77: Do we need to call cancel here if `worker.keycloakClient.CreateUser` has already returned?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+				continue
+			}
+
+			user.AuthData = model.NewString(kcUserID)
+			user.AuthService = model.UserAuthServiceSaml
+			user.Password = ""
+
+			_, _, err = worker.mmClient.UpdateUser(ctx, user)
+			if err != nil {
+				cancel()
```
agarciamontoro, line 94: Same here: do we need to call cancel after `worker`.mmClient.UpdateUser` has returned?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+		case <-refreshTokenTicker.C:
+			ctx, cancel := context.WithTimeout(context.Background(), operationTimeout)
+
+			mlog.Info("refreshing keycloak token", mlog.Int("worker", worker.workerNumber))
+			var err error
+			worker.keycloakToken, err = worker.keycloakClient.LoginAdmin(
+				ctx,
+				worker.deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminUser,
+				worker.deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminPassword,
+				"master", // TODO: Allow specifying the master realm
```
agarciamontoro, line 110: Should this TODO be filed as a ticket?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+			)
+			if err != nil {
+				cancel()
+				mlog.Error("failed to refresh keycloak token", mlog.String("err", err.Error()))
+				close(worker.doneChan)
+			}
+
+			cancel()
```
agarciamontoro, line 118: Same comments about `cancel` here.

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+
+		if len(users) == 0 {
+			break
+		}
+
+		page++
+	}
+
+	wg.Wait()
+	close(doneChan)
```
agnivade, line 266: One last bug is to wait until the workers and `userTxtWriter` has actually returned. Refer to the web_conn.go link that I mentioned earlier. Basically it's ensuring that the goroutines are actually exiting before you move on with the code. I will leave it to you if you want to ensure that.

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -0,0 +1,286 @@
...
+		return fmt.Errorf("failed to login to mattermost: %w", err)
+	}
+
+	var keycloakHost string
+	keycloakHost, err = cmd.Flags().GetString("keycloak-host")
+	if err != nil {
+		return fmt.Errorf("failed to read flag: %w", err)
+	}
+
+	// Use the terraform output terraform host if a manual one is not provided. Useful for development.
```
agarciamontoro, line 157: Typo? Should the second `terraform` be `keycloak`? Not sure

cmd/ltkeycloak/from_mattermost.go
agarciamontoro, line 260: Nit because the semantics don't change, but: can we move this right after we get `users`, before the loop?

deployment/terraform/keycloak.go
```diff
@@ -331,3 +332,67 @@ func (t *Terraform) IngestKeycloakDump() error {
...
+
+	keycloakUrl := keycloakScheme + "://" + t.output.KeycloakServer.PublicDNS + ":8080"
+
+	cfg.OpenIdSettings.Enable = model.NewPointer(true)
+	cfg.OpenIdSettings.ButtonText = model.NewPointer("Keycloak Login")
+	cfg.OpenIdSettings.DiscoveryEndpoint = model.NewPointer(keycloakUrl + "/realms/" + t.config.ExternalAuthProviderSettings.KeycloakRealmName + "/.well-known/openid-configuration")
+	cfg.OpenIdSettings.Id = model.NewPointer(t.config.ExternalAuthProviderSettings.KeycloakClientID)
+	cfg.OpenIdSettings.Secret = model.NewPointer(t.config.ExternalAuthProviderSettings.KeycloakClientSecret)
+	cfg.SamlSettings.Enable = model.NewPointer(false)
+	cfg.SamlSettings.Enable = model.NewPointer(true)
```
agarciamontoro, line 362: Typo
```suggestion
	cfg.SamlSettings.Enable = model.NewPointer(true)
```

agarciamontoro (changes requested):
Great work! I found the sync code a bit harder to read, but it's an inherently complex operation, so I don't have lots of suggestions to offer on that. I have a few other comments below, thanks!

cmd/ltkeycloak/from_mattermost.go
agarciamontoro: Oh, true!

cmd/ltkeycloak/from_mattermost.go
fmartingr: Yeah, it has very limited expiration time.

cmd/ltkeycloak/from_mattermost.go
fmartingr: I left it that way in case something unexpected hapenned with one user and it prevented completing the rest of the users in other runs, stopping here.

cmd/ltkeycloak/from_mattermost.go
fmartingr: We need to call `cancel` when the operation finishes (same as `defer`), that means correctly or incorrectly, in order to free resources. If we don't mind that I can just keep using `defer`.

agarciamontoro: Ah, sorry, you're right, a function could return while its goroutines may still be around listening for the ctxt :facepalm: 
I was gonna suggest to call `defer cancel()` right after its definition so that we can skip all the individual calls to `cancel`, but we need to `continue` in some cases. Would it be worth it to wrap all this logic in a function? I'll leave it to you.

agarciamontoro (approved):
Looks good! I just left [a comment](https://github.com/mattermost/mattermost-load-test-ng/pull/758#discussion_r1727075203) in a previous discussion.

agnivade (approved):
Looking great!

cmd/ltkeycloak/from_mattermost.go
fmartingr: Something like https://github.com/mattermost/mattermost-load-test-ng/pull/758/commits/19222db309f5ff41ce6eac099204005d95b8fa74 ?

agnivade:
@agarciamontoro - can we merge this? Felipe is on PTO.

agarciamontoro:
Indeed we can!

PR #759, opened 2024-07-01
feat: allow saml with the external auth provider

fmartingr:
#### Summary
- Allow using `saml` with the external auth provider
- Logic is mostly the same OpenID uses, just added logic to control the redirect from the IDP (keycloak) back to Mattermost to retrieve the cookie, since that uses a form and javascript on the keycloak side.

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -508,7 +508,7 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
 				return nil, fmt.Errorf("invalid custom authentication found in %q", email)
 			}
 			authService = split[0]
-			if authService != userentity.AuthenticationTypeOpenID && authService != userentity.AuthenticationTypeMattermost {
+			if authService != userentity.AuthenticationTypeOpenID && authService != userentity.AuthenticationTypeSAML && authService != userentity.AuthenticationTypeMattermost {
 				return nil, fmt.Errorf("invalid custom authentication type %q", authService)
 			}
 
--- a/api/agent.go
+++ b/api/agent.go
@@ -508,7 +508,11 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
 				return nil, fmt.Errorf("invalid custom authentication found in %q", email)
 			}
 			authService = split[0]
-			if authService != userentity.AuthenticationTypeOpenID && authService != userentity.AuthenticationTypeSAML && authService != userentity.AuthenticationTypeMattermost {
+			switch(authService) {
+			case userentity.AuthenticationTypeOpenID:
+			case userentity.AuthenticationTypeSAML:
+			case userentity.AuthenticationTypeMattermost:
+			default :
 				return nil, fmt.Errorf("invalid custom authentication type %q", authService)
 			}
 
--- a/api/agent.go
+++ b/api/agent.go
@@ -508,11 +508,11 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
 				return nil, fmt.Errorf("invalid custom authentication found in %q", email)
 			}
 			authService = split[0]
-			switch(authService) {
+			switch authService {
 			case userentity.AuthenticationTypeOpenID:
 			case userentity.AuthenticationTypeSAML:
 			case userentity.AuthenticationTypeMattermost:
-			default :
+			default:
 				return nil, fmt.Errorf("invalid custom authentication type %q", authService)
 			}
 
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -31,6 +31,7 @@ const (
 
 var (
 	keycloakIDPLoginFormActionRegex = regexp.MustCompile(`action=["'](.*?)["']`)
+	keycloakSAMLInputValueRegex     = regexp.MustCompile(`input type=["']hidden["'] name=["'](\w+)["'] value=["'](.*?)["']`)
 )
 
 // SignUp signs up the user with the given credentials.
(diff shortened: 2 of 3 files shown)


api/agent.go
```diff
@@ -508,7 +508,7 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
 				return nil, fmt.Errorf("invalid custom authentication found in %q", email)
 			}
 			authService = split[0]
-			if authService != userentity.AuthenticationTypeOpenID && authService != userentity.AuthenticationTypeMattermost {
+			if authService != userentity.AuthenticationTypeOpenID && authService != userentity.AuthenticationTypeSAML && authService != userentity.AuthenticationTypeMattermost {
 				return nil, fmt.Errorf("invalid custom authentication type %q", authService)
 			}
```
agarciamontoro, line 513: Nit: Should this be a switch? It's getting a bit harder to read. Something like 
```suggestion
			switch(authService) {
			case userentity.AuthenticationTypeOpenID:
			case userentity.AuthenticationTypeSAML:
			case userentity.AuthenticationTypeMattermost:
			default :
				return nil, fmt.Errorf("invalid custom authentication type %q", authService)
			}
```

loadtest/user/userentity/actions.go
```diff
@@ -110,7 +124,48 @@ func (ue *UserEntity) authOpenID(action authOpenIDAction) error {
 		return fmt.Errorf("%s failed with status code %d", action, loginResponse.StatusCode)
 	}
 
-	cookies := jar.Cookies(loginResponse.Request.URL)
+	// Perform an extra step when doing SAML authentication, since the Keycloak server wont redirect
```
agarciamontoro, line 127: Typo
```suggestion
	// Perform an extra step when doing SAML authentication, since the Keycloak server won't redirect
```

loadtest/user/userentity/actions.go
```diff
@@ -110,7 +124,48 @@ func (ue *UserEntity) authOpenID(action authOpenIDAction) error {
...
+	// to the Mattermost server automatically, instead it will display a form with the SAML response
+	// that needs to be submitted to the Mattermost server. This is done via Javascript in the browser,
+	// but we can simulate it by parsing the form and doing the same request manually.
+	if provider == AuthenticationTypeSAML {
+		samlResponseBody, err := io.ReadAll(loginResponse.Body)
+		if err != nil {
+			return fmt.Errorf("error while reading saml response body: %w", err)
+		}
```
agarciamontoro, line 135: We should close the body after this.

loadtest/user/userentity/actions.go
```diff
@@ -21,16 +21,17 @@ import (
...
-	authOpenIDLogin  authOpenIDAction = "login"
-	authOpenIDSignup authOpenIDAction = "signup"
+	authIDPLogin  authIDPAction = "login"
+	authIDPSignup authIDPAction = "signup"
 )
 
 var (
-	openIDLoginFormActionRegex = regexp.MustCompile(`action=["'](.*?)["']`)
+	keycloakIDPLoginFormActionRegex = regexp.MustCompile(`action=["'](.*?)["']`)
+	keycloakSAMLInputValueRegex     = regexp.MustCompile(`input type=["']hidden["'] name=["'](\w+)["'] value=["'](.*?)["']`)
```
agarciamontoro, line 34: "a mere glimpse of the world of reg​ex parsers for HTML will ins​tantly transport a programmer's consciousness into a world of ceaseless screaming" https://stackoverflow.com/a/1732454/3248221

(It's good for this specific case, just kidding)

loadtest/user/userentity/actions.go
```diff
@@ -110,7 +124,48 @@ func (ue *UserEntity) authOpenID(action authOpenIDAction) error {
...
+			queryParams.Add(string(matches[1]), string(matches[2]))
+		}
+
+		samlForm, err := client.PostForm(formURL, queryParams)
+		if err != nil {
+			return fmt.Errorf("error while posting SAML form: %w", err)
+		}
+		if samlForm.StatusCode != http.StatusOK {
+			return fmt.Errorf("SAML form failed with status code %d", samlForm.StatusCode)
+		}
```
agarciamontoro, line 161: We should close the body here as well even if we don't use it

agarciamontoro (changes requested):
Looks great! A couple of nits and minor comments, nothing major.

agarciamontoro (approved):
Looks great, thank you!

loadtest/user/userentity/actions.go
```diff
@@ -110,7 +124,50 @@ func (ue *UserEntity) authOpenID(action authOpenIDAction) error {
...
+	if provider == AuthenticationTypeSAML {
+		samlResponseBody, err := io.ReadAll(loginResponse.Body)
+		if err != nil {
+			return fmt.Errorf("error while reading saml response body: %w", err)
+		}
+		loginResponse.Body.Close()
+
+		redirectURLMatcher := keycloakIDPLoginFormActionRegex.FindSubmatch(samlResponseBody)
+		if len(redirectURLMatcher) == 0 {
+			return errors.New("redirect URL not found in SAML login page, there was probably an error or the configuration is wrong")
```
agnivade, line 140: Maybe we can be a bit more clear in the error mentioning that it's also possible that the UI has changed which has broken the regex?

agnivade (commented):
One small comment. Not sure how this got delayed for so long.

loadtest/user/userentity/actions.go
fmartingr: 👍

agnivade (approved):

PR #762, opened 2024-07-03
MM-59316: Remove cloudwatch log policy

agarciamontoro:
#### Summary
Instead of creating a log policy per cluster, given that we can only have 10 policies per region per account, I've manually created the one we need and removed the automatically deployed one.

~~If other people use this, they'll have to create such policy. If not, they won't have access to Opensearch logs in Cloudwatch, but that's the only drawback here. Everything else should still work (testing it now).~~ EDIT: This is now automatically done by the deployer. See [comment below](https://github.com/mattermost/mattermost-load-test-ng/pull/762#issuecomment-2209458497).

The manual policy was created with:
```
aws --profile mm-loadtest logs put-resource-policy --policy-name lt-cloudwatch-log-policy --policy-document "{\"Version\":\"2012-10-17\",\"Statement\":[{\"Effect\":\"Allow\",\"Principal\":{\"Service\":\"es.amazonaws.com\"},\"Action\":[\"logs:PutLogEventsBatch\",\"logs:PutLogEvents\",\"logs:CreateLogStream\"],\"Resource\":\"arn:aws:logs:*\"}]}"
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59316


agnivade (commented):
How are other customers expected to know this? Should the tool throw some warnings about it?

agarciamontoro:
That's a good question. There's two options I see now:
1. Document it as a pre-requisite (and expect customers to read the docs).
2. Check, before running `terraform apply`, whether such a policy exists, by inspecting their permissions. If not, throw a warning.

Let me see if 2. is easily doable and, if not, we can go with 1.

agarciamontoro:
Ok, so I went with solution 3: check in the code that such a policy exists, and create it if it doesn't. This keeps the deployment idempotent and expects no user action (unless something goes wrong, of course, in which case the error logged is self-explanatory).

Let me know what you think!

deployment/terraform/create.go
```diff
@@ -324,6 +332,175 @@ func (t *Terraform) Create(initData bool) error {
...
+		Statement: []polDocStmt{
+			{
+				Effect: "Allow",
+				Principal: polDocStmtPrincipal{
+					Service: "es.amazonaws.com",
+				},
+				Action: []string{
+					"logs:PutLogEventsBatch",
+					"logs:PutLogEvents",
+					"logs:CreateLogStream",
```
agnivade, line 480: Maybe we can store them as constants? Since they are also being used in the comparison function.

deployment/terraform/create.go
```diff
@@ -324,6 +332,175 @@ func (t *Terraform) Create(initData bool) error {
...
+//		},
+//	}
+//
+// If such policy does not exist, we create it. Note that we are doing this
+// outside of Terraform. This is because we only need one policy in the
+// account, not one per deployment. This behaviour is also enforced by the rate
+// limitation on this type of policies: there can only be 10 such policies per
+// region per account. Check the docs for more information:
+// https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/cloudwatch_limits_cwl.html
+func (t *Terraform) checkCloudWatchLogsPolicy() error {
```
agnivade, line 365: This does more than just "check", so can we rename it to "createCloudWatchLogsPolicyIfNotExists"?

agnivade (commented):
Lastly, the `terraform/create.go` file is becoming too big. Can we split `checkCloudWatchLogsPolicy` into its own separate file?

agarciamontoro:
@agnivade, I've refactored everything to a new file into two functions: one for checking, another for creating. I've adjusted the names and tested the verification step. Thanks for the review!

agnivade (approved):

isacikgoz (approved):

PR #764, opened 2024-07-04
Allow 0 shard replicas

agarciamontoro:
#### Summary
I knew I had tested 1-node ES clusters before! I must have tested them prior to the implementation of the MarshalJSON for the RestoreSnapshotOpts struct, though, where I mistakenly assumed the default number of replicas must be 1 instead of 0. This is what made the cluster status to be yellow, and, [contrary to the docs](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/handling-errors.html#handling-errors-yellow-cluster-status) (:shrug:), if we set the number of replicas to 0 when there are no replicas, the cluster status is green, as it should be.

Check an example of a 1-node cluster with a green status :tada: 
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-59312

### DIFF
--- a/deployment/elasticsearch/elasticsearch.go
+++ b/deployment/elasticsearch/elasticsearch.go
@@ -243,7 +243,7 @@ func (r RestoreSnapshotOpts) MarshalJSON() ([]byte, error) {
 	}
 
 	// Set the number of replicas of each shard
-	numReplicas := 1
+	numReplicas := 0
 	if r.NumberOfReplicas > 0 {
 		numReplicas = r.NumberOfReplicas
 	}
--- a/deployment/elasticsearch/elasticsearch.go
+++ b/deployment/elasticsearch/elasticsearch.go
@@ -242,23 +242,29 @@ func (r RestoreSnapshotOpts) MarshalJSON() ([]byte, error) {
 		indices = append(indices, "-"+i)
 	}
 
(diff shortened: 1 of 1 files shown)


agnivade (approved):
Weird coincidence for a bug like that to happen!

deployment/elasticsearch/elasticsearch.go
```diff
@@ -243,7 +243,7 @@ func (r RestoreSnapshotOpts) MarshalJSON() ([]byte, error) {
 	}
 
 	// Set the number of replicas of each shard
-	numReplicas := 1
+	numReplicas := 0
```
streamer45, line 246: Why not use `r.NumberOfReplicas` directly in the first place?

agarciamontoro: In case the caller set a negative number. But in this case we should throw an error, so I added a small validation step before. See https://github.com/mattermost/mattermost-load-test-ng/pull/764/commits/32788e6abf8395768afb2d189a89b58b1e9fd5b1

streamer45: Why not make it unsigned, then? Just kidding, it's the endless dilemma :stuck_out_tongue: 

I like the new approach, failing fast in case of a clear input error.

streamer45 (approved):

PR #765, opened 2024-07-11
Improvements from Redis investigation

agnivade:
- Turn off the cache_lock. This is unnecessary and
we shouldn't even have done this to start with.
- Set timeout for entire response and not just response header.
It is unclear why exactly this causes the problem. So open to
thoughts and suggestions.

### DIFF
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -194,7 +194,6 @@ proxy_cache mattermost_cache;
 proxy_cache_revalidate on;
 proxy_cache_min_uses 2;
 proxy_cache_use_stale timeout;
-proxy_cache_lock on;
 `
 
 const nginxSiteConfigTmpl = `
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -181,7 +181,7 @@ proxy_buffers 256 16k;
 proxy_buffer_size 16k;
 client_body_timeout 60s;
 send_timeout        300s;
-lingering_timeout   5s;
+lingering_timeout   30s;
 proxy_connect_timeout   30s;
 proxy_send_timeout      90s;
 proxy_read_timeout      90s;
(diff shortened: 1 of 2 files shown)


deployment/terraform/strings.go
```diff
@@ -194,7 +194,6 @@ proxy_cache mattermost_cache;
 proxy_cache_revalidate on;
 proxy_cache_min_uses 2;
 proxy_cache_use_stale timeout;
-proxy_cache_lock on;
```
streamer45, line 197: It's fine to make the change here, but I'd also welcome a proper review of all these settings since I suspect most of us don't really know why and how they were decided.

Also pointing out we are documenting this as part of https://docs.mattermost.com/scale/high-availability-cluster.html

streamer45 (approved):
Looks good :+1: 


> Set timeout for entire response and not just response header.
It is unclear why exactly this causes the problem. So open to
thoughts and suggestions.

That would probably take a separate investigation. One random thought, although with no real proof attached, is that when originally figuring out `proxy_cache_lock` could be an issue, I noticed that we share a few timeout values with nginx (e.g. `lingering_timeout` and `proxy_cache_lock_timeout` both default to the same 5s) and given the high concurrency and low latencies it's often a race fest in terms of which side actually times out first. Depending on that, I could see the client behaving slightly differently.

agnivade:
>  I noticed that we share a few timeout values with nginx (e.g. lingering_timeout and proxy_cache_lock_timeout both default to the same 5s

Wow, good point there. Let me take a deeper look.

agnivade:
Bumped the `lingering_timeout`. `proxy_cache_lock_timeout` is a noop now since we don't have locking.

agarciamontoro (approved):
Thanks!

I'd love to sit down one morning and jot down every timeout we have and how they interact. Someday, I guess.

PR #767, opened 2024-07-18
Allow binaries with specific Mattermost builds; don't fail on license check

cpoile:
#### Summary
- I had some troubles testing old Mattermost builds (7.10), things kept failing after I put in my custom Mattemost build. Then I saw that if you put in a custom binary it would use the `latest` MM build. So I'm adding flexibility to specify your base build, if you need to.
- Note, this will break previous `deployment.json`s if you were using a `file://` for `MattermostDownloadURL`, but it will give you an error telling you what to do to fix it. Not sure if you really want to avoid that or not.
- Also changed the license check to log the error instead of exit out, since the license file I had to use was not passing the check for whatever reason. I could keep that as a local change, so tell me if you want it removed.

#### Ticket Link
- none

### DIFF
--- a/loadtest/control/simulcontroller/config.go
+++ b/loadtest/control/simulcontroller/config.go
@@ -27,7 +27,7 @@ type Config struct {
 func ReadConfig(configFilePath string) (*Config, error) {
 	var cfg Config
 
-	if err := defaults.ReadFrom(configFilePath, "./config/simulcontoller.json", &cfg); err != nil {
+	if err := defaults.ReadFrom(configFilePath, "./config/simulcontroller.json", &cfg); err != nil {
 		return nil, err
 	}
 
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -53,13 +53,12 @@ type Config struct {
 	ExternalBucketSettings ExternalBucketSettings
 	// ExternalAuthProviderSettings contains the settings for configuring an external auth provider.
 	ExternalAuthProviderSettings ExternalAuthProviderSettings
-	// URL from which to download the Mattermost release.
+	// MattermostDownloadURL supports the following use cases:
+	// 1. If it is a URL, it should be the Mattermost release to use.
+	// 2. If it is a file:// uri pointing to a binary, use the latest Mattermost release and replace
+	//    its binary with the binary pointed to by the file:// uri.
+	// 3. If it is a file:// pointing to a tar.gz, use that as the Mattermost release.
 	MattermostDownloadURL string `default:"https://latest.mattermost.com/mattermost-enterprise-linux" validate:"url"`
-	// Point to a local binary path if the user wants to run the loadtest
-	// on a custom build. The path should be prefixed with "file://". In that case,
-	// only the binary gets replaced, and the rest of the build comes
-	// from the release at MattermostDownloadURL
-	MattermostBinaryPath string `default:"" validate:"url"`
 	// Path to the Mattermost EE license file.
 	MattermostLicenseFile string `default:"" validate:"file"`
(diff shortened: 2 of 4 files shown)


agnivade:
Hmm .. this is indeed a breaking change in concept. We didn't have much of a use case testing UI, so just copy pasting the binary worked most of the times. But blindly using the `latest` stable release with any binary version is also not a good idea.

Instead of providing a base build, I am wondering if we can allow both a simple binary or a tar.gz file in the build url? Currently, we check if the prefix is `file://` then it has to be a binary. But we can remove that check to allow a tar.gz as well.

In that case, if it's a:
- binary, then do binary + latest release.
- tar.gz, then untar the whole thing
- URL, then untar from the URL.

This way existing logic works fine, and we can just start using the .tar.gz.

agarciamontoro (commented):
I'm not against the binary/whole build differentiation (although I like Agniva's idea to keep current configs valid), but I would like to understand better the use-case for accepting licenses that do not pass the validation check. That early check is quite useful to avoid realizing the license is wrong only after the whole deploying phase, which can be quite time-consuming.

cpoile:
I considered having doing what you suggested @agnivade but then I thought it was added burden, and a case of YAGNI. First, until now I think you've only needed to make modifications to the binary and it's been fine to just use the release package, right? So I didn't want to add even more work for my own edge case. Second, if we find in the future that there's something in the tar bundle that we would also want to modify, _then_ it would make sense to upload a new dist package. But for now, keep it simple and update the binary. I think we can leave that bigger change for the future if we ever end up needing it.

> I would like to understand better the use-case for accepting licenses that do not pass the validation check. That early check is quite useful to avoid realizing the license is wrong only after the whole deploying phase, which can be quite time-consuming.

It does still output the error, it just doesn't stop on that. But I'm fine taking that part out, that was kind of an issue with the license file I got from Maria I think. Done.

agnivade:
> So I didn't want to add even more work for everyone else just for my own edge case. 

Can you clarify where you are seeing "more work"? Do you mean "more work" in the code vs "more work" for the user? If we just allow for binary or tar.gz file, then the existing approach works as-is. Whereas in the current approach, passing binaries would throw an error.

cpoile:
> Can you clarify where you are seeing "more work"? Do you mean "more work" in the code vs "more work" for the user? If we just allow for binary or tar.gz file, then the existing approach works as-is. Whereas in the current approach, passing binaries would throw an error.

@agnivade I was referring to more work for the end user (because now they'd have to bundle everything instead of one binary... but I guess it's not that big of a deal). I tried it and it looks good.

agnivade:
> I was referring to more work for the end user

Yeah, it's just a single `Makefile` command really. 

> I'm not a huge fan of overloading this MattermostDownloadURL, but I guess we're stuck with it.

It's a bit more magic, I agree. But I wanted to preserve the existing use case without any change, and still let the new use case be unlocked. Since the new use case will mostly be used by internal devs, it seemed like worth the cost.

loadtest/control/simulcontroller/config.go
```diff
@@ -27,7 +27,7 @@ type Config struct {
 func ReadConfig(configFilePath string) (*Config, error) {
 	var cfg Config
 
-	if err := defaults.ReadFrom(configFilePath, "./config/simulcontoller.json", &cfg); err != nil {
+	if err := defaults.ReadFrom(configFilePath, "./config/simulcontroller.json", &cfg); err != nil {
```
agnivade, line 30: :facepalm: 

Just how long has this remained unseen?

agnivade (approved):

agarciamontoro (approved):
Looks good, thanks!

PR #778, opened 2024-08-01
Properly use EC2 API to get the proxy details

agnivade:
Previously, there was a bug in the logic. Do it properly now.

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -22,6 +22,9 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/assets"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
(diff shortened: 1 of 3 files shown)


agnivade:
Will run a test for this now. But if you want to try it now @agarciamontoro, please feel free.

deployment/terraform/create.go
```diff
@@ -762,22 +785,22 @@ func (t *Terraform) setupProxyServer(extAgent *ssh.ExtAgent) {
...
-				rxQueueSize = 8192
-			}
+
+		info, err := t.getProxyInstanceInfo()
+		if err != nil {
+			mlog.Error("Error while getting proxy info", mlog.Err(err))
+			return
+		}
+
+		if *info.MemoryInfo.SizeInMiB >= 32768 { // 32GiB
```
streamer45, line 795: Could we add a note to explain to which instance classes this would apply?

agnivade:
Confirmed that it works.

streamer45 (approved):

PR #781, opened 2024-08-07
feat: allow disabling the proxy server and customize client server url

fmartingr:
#### Proxy configuration
- Adds new `ProxyInstanceCount` configuration that allows to disable the proxy server by setting it to `0`. Only allows `1` (which is the same behaviour as having `AppInstanceCount > 1` before, that does not work like that anymore).
- Added a new `ServerURL` configuration parameter that allows a customizable site url only for the client used in the agents, in order to override other settings in the config.

#### General
- Fixed a bug where a channel didn't have enough buffer capacity to allocate possible errors
- Added the proxy server IP to `deployment info`.

With this an external reverse proxy can be used, like an AWS ALB, by disabling the default one and pointing the custom reverse proxy to the application nodes.


deployment/config.go
```diff
@@ -41,6 +41,8 @@ type Config struct {
 	AgentInstanceType string `default:"c7i.xlarge" validate:"notempty"`
 	// Logs the command output (stdout & stderr) to home directory.
 	EnableAgentFullLogs bool `default:"true"`
+	// Number of proxy instances.
+	ProxyInstanceCount int `default:"1" validate:"range:[0,1]"`
```
agnivade, line 45: Let's keep default 0

deployment/terraform/utils.go
```diff
@@ -276,10 +277,15 @@ func (t *Terraform) getAsset(filename string) string {
...
 // getServerURL returns the URL of the server to be used for testing.
 // server URL priority:
-// 1. SiteURL
-// 2. Proxy IP
-// 3. First app server IP
+// 1. ServerURL
+// 2. SiteURL
+// 3. Proxy IP
+// 4. First app server IP
 func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
```
agnivade, line 284: Can we update the test case in `TestGetServerURL`?

agnivade (commented):
This was easier than I thought :)

deployment/terraform/utils.go
```diff
@@ -276,10 +277,15 @@ func (t *Terraform) getAsset(filename string) string {
 
 // getServerURL returns the URL of the server to be used for testing.
 // server URL priority:
-// 1. SiteURL
-// 2. Proxy IP
-// 3. First app server IP
+// 1. ServerURL
+// 2. SiteURL
+// 3. Proxy IP
+// 4. First app server IP
```
agarciamontoro, line 283: This is starting to become very interesting :joy:

agarciamontoro (changes requested):
Looks great! Simple and to the point, I like it. The only concern I have is that the name `ProxyInstanceCount` implicitly offers the possibility to use more than 1 instance. Could we use something like `EnableProxy` instead?

Also, could we add a FAQ describing succinctly how to setup an external proxy using the new settings?

Thanks for your work!

agnivade:
> The only concern I have is that the name ProxyInstanceCount implicitly offers the possibility to use more than 1 instance. Could we use something like EnableProxy instead?

Actually, in the future, we might be looking at a point which uses multiple proxies. So this feels like the right choice. LMK what you think.

fmartingr:
> > The only concern I have is that the name ProxyInstanceCount implicitly offers the possibility to use more than 1 instance. Could we use something like EnableProxy instead?
> 
> Actually, in the future, we might be looking at a point which uses multiple proxies. So this feels like the right choice. LMK what you think.

Used `Count` to maintain the same attribute types as the other parameters and because it made sense, even if right now we only allow two values, I see a future where we may end up with multiple nginx proxies deployed.

agarciamontoro:
I'm usually of the opinion that future changes should be tackled in the future, not now. That said, I don't have a super strong opinion on this because we do validate it is either 0 or 1, so if you both think it's better, let's go ahead with it, no problem.

agarciamontoro:
The FAQ documenting how to setup an external proxy is still something I'd like to add before we merge. It doesn't need to be incredibly detailed, just mentioning the two settings that need to be configured.

fmartingr:
**IMPORTANT**: Keep in mind that for this to work I have downgraded the ES dependency here: https://github.com/mattermost/mattermost-load-test-ng/pull/781/commits/8409ed56a5b8144261a95c187423e2db028f0537

agnivade (approved):
Looks good. Can we revert the ES dependency change? We should look into that in a separate PR.

docs/faq.md
```diff
@@ -111,6 +111,14 @@ If you want to stress test this specific job during a load-test, you can use the
     }
 }
 ```
+
+### Can I use a custom load balancer (like an ALB/NLB) in front of the Mattermost server?
+
+Yes, it's possible by disabling the proxy server and setting up the SiteURL manually pointint to the reverse proxy. The app servers must be registered with the load balancer manually while the environment is being created, so the ideal scenario is to setup the LB/Target Group in advance and then register the instances as they become available:
```
agarciamontoro, line 117:
```suggestion
Yes, it's possible by disabling the proxy server and setting up the `SiteURL` manually pointing to the reverse proxy. The app servers must be registered with the load balancer manually while the environment is being created, so the ideal scenario is to setup the LB/Target Group in advance and then register the instances as they become available:
```

agarciamontoro (changes requested):
Thank you! About the dependency downgrade, I agree with Agniva, I think we should tackle that separately.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+		return fmt.Errorf("failed to read flag: %w", err)
+	}
+
+	updateExistingUsers, err := cmd.Flags().GetBool("update-existing-users")
+	if err != nil {
+		return fmt.Errorf("failed to read flag: %w", err)
+	}
+
+	keycloakClient := gocloak.NewClient(keycloakHost)
+	ctx := context.Background()
```
agnivade, line 95: Should this also not have a timeout?

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+					}
+				}
+
+				// Add user to migrated group in keycloak to avoid syncing them again
+				if err := keycloakClient.AddUserToGroup(ctx, token.AccessToken, deploymentConfig.ExternalAuthProviderSettings.KeycloakRealmName, *user.ID, keycloakMigratedGroup); err != nil {
+					return fmt.Errorf("failed to mark user migrated in keycloak: %w", err)
+				}
+			}
+
+			slog.Info("migrated user", slog.String("username", *user.Username))
```
agnivade, line 179: Let's keep using our own `mlog` for consistency.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+
+			if len(users) == 0 {
```
agnivade, line 181: Shouldn't this be `len(users) < perPage`? Also, shouldn't this be outside this loop and just below L130?

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+		users, err := keycloakClient.GetUsers(requestCtx, token.AccessToken, deploymentConfig.ExternalAuthProviderSettings.KeycloakRealmName, gocloak.GetUsersParams{
+			First: &start,
+			Max:   &perPage,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to get users from keycloak: %w", err)
+		}
+
+		for _, user := range users {
+			// Check if user is already migrated
```
agnivade, line 133: Can we move this logic inside the for loop to a separate function? Would be cleaner that way and also allow us to add a unit test.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read flag: %w", err)
+	}
+
+	mmClient := model.NewAPIv4Client("http://" + deploymentConfig.SiteURL)
+
+	_, _, err = mmClient.Login(cmd.Context(), deploymentConfig.AdminEmail, deploymentConfig.AdminPassword)
+	if err != nil {
+		return fmt.Errorf("failed to login to mattermost: %w", err)
+	}
```
agnivade, line 216: Seems like this logic is duplicated again. Refactor?

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+		"master", // TODO: Allow specifying the master realm
+	)
+	if err != nil {
+		return fmt.Errorf("failed to refresh keycloak token: %w", err)
+	}
+
+	usersChan := make(chan *model.User, 10000)
+	usersTxtChan := make(chan string, 10000)
+	doneChan := make(chan struct{})
+	workers := 4
```
agnivade, line 266: Bad practice to hardcode workers. Should be always `runtime.NumCPU()` capped to a limit.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package main
```
agnivade, line 4: Overall, this file is too big. Suggest moving out `RunSyncFromKeycloakCommandF` and `RunSyncFromMattermostCommandF` to their own files.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+	for i := 0; i < workers; i++ {
+		workerNumber := i + 1
+		go func() {
+			refreshTokenTicker := time.NewTicker(30 * time.Second)
+			defer refreshTokenTicker.Stop()
+
+			for {
+				select {
+				case user := <-usersChan:
+					kcUserID, err := keycloakClient.CreateUser(ctx, keycloakToken.AccessToken, keycloakRealm, gocloak.User{
```
agnivade, line 280: We are re-using the same context here, without any timeout. Intentional?

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+				return fmt.Errorf("failed to create group for user migrations in keycloak: %w", err)
+			}
+		}
+	}
+
+	start := 0
+	perPage := 100
+	for {
+		requestCtx, cancel := context.WithTimeout(context.Background(), operationTimeout)
+		defer cancel()
```
agnivade, line 123: Note that since this is a loop, cancel might not get called until the entire function exits. Not a big deal, but worth keeping in mind. Ideally, we call `cancel` as soon as the operation is done.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+							{
+								Temporary: gocloak.BoolP(false),
+								Type:      gocloak.StringP("password"),
+								Value:     gocloak.StringP(userPassword),
+							},
+						},
+					})
+					if err != nil {
+						// Ignore already existing users
+						if apiErr, ok := err.(*gocloak.APIError); ok && apiErr.Code == 409 {
```
agnivade, line 295: Can we do a debug log here mentioning that user exists?

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+					keycloakToken, err = keycloakClient.LoginAdmin(
+						ctx,
+						deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminUser,
+						deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminPassword,
+						"master", // TODO: Allow specifying the master realm
+					)
+					if err != nil {
+						slog.Error("failed to refresh keycloak token", slog.String("err", err.Error()))
+						close(doneChan)
+						panic(err)
```
agnivade, line 326: No panic please.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+			}
+		}()
+	}
+
+	// File writter for users.txt
+	go func() {
+		usersTxtFile, err := os.OpenFile("users.txt", os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			slog.Error("failed to open users.txt file")
+			panic(err)
```
agnivade, line 341: No panics please.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+	usersTxtChan := make(chan string, 10000)
+	doneChan := make(chan struct{})
+	workers := 4
+
+	wg := sync.WaitGroup{}
+
+	// Workers to run the sync and call the MM and Keycloak APIs
+	for i := 0; i < workers; i++ {
+		workerNumber := i + 1
+		go func() {
```
agnivade, line 273: This function has a lot of code all at once. Suggest refactoring out the worker code, the file writer code in separate functions.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+				continue
+			}
+
+			if !dryRun {
+				wg.Add(1)
+				usersChan <- user
+			} else {
+				slog.Info("dry-run: would migrate user", slog.String("username", user.Username))
+			}
+		}
```
agnivade, line 388: This is the code which should be earlier. Because this is the one which is feeding the data to the consumer goroutines. Otherwise it's difficult to follow the flow of the code.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+
+		if len(users) == 0 {
```
agnivade, line 390: Same comment as earlier for this.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+		deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminUser,
+		deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminPassword,
+		"master", // TODO: Allow specifying the master realm
+	)
+	if err != nil {
+		return fmt.Errorf("failed to refresh keycloak token: %w", err)
+	}
+
+	usersChan := make(chan *model.User, 10000)
+	usersTxtChan := make(chan string, 10000)
```
agnivade, line 264: Why such random and large numbers for the channel buffer? I'd just start with the same number as workers, or probably double that to keep things proportional.

cmd/ltkeycloak/main.go
```diff
@@ -0,0 +1,468 @@
...
+						deploymentConfig.ExternalAuthProviderSettings.KeycloakAdminPassword,
+						"master", // TODO: Allow specifying the master realm
+					)
+					if err != nil {
+						slog.Error("failed to refresh keycloak token", slog.String("err", err.Error()))
+						close(doneChan)
+						panic(err)
+					}
+
+				case <-doneChan:
```
agnivade, line 329: There is a bug here. Actually 2.

1. Closing a channel won't really wait for the `return` statement to execute. So it could be that you call `close(doneChan)` and move on, but this goroutine is yet to return. 
2. You are using a buffered chan, which means that the sender side would be complete, but the consumer might still be waiting to drain the channel. So it would be incorrect to return before draining a channel.

Fixing #1 is easy. Just add yet another waitgroup, or a signalling channel to indicate that the goroutine has finished. See: https://github.com/mattermost/mattermost/blob/e666f7ccfc3575dbe9507c7c68533f717355e0c8/server/channels/app/platform/web_conn.go#L293-L294.

Fixing #2 is complicated. You have to keep reading from the channel until it's drained. I'd suggest to avoid buffered channel altogether and just keep it a simple non-buffered channel.

deployment/terraform/create.go
```diff
@@ -1063,11 +1064,58 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
...
+			return fmt.Errorf("error uploading saml-idp.crt: %s - %w", out, err)
+		}
+
+		keycloakUrl := keycloakScheme + "://" + t.output.KeycloakServer.PublicDNS + ":8080"
+
+		cfg.OpenIdSettings.Enable = model.NewBool(true)
+		cfg.OpenIdSettings.ButtonText = model.NewString("Keycloak Login")
+		cfg.OpenIdSettings.DiscoveryEndpoint = model.NewString(keycloakUrl + "/realms/" + t.config.ExternalAuthProviderSettings.KeycloakRealmName + "/.well-known/openid-configuration")
+		cfg.OpenIdSettings.Id = model.NewString(t.config.ExternalAuthProviderSettings.KeycloakClientID)
+		cfg.OpenIdSettings.Secret = model.NewString(t.config.ExternalAuthProviderSettings.KeycloakClientSecret)
```
agnivade, line 1083: Only `model.NewPointer` please.

deployment/terraform/strings.go
```diff
@@ -429,7 +429,7 @@ KC_DB_POOL_MIN_SIZE=20
 KC_DB_POOL_INITIAL_SIZE=20
 KC_DB_POOL_MAX_SIZE=200
 KC_DB=postgres
-C_DB_URL=jdbc:psql://localhost:5433/keycloak"
+KC_DB_URL=jdbc:psql://localhost:5433/keycloak
```
agnivade, line 432: oof, was this a bug?

deployment/terraform/create.go
```diff
@@ -1063,11 +1064,58 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
 			keycloakScheme = "http"
 		}
 
-		cfg.OpenIdSettings.Enable = model.NewPointer(true)
-		cfg.OpenIdSettings.ButtonText = model.NewPointer("Keycloak Login")
-		cfg.OpenIdSettings.DiscoveryEndpoint = model.NewPointer(keycloakScheme + "://" + t.output.KeycloakServer.PublicDNS + ":8080/realms/" + t.config.ExternalAuthProviderSettings.KeycloakRealmName + "/.well-known/openid-configuration")
-		cfg.OpenIdSettings.Id = model.NewPointer(t.config.ExternalAuthProviderSettings.KeycloakClientID)
-		cfg.OpenIdSettings.Secret = model.NewPointer(t.config.ExternalAuthProviderSettings.KeycloakClientSecret)
+		// Setup SAML certificate for Keycloak
+		samlIDPCert, err := os.Open(t.getAsset("saml-idp.crt"))
```
agnivade, line 1068: This file is starting to get too long. Can we move these to a separate (terraform/keycloak.go)?

agnivade (commented):
Gave it a first pass. Really great work on this :ok_hand:

deployment/terraform/strings.go
fmartingr: Yep...

fmartingr:
Good call @agarciamontoro

agnivade (approved):
Looks good

deployment/terraform/create.go
```diff
@@ -273,7 +273,7 @@ func (t *Terraform) Create(initData bool) error {
 
 	// Ingesting the DB dump and restoring the Elasticsearch snapshot can
 	// happen concurrently to reduce the deployment's total time
-	errorsChan := make(chan error, 2)
+	errorsChan := make(chan error, 5)
```
agarciamontoro, line 276: Should we keep this to 2 and return on error in the goroutines below? I think the bug here is that we don't return after detecting an error, not that the capacity of this buffered channel is 2.

agarciamontoro (changes requested):
Looks great, thank you! I left a single comment below, but wanted to discuss it before approving.

deployment/terraform/create.go
fmartingr: I'm happy either way, since we are going to stop enviornment creation I'd say is better to `return` directly.

fmartingr: Changed.

agarciamontoro (approved):
Looks great, thank you!

PR #782, opened 2024-08-12
feat: add concurrency to configureAndRunAgents

fmartingr:
Add some concurrency to agent creation for large load tests


agnivade:
This is a great improvement. Just take care to annotate every log line with the agent name, otherwise the log lines would get interleaved and might be difficult to debug in case of errors.

agarciamontoro:
Oooh, looking good :eyes: Feel free to ask for a review when it's ready!

deployment/terraform/agent.go
```diff
@@ -85,83 +86,113 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent) error {
...
 
-		mlog.Info("Starting load-test api server")
-		if out, err := sshc.RunCommand("sudo systemctl daemon-reload && sudo service ltapi restart"); err != nil {
-			return fmt.Errorf("error running command, got output: %q: %w", out, err)
-		}
+	// Run another goroutine to check when all agents have finished configuring
+	go func() {
+		wg.Wait()
+		close(wgDone)
+	}()
```
agnivade, line 184: I would restructure this a bit. Make `wg.Wait()` to be synchronous and remove the err chan. Closing the `errChan` while other goroutines are still sending data to it is actually a bug and will panic. 

What I'd suggest is:

```go
var wg sync.WaitGroup
for i, val := range agents {
  wg.Add(1)
  go func(i int, val string) {
    defer wg.Done()
    // do stuff here
    // log error as usual, but just set an atomic.Bool to true.
  }(i, val) // This is important to localize scope. Then you don't need the weird variable copying style.
}

wg.Wait()
if atomic.Bool == true, then return error. 
```

agnivade (approved):
Nicely done

deployment/terraform/agent.go
```diff
@@ -85,83 +87,118 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent) error {
...
+		agentNumber := i
+		instance := val
 
-		cmd := strings.Join(commands, " && ")
-		if out, err := sshc.RunCommand(cmd); err != nil {
-			return fmt.Errorf("error running command, got output: %q: %w", out, err)
-		}
+		go func() {
+			// Copy val into instance
+			defer wg.Done()
```
agarciamontoro, line 100: Why is the `// Copy val into instance` comment inside the goroutine?

agarciamontoro (approved):
Looks amazing, thanks!

deployment/terraform/agent.go
fmartingr: Probably I moved some lines around during development. Let me remove it.

PR #783, opened 2024-08-13
Revert go-elasticsearch version bump

agarciamontoro:
#### Summary

The newer v7.17 doesn't allow us to test proper Elasticsearch clusters, only Opensearch, so we're downgrading while we find another solution for this so that we can release the new v1.20

#### Ticket Link
--


agarciamontoro:
@fmartingr, FYI: I'm adding a new PR with only this because we need to revert it for releasing v1.20

fmartingr (approved):

streamer45 (approved):

agnivade (approved):
Aha! So this happened in our usual version bump. Nice find Alejandro!

agarciamontoro:
Actually, the find that this was the error @fmartingr's! Thank you for the reviews, guys :)

PR #785, opened 2024-08-19
MM-59994: Added the ethtool metrics

agnivade:
https://mattermost.atlassian.net/browse/MM-59994


agarciamontoro (changes requested):
Thank you! A couple of things:
- We use `[1m]` in all other expressions, should we be consistent there? In any case, I think we should update all to use `[$__rate_interval]` instead ([docs](https://grafana.com/docs/grafana/latest/datasources/prometheus/template-variables/#use-__rate_interval)) at some point (not in this PR)
- If we're using `increase`, I'd prefer to specify the time window in the panel title, something like `Count of conntrack exceeded during the last 5 minutes`.

agarciamontoro (approved):
Neat, thanks!

deployment/terraform/assets/dashboard_data.json
```diff
@@ -21,6 +21,7 @@
   "graphTooltip": 1,
   "id": 3,
   "links": [],
+  "liveNow": false,
```
streamer45, line 24: Is this disabling auto refresh?

streamer45 (approved):

deployment/terraform/assets/dashboard_data.json
agnivade: Not sure. Shouldn't matter much since it's internal anyways.

PR #788, opened 2024-08-26
MM-58729: Improve proxy cache configuration

agnivade:
Tested with various combinations of settings. None of them showed a major improvement. Therefore, settled on the very basic changes.

Full thread here: https://community.mattermost.com/core/pl/rjbg3yr3mfrrfrcq7zk3eo6gky

https://mattermost.atlassian.net/browse/MM-58729
https://mattermost.atlassian.net/browse/MM-59871

### DIFF
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -195,9 +195,11 @@ proxy_pass http://backend;
 const nginxCacheCommonConfig = `
 proxy_cache mattermost_cache;
 proxy_cache_revalidate on;
-proxy_cache_min_uses 2;
-proxy_cache_use_stale timeout;
+proxy_cache_min_uses 1;
+proxy_cache_use_stale updating;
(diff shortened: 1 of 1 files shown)


agarciamontoro (commented):
Thanks for the investigation! I wonder that given that none showed improvements, do we want/need to change them? I'm ok with  the change of `proxy_cache_lock_timeout`, since we wanted to make timeouts easier to debug, but what about `proxy_cache_min_uses`? I'm not opposed to it, but I'm not entirely sure of the benefit.

agnivade:
The benefit of `min_uses` would be that it populates the cache faster, and avoids an extra request to the server. At high scales, this _should_ make a difference. Feels like a safe optimization IMO, but only 1/5.

lieut-data:
Closing old change, not likely to take action directly from here, but will revisit from first principles if needed.

PR #789, opened 2024-08-27
MM-57876: Use opensearch only

agnivade:
https://mattermost.atlassian.net/browse/MM-57876

### DIFF
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -354,8 +354,16 @@ func (c *Config) validateElasticSearchConfig() error {
 
 	}
 
-	if !strings.HasPrefix(c.ElasticSearchSettings.Version, "Elasticsearch") && !strings.HasPrefix(c.ElasticSearchSettings.Version, "OpenSearch") {
-		return fmt.Errorf("Incorrect engine version: %s. Must start with either %q or %q", c.ElasticSearchSettings.Version, "Elasticsearch", "OpenSearch")
+	if !strings.HasPrefix(c.ElasticSearchSettings.Version, "OpenSearch") {
+		return fmt.Errorf("Incorrect engine version: %s. Must start with %q", c.ElasticSearchSettings.Version, "OpenSearch")
+	}
+
+	if c.ElasticSearchSettings.SnapshotRepository == "" {
+		return fmt.Errorf("Empty SnapshotRepository. Must supply a value")
+	}
+
+	if c.ElasticSearchSettings.SnapshotName == "" {
+		return fmt.Errorf("Empty SnapshotName. Must supply a value")
 	}
 
 	return nil
--- a/deployment/opensearch/opensearch.go
+++ b/deployment/opensearch/opensearch.go
@@ -1,20 +1,19 @@
-package elasticsearch
+package opensearch
 
 import (
 	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
-	"io"
 	"strings"
 
-	es "github.com/elastic/go-elasticsearch/v7"
-	"github.com/elastic/go-elasticsearch/v7/esapi"
 	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
+	"github.com/opensearch-project/opensearch-go/v4"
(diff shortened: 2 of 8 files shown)


deployment/config.go
```diff
@@ -354,8 +354,16 @@ func (c *Config) validateElasticSearchConfig() error {
...
+		return fmt.Errorf("Incorrect engine version: %s. Must start with %q", c.ElasticSearchSettings.Version, "OpenSearch")
+	}
+
+	if c.ElasticSearchSettings.SnapshotRepository == "" {
+		return fmt.Errorf("Empty SnapshotRepository. Must supply a value")
+	}
+
+	if c.ElasticSearchSettings.SnapshotName == "" {
+		return fmt.Errorf("Empty SnapshotName. Must supply a value")
 	}
```
agarciamontoro, line 367: Can we create a ticket to allow deploying Opensearch without a snapshot?

deployment/opensearch/opensearch.go
```diff
@@ -65,14 +63,13 @@ type Repository struct {
 
 // ListRepositories returns the list of repositories registered in the server
 func (c *Client) ListRepositories() ([]Repository, error) {
-	req := esapi.SnapshotGetRepositoryRequest{}
-	repositoriesResponse := make(map[string]repositoryResponse)
-	if err := c.get(req, &repositoriesResponse); err != nil {
+	resp, err := c.client.Snapshot.Repository.Get(context.Background(), nil)
```
agarciamontoro, line 66: Oh, this is much better than before :joy:

deployment/terraform/create.go
```diff
@@ -994,9 +994,6 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
 	cfg.SqlSettings.MaxIdleConns = model.NewPointer(100)
 	cfg.SqlSettings.MaxOpenConns = model.NewPointer(100)
 	cfg.SqlSettings.Trace = model.NewPointer(false) // Can be enabled for specific tests, but defaulting to false to declutter logs
-	if t.output.HasElasticSearch() {
-		cfg.SqlSettings.DisableDatabaseSearch = model.NewPointer(true)
-	}
```
agarciamontoro, line 999: Thanks for this! Can we split it to another PR, thought? It feels like we may need to point to this change in the future, and it's quite hidden here.

go.mod
```diff
@@ -33,6 +33,7 @@ require (
 	github.com/graph-gophers/graphql-go v1.5.1-0.20230110080634-edea822f558a
 	github.com/mattermost/mattermost/server/public v0.1.7-0.20240806035841-540febd866aa
 	github.com/mattermost/mattermost/server/v8 v8.0.0-20240726090344-5547504c1d68
+	github.com/opensearch-project/opensearch-go/v4 v4.1.0
```
agarciamontoro, line 36: I see this one was added, but the `github.com/elastic/go-elasticsearch/v7` dependency was not removed. Is it still needed?

agarciamontoro (changes requested):
Thank you for this! The new client is soooo much better, it really simplifies the code.

Some additional questions:
- Have we tested this? Do you want a test from my side?
- Do we need to update the current snapshot of the 100M posts database dump? Or can we use the same one?

go.mod
agnivade: Yeah `go mod tidy` didn't remove it. We still use some elasticsearch types for the requests I believe.

deployment/terraform/create.go
agnivade: ok

agnivade:
> Have we tested this? Do you want a test from my side?

I have tested this.

> Do we need to update the current snapshot of the 100M posts database dump? Or can we use the same one?

I have restored the existing ES snapshot we have, and the restore works fine. Although I didn't use it with the 100M one, I don't think it would be an issue. The main thing to check was the restore and it works fine.

deployment/config.go
agnivade: https://mattermost.atlassian.net/browse/MM-60336

agarciamontoro (approved):
Thank you!

streamer45 (approved):
Looks good!

PR #790, opened 2024-08-28
[MM-60255] Add basic retry logic to provisioning scripts

streamer45:
#### Summary

I couldn't reproduce the issue (I tried a bunch of times with lots of agents), but most reported errors suggested a problem with `apt-get` commands (exit code `100`), possibly due to some network (i.e. remote) flakiness. Regardless, I opted to implement some basic retry logic which should take care of most temporary issues. 

To do this, we move away from inline scripts and use dedicated provisioning files.

I tested deploying all the affected instance types, both in happy and failure cases and didn't spot any glaring issues. That said, changes may welcome some soaking time.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60255


agnivade (approved):
Sad we'll never find out the true cause

agarciamontoro (changes requested):
Can we use something like [`set -euo pipefail`](https://wizardzines.com/comics/bash-errors/) in all scripts? That way we can probably remove all the `&& \` as well.

As an aside for the future: we should at some point move away from scripts altogether and use ad-hoc AMIs that have everything we want installed and all services already enabled. It would be more work when we want to change some version of a package or we want to install a new one, but it would be static and more robust. Thoughts?

streamer45:
> Can we use something like [`set -euo pipefail`](https://wizardzines.com/comics/bash-errors/) in all scripts? That way we can probably remove all the `&& \` as well.

Let me look into it. I am worried a failure could cause the whole thing to exit which is not what we want. 

> As an aside for the future: we should at some point move away from scripts altogether and use ad-hoc AMIs that have everything we want installed and all services already enabled. It would be more work when we want to change some version of a package or we want to install a new one, but it would be static and more robust. Thoughts?

I think we should simply use the right tool for the job (configuration, system state); something like [Ansible playbooks](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html) would work. Having static AMIs could work as well, but it may be a little hard-core since (as you said) it would then be painful to make any changes.

streamer45:
@agarciamontoro I added `set -euo pipefail` but we cannot remove the `&&` because we need to be able to repeat the `until` body.

agarciamontoro:
> but we cannot remove the && because we need to be able to repeat the until body.

Ah, of course!

> I think we should simply use the right tool for the job (configuration, system state); something like [Ansible playbooks](https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_intro.html) would work. Having static AMIs could work as well, but it may be a little hard-core since (as you said) it would then be painful to make any changes.

Maybe [cloud-init](https://developer.hashicorp.com/terraform/tutorials/provision/cloud-init) is enough for this? I haven't used any of those, though, so not sure what's the best tool for the job.

agarciamontoro (approved):
Thank you!

streamer45:
> Maybe [cloud-init](https://developer.hashicorp.com/terraform/tutorials/provision/cloud-init) is enough for this? I haven't used any of those, though, so not sure what's the best tool for the job.

I think it depends on which guarantees it gives us. Ansible ensures everything is idempotent. If we are just wrapping a script, then we may as well keep doing what we are doing.

PR #791, opened 2024-08-30
Remove incorrect DB search config

agnivade:
We were incorrectly disabling a config, which
wasn't necessary.


streamer45 (approved):

agarciamontoro (approved):
Thank you!

PR #792, opened 2024-08-30
Restore the saml-idp.crt file in assets

agnivade:
Without this line, the deployment fails because
it cannot do t.getAsset


agnivade:
I also see another asset `keycloak-users.txt`, which is not restored. Will leave it to @fmartingr

fmartingr:
`keycloak-users.txt` is a file created by the process if the `GenerateUsersCount` is higher than zero, unsure if this applies. I haven't used this for any asset IIRC, what is this used for? And how it could be that it didn't fail for me before?

agnivade:
Okay, if it's not needed always, then it's fine. But we do need the `saml-idp.crt`. Maybe you can test with a fresh setup? Perhaps you had the `saml-idp.crt` file somehow already in your terraform state dir?

fmartingr:
> Okay, if it's not needed always, then it's fine. But we do need the `saml-idp.crt`. Maybe you can test with a fresh setup? Perhaps you had the `saml-idp.crt` file somehow already in your terraform state dir?

I just checked and I had it in my state dir. what are the ways a file can end up there? I started from scratch two weeks ago.

agnivade:
Hmm no idea :man_shrugging: But anyways, do test with a fresh setup and see.

agarciamontoro (approved):

fmartingr (approved):
Tested on a cluster creation from scratch, 👌

agarciamontoro:
@agnivade, can you fix the conflicts to get this merged for the new v1.21 release?

PR #795, opened 2024-09-04
fix: use proxy_instance_count for the security group

fmartingr:
#### Summary

Fixes a bug created in #781 where the security group for the proxy was still using the `AppInstanceCount` as `count` attribute.


agarciamontoro (approved):
Oh, that makes sense. Thanks!

agnivade (approved):

PR #796, opened 2024-09-09
MM-60439: Set websocket scope early

agnivade:
Because there is a return statement which gets executed
if the user isn't part of any teams, this optimization
won't kick in if that's the case.

To let the optimization happen in all cases, we move the
line earlier.

https://mattermost.atlassian.net/browse/MM-60439

### DIFF
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -1393,14 +1393,6 @@ func (c *SimulController) initialJoinTeam(u user.User) control.UserActionRespons
 		return resp
 	}
 
-	team, err := c.user.Store().CurrentTeam()
-	if err != nil {
-		return control.UserActionResponse{Err: control.NewUserError(err)}
-	} else if team == nil {
-		// only join a team if we are not in one already.
-		return c.joinTeam(c.user)
-	}
-
 	if c.user.Store().FeatureFlags()["WebSocketEventScope"] {
 		// Setting the active thread to empty to allow the optimization to kick in early.
 		// We don't do this in the webapp to keep the code simple, but it's okay to do this in load-test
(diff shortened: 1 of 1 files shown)


agarciamontoro (approved):
Great catch, thanks!

streamer45:
Would any of this be happening with a real client? According to the comment that may not be the case.

agnivade:
No, as the comment says, all this is just to trigger the optimization early on. The re-arranging of the code bears no relation to that.

streamer45:
> No, as the comment says, all this is just to trigger the optimization early on. The re-arranging of the code bears no relation to that.

Yeah, I guess I am just trying to understand why we'd want to boost "lab" performance over real world. We usually want to be on the underestimating side, unless there's a very good reason not to.

agnivade:
We had the same discussion here: https://github.com/mattermost/mattermost-load-test-ng/pull/678#discussion_r1456023606. I think we can remove this optimization once we improve the load test to differentiate between loading RHS thread and global thread.

There's also the fact that we are not setting this default state from the server side, rather letting the client set the default state. This was done to let older mobile clients work with the new server code. When the minimum mobile version is bumped up higher than the version where this was introduced, we can get rid of this entire logic, and simply set the default from the server side itself.

So, some work to do, but it's mainly due to some limitations on the load test side. I wouldn't say we are taking any short cuts to boost performance.

streamer45:
Thanks for the context dump, it was very helpful.

> When the minimum mobile version is bumped up higher than the version where this was introduced, we can get rid of this entire logic, and simply set the default from the server side itself.

Do we know when that will happen?

streamer45 (approved):

agnivade:
> Do we know when that will happen?

Actually, scratch that, we haven't even implemented websocket scope in mobile. We need to do that first. :man_facepalming:

PR #797, opened 2024-09-09
Default proxy count to 1

agarciamontoro:
#### Summary
Running a 2-node deployment, I got confused because only one node was getting requests, although both of them were healthy. It took me a while to understand that I was running proxy-less! This happened because I was running an older config with a newer load-test version, so I think it's safer to default the new `ProxyInstanceCount` to 1. This would avoid breaking other people's configurations.

I've also added it to both sample files and the docs.

#### Ticket Link
--

### DIFF
--- a/docs/config/deployer.md
+++ b/docs/config/deployer.md
@@ -154,6 +154,12 @@ The type of EC2 instance for the Job Server. See type [here](https://aws.amazon.
 
 Allows to log the agent service command output (`stdout` & `stderr`) to home directory.
 
+## ProxyInstanceCount
+
+*int*
+
+Number of proxy instances to run. Right now, only values `0` and `1` are allowed. Check [this FAQ](../faq.md#can-i-use-a-custom-load-balancer-like-an-albnlb-in-front-of-the-mattermost-server) for more information.
+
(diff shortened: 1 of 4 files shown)


agnivade (commented):
Can't we do a check during deployment if the number of nodes is greater than 1, and if no external proxy is specified, and proxy_instance_count is 0, then throw an error?

fmartingr (approved):
Agree that this should be the default, but the configuration check could be added to be more future-proof.

agarciamontoro:
> Can't we do a check during deployment if the number of nodes is greater than 1, and if no external proxy is specified, and proxy_instance_count is 0, then throw an error?

Added a validation step for this. Let me know your thoughts :)

agnivade (approved):
Nicely done!

docs/config/deployer.md
```diff
@@ -154,6 +154,12 @@ The type of EC2 instance for the Job Server. See type [here](https://aws.amazon.
 
 Allows to log the agent service command output (`stdout` & `stderr`) to home directory.
 
+## ProxyInstanceCount
```
fmartingr, line 157: 🙇

PR #798, opened 2024-09-12
MM-60020: Cloudwatch exporter

agarciamontoro:
#### Summary
This PR adds a new service to the metrics instance: [YACE](https://github.com/nerdswords/yet-another-cloudwatch-exporter), an exporter for retrieving Cloudwatch metrics into Prometheus. I needed to make a couple of adjustments to make this work:
- I added a new `ClusterName` tag to all resources, since YACE only discovers tagged resources. This will also help us analyze the cost of individual resources and whole clusters after we run large load-tests
- I also added a missing `Name` tag to RDS clusters and instances
- I added a couple of new panels in the default and the Elasticsearch dashboards (see screenshots below)
- I converted the old `dashboard_data.json` file into a template file, and renamed it to `default_dashboard_tmpl.json`

A couple of important issues to note with Cloudwatch metrics:
- These metrics are inherently delayed due to how AWS export them.
- Scraping the Cloudwatch API is not free: $0.01 per 1,000 metrics requested (see the Metrics tab [in this doc](https://aws.amazon.com/cloudwatch/pricing/), the entry for `GetMetricData`).

Considering these two points, YACE is right now configured to get a single data point every 5 minutes, with a 5 minute delay. That is: YACE scrapes Cloudwatch at times T, T+5m, T+10m... and, respectively, get data from T-5m, T, T+5m... This is all configured with the `Period`, `Length` and `Delay` options, which are all set to 300 seconds. [This issue comment](https://github.com/nerdswords/yet-another-cloudwatch-exporter/issues/865#issuecomment-1517656590) is very helpful to understand how that works. We can tune this in the future, probably having more resolution, but I wanted to be conservative at the beginning.

##### New section and panels in default dashboard
image
##### New section and panels in Elasticsearch dashboard
image
##### New panel at the very beginning for instance health checks
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60020


deployment/terraform/assets/cluster.tf
```diff
@@ -198,17 +262,21 @@ EOF
 resource "aws_elasticache_cluster" "redis_server" {
   cluster_id           = "${var.cluster_name}-redis"
   engine               = "redis"
-  node_type            = "${var.redis_node_type}"
+  node_type            = var.redis_node_type
```
agnivade, line 265: Thanks for fixing my mistakes :pray:

deployment/terraform/strings.go
```diff
@@ -319,6 +322,147 @@ Group=ubuntu
...
+        - name: DiskReadOps
+          statistics: [Sum]
+        - name: DiskWriteOps
+          statistics: [Sum]
+        - name: StatusCheckFailed
+          statistics: [Sum]
+        - name: StatusCheckFailed_Instance
+          statistics: [Sum]
+        - name: StatusCheckFailed_System
+          statistics: [Sum]
```
agnivade, line 445: I don't think we need the EC2 metrics. 0/5

agnivade (approved):
Great work! Left a non-blocking comment for your consideration.

deployment/terraform/assets/cluster.tf
agarciamontoro: It was actually my editor :joy:

deployment/terraform/strings.go
agarciamontoro: Yeah, I added those when I was testing the whole thing and did not remove them even though I'm not using them in any of the panels. The only interesting ones may be the `StatusCheckFailed` ones to monitor if an instance is down. Let me see if I can add a panel with that info and I can remove the other ones.

deployment/terraform/assets/cluster.tf
agnivade: Thank you emacs :pray:

agarciamontoro:
Added a new panel plotting the health check of all instances:
image

cpoile (approved):
I'm not sure what I can add, looks good. :)

deployment/terraform/assets/cluster.tf
```diff
@@ -77,6 +82,63 @@ resource "aws_instance" "app_server" {
...
+
+resource "aws_iam_role" "metrics_role" {
+  name               = "${var.cluster_name}-metrics-role"
+  assume_role_policy = data.aws_iam_policy_document.metrics_assume_role.json
+}
+
+resource "aws_iam_instance_profile" "metrics_profile" {
+  name = "metrics_profile"
+  role = aws_iam_role.metrics_role.name
+}
```
streamer45, line 107: @agarciamontoro Heads up this seems to be causing issues on my side, apparently conflicting with other deployments. Getting the following error when deploying:

```
info  [2024-09-25 15:49:43.890 -06:00] │ Error: creating IAM Instance Profile (metrics_profile): operation error IAM: CreateInstanceProfile, https response error StatusCode: 409, RequestID: 5a5d50ea-fa2d-411b-b185-2009a4975393, EntityAlreadyExists: Instance Profile metrics_profile already exists. caller="terraform/engine.go:97"
```

agarciamontoro: Of course! :facepalm: Thank you for catching this!

PR #800, opened 2024-09-16
MM-60548: Use DB instance endpoints directly

agnivade:
AWS enforces a limit of 5 custom endpoints per cluster: https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Aurora.Endpoints.Custom.html

I am not sure how we didn't run into this in our
previous ceiling tests, but anyways I got blocked
by this randomly.

Turns out that we don't need custom endpoints in the first place, because we were anyways mapping 1:1 from endpoint to instance. So we can directly use the
instance endpoint.

This also allows us to use the writer field in the instance struct to let us cleanly identify the writer rather than looking for an instance prefix.

https://mattermost.atlassian.net/browse/MM-60548


agarciamontoro (commented):
Oh, this is way nicer! Do we know why we were using custom endpoints in the first place?

agnivade:
Yeah, I've been trying to remember myself. I think it was an oversight, but I'll let Claudio chime in if there was a specific reason.

streamer45:
Can't recall if there was a specific reason, although, as mentioned in https://hub.mattermost.com/private-core/pl/7h18pessxidwfk6y7ocjz5cqxo, I am not sure what would happen with regards to failover (e.g. writer instance going down).

At a high level, I don't think these changes would be any more problematic than the former approach but it's something to keep in mind as I remember back in the days that in case of a failure Aurora would run some failover logic that would mess up the deployment.

agnivade:
The current approach anyways doesn't do anything about failover because the endpoint is mapped directly to a single instance. Failover only comes into play when you use cluster endpoints, which we are not using here. But agree that these changes isn't any worse than the earlier approach. Although, I do like the benefit of using the `writer` field rather than relying of `rd`/`wr` prefixes.

streamer45 (approved):

agarciamontoro (approved):

PR #801, opened 2024-09-16
Update Redis dashboard

agnivade:
- Migrate some panels to the new format.
- Change the crosshair to show all series.


agarciamontoro (approved):
Reviewing the JSON here is pretty much impossible :joy: If you can add a screenshot for these types of PRs, that would make it easier in my opinion. In any case, approving already!

agnivade:
There's no real UI changes. Just that the new panels look a bit better.

And I don't have a screenshot currently. I destroyed the setup after exporting the dashboard. I have a new setup, but it's the old dashboard.

deployment/terraform/assets/redis_dashboard_data.json
```diff
@@ -1,25 +1,32 @@
...
       "id": "grafana",
       "name": "Grafana",
-      "version": "6.5.2"
+      "version": "10.2.3"
     },
     {
       "type": "panel",
       "id": "graph",
-      "name": "Graph",
+      "name": "Graph (old)",
```
streamer45, line 29: Is this expected?

streamer45 (approved):
Cool

deployment/terraform/assets/redis_dashboard_data.json
agnivade: Yep, it is. The new panel is called "timeseries", as you can see below the "type" of panel has changed. And the old one is called "graph".

PR #802, opened 2024-09-17
feat(ltkeycloak): added flag to force migrate all users

fmartingr:
#### Summary
Added a `--migrate-all-users` flag that forces migration of all users, which changes two things:
- Skips checking if the `authmethod` for the Mattermost user is `password`
- When creating a Keycloak user, instead of directly failing if the user exists, when the flag is set we will try to retrieve that user and update the Mattermost user accordingly.
- Changed the `UpdateUser` to `UpdateUserAuth` to work with newer Mattermost versions.

This flag is required for environment where the data loaded into Mattermost already points users to an external auth provider. Even if the data was wrong and wasn't working, the tool couldn't migrate those users as it was.

### DIFF
--- a/cmd/ltkeycloak/README.md
+++ b/cmd/ltkeycloak/README.md
@@ -28,7 +28,7 @@ Global Flags:
       --dry-run                  perform a dry run without making any changes
       --keycloak-host string     keycloak host (default "http://localhost:8484")
       --mattermost-host string   The Mattermost host to migrate users from
-      --migrate-all-users        Migrate all users ignoring their current auth method
+      --force-migrate            Migrate all users ignoring their current auth method
 ```
 
 #### Example
--- a/cmd/ltkeycloak/README.md
+++ b/cmd/ltkeycloak/README.md
@@ -15,7 +15,11 @@ ltkeycloak [command] [flags]
 This command will sync users from a Mattermost server to a Keycloak server, setting the same password for all users and creating a txt file to be used with the loadtest.
 
 ```
+Usage:
+  ltkeycloak sync from_mattermost [flags]
+
 Flags:
+  -h, --help                          help for from_mattermost
       --keycloak-realm string         The Keycloak realm to migrate users to (default "master")
       --set-user-password-to string   Set's the user password to the provided value (default "testpassword")
 
@@ -24,6 +28,7 @@ Global Flags:
       --dry-run                  perform a dry run without making any changes
       --keycloak-host string     keycloak host (default "http://localhost:8484")
       --mattermost-host string   The Mattermost host to migrate users from
+      --migrate-all-users        Migrate all users ignoring their current auth method
 ```
 
 #### Example
--- a/cmd/ltkeycloak/from_mattermost.go
+++ b/cmd/ltkeycloak/from_mattermost.go
@@ -266,7 +266,7 @@ func RunSyncFromMattermostCommandF(cmd *cobra.Command, _ []string) error {
 		return fmt.Errorf("failed to refresh keycloak token: %w", err)
 	}
 
-	migrateAllUsers, err := cmd.Flags().GetBool("force-migrate")
+	forceMigrate, err := cmd.Flags().GetBool("force-migrate")
 	if err != nil {
 		return fmt.Errorf("failed to read flag: %w", err)
 	}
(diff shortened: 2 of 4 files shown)


cmd/ltkeycloak/README.md
```diff
@@ -24,6 +28,7 @@ Global Flags:
       --dry-run                  perform a dry run without making any changes
       --keycloak-host string     keycloak host (default "http://localhost:8484")
       --mattermost-host string   The Mattermost host to migrate users from
+      --migrate-all-users        Migrate all users ignoring their current auth method
```
agnivade, line 31: Maybe `--force-migrate` sounds better given that we are ignoring the current auth method? 0/5

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -40,12 +40,43 @@ func migrateUser(worker *workerConfig, user *model.User) error {
...
+				)
+				if err != nil {
+					mlog.Error("failed to get user from keycloak", mlog.String("err", err.Error()))
+					worker.errorsChan <- err
+					return err
+				}
+
+				if len(users) != 1 {
+					err = fmt.Errorf("expected 1 user, got %d", len(users))
+					mlog.Error("got more users than expected in keycloak", mlog.String("username", user.Username), mlog.Int("users_count", len(users)))
```
agnivade, line 65: We typically avoid logging and returning an error. If we are returning the error, then let's avoid logging it here, and let it be logged at a higher layer.

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -40,12 +40,43 @@ func migrateUser(worker *workerConfig, user *model.User) error {
 		// Ignore already existing users
```
agnivade, line 40: Should this comment be changed now?

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -40,12 +40,43 @@ func migrateUser(worker *workerConfig, user *model.User) error {
...
+					err = fmt.Errorf("expected 1 user, got %d", len(users))
+					mlog.Error("got more users than expected in keycloak", mlog.String("username", user.Username), mlog.Int("users_count", len(users)))
+					worker.errorsChan <- err
+					return err
+				}
+
+				kcUserID = *users[0].ID
+			} else {
+				return err
+			}
```
agnivade, line 73: We can invert the condition and return early and outdent the larger block. More idiomatic that way.

deployment/terraform/create.go
```diff
@@ -1118,6 +1118,7 @@ func (t *Terraform) init() error {
 	assets.RestoreAssets(t.config.TerraformStateDir, "keycloak.service")
 	assets.RestoreAssets(t.config.TerraformStateDir, "saml-idp.crt")
 	assets.RestoreAssets(t.config.TerraformStateDir, "provisioners")
+	assets.RestoreAssets(t.config.TerraformStateDir, "saml-idp.crt")
```
agnivade, line 1121: Wasn't this already fixed in https://github.com/mattermost/mattermost-load-test-ng/commit/7f6be66fa3abb1ef4fe9636d24ed3b7145d8a4c8?

fmartingr: I may had that in my local from testing, removed 👍

agnivade (approved):

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -37,22 +37,59 @@ func migrateUser(worker *workerConfig, user *model.User) error {
...
 		if apiErr, ok := err.(*gocloak.APIError); ok && apiErr.Code == 409 {
 			mlog.Debug("user already exists in keycloak", mlog.String("username", user.Username))
-			return nil
-		}
 
-		mlog.Error("failed to create user in keycloak", mlog.String("err", err.Error()))
-		worker.errorsChan <- err
-		return err
-	}
+			// Return early if we are not forcing the migration of all the users users
```
agarciamontoro, line 44: Typo here
```suggestion
			// Return early if we are not forcing the migration of all the users
```

cmd/ltkeycloak/from_mattermost.go
```diff
@@ -262,11 +305,13 @@ func RunSyncFromMattermostCommandF(cmd *cobra.Command, _ []string) error {
 			userPassword:     userPassword,
 			deploymentConfig: deploymentConfig,
 			doneChan:         doneChan,
+			forceMigrate:     migrateAllUsers,
```
agarciamontoro, line 308: Nit: should we rename the variable from `migrateAllUsers` to `forceMigrate`?

agarciamontoro (approved):
Looks great! I've only added a couple of nits. Thanks!

PR #804, opened 2024-09-18
MM-60564: Allow a multi-proxy setup

agnivade:
Most of the terraform work was already done
with ProxyInstanceCount. Thanks @felipemartingr!

The core change was in the /etc/hosts file where
we round-robin between the different proxy IPs
depending on the agent.

Rest of the changes are essentially splitting a single
value to iterate from a loop.

https://mattermost.atlassian.net/browse/MM-60564


deployment/terraform/agent.go
```diff
@@ -243,3 +232,27 @@ func (t *Terraform) initLoadtest(extAgent *ssh.ExtAgent, initData bool) error {
...
+
+func (t *Terraform) getAppHostsFile(index int) (string, error) {
+	output, err := t.Output()
+	if err != nil {
+		return "", err
+	}
+
+	// The new entry in /etc/hosts will make SiteURL point to:
+	// - The first instance's IP if there's a single node
+	// - The proxy's IP(s) if there's more than one node
```
agarciamontoro, line 244: Can we specify here what "node" means? And maybe indicate that we  It's starting to get confusing :joy: 
```suggestion
	// - The first instance's IP if there's a single app node
	// - The IP of one of the proxy nodes if there's more than one app node
```

deployment/terraform/agent.go
```diff
@@ -243,3 +232,27 @@ func (t *Terraform) initLoadtest(extAgent *ssh.ExtAgent, initData bool) error {
...
+	proxyHost := ""
+	if output.HasProxy() {
+		// This allows a bare-bones multi-proxy setup
+		// where we simply round-robin between the proxy nodes amongst the agents.
+		// Not a perfect solution, because ideally we would have used
+		// a dedicated DNS service (like Route53), but it works for now.
```
agarciamontoro, line 250: I guess this is something we'll have to investigate once the multi-proxy setup in the app is ready for production

deployment/terraform/info.go
```diff
@@ -54,7 +54,10 @@ func displayInfo(output *Output) {
 	}
 
 	if output.HasProxy() {
-		fmt.Println("Proxy: " + output.Proxy.PublicIP)
+		fmt.Println("Proxy(ies):")
```
agarciamontoro, line 57: Can we add the humblest `if` here to choose between "Proxy" and "Proxies"? Please :pleading_face:  :pray:

deployment/terraform/utils.go
```diff
@@ -113,10 +113,15 @@ func (t *Terraform) makeCmdForResource(resource string) (*exec.Cmd, error) {
...
 	}
 
+	// Match against proxy names
+	for _, inst := range output.Proxies {
+		if resource == inst.Tags.Name {
+			return exec.Command("ssh", fmt.Sprintf("ubuntu@%s", inst.PublicIP)), nil
+		}
+	}
+
 	// Match against the proxy or metrics servers, as well as convenient aliases.
```
agarciamontoro, line 123: We can remove "proxy" from this comment

agarciamontoro (changes requested):
Looks great! Thanks for the work on this, and indeed thanks to @fmartingr for the foundational work! Just a couple of nits below, nothing too important.

deployment/config.go
```diff
@@ -42,7 +42,7 @@ type Config struct {
 	// Logs the command output (stdout & stderr) to home directory.
 	EnableAgentFullLogs bool `default:"true"`
 	// Number of proxy instances.
-	ProxyInstanceCount int `default:"1" validate:"range:[0,1]"`
+	ProxyInstanceCount int `default:"1" validate:"range:[0,2]"`
```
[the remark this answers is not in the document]
agnivade: No reason. Can be 3 as well. Just chose a number.

deployment/terraform/info.go
agnivade: :joy: :joy: 

Of course

deployment/terraform/agent.go
agnivade: What do you mean "in the app"?

agarciamontoro: I mean the work needed to get multi-proxy setups ready for general consumption. Not sure if we'll need code changes in the server. What I mean generally is that we should use whatever our users have configured (be it Route53 or something else), but that's getting too far into the future. It was just a comment with no expected action in this PR :)

streamer45: I am supportive of the idea of unblocking testing through this simple logic, but I'd also encourage us to look into proper DNS load balancing soon enough.

Doing round robin this way wouldn't necessarily result in fair balancing either as it comes with some edge cases depending on the number of agents and proxies deployed.

agnivade: Yep, fully agree. Proper Route53 support should be something on our long term roadmap.

deployment/config.go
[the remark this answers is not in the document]
agnivade: Set it to 5 for now.

agarciamontoro (approved):
Thank you!

deployment/terraform/info.go
```diff
@@ -54,7 +54,14 @@ func displayInfo(output *Output) {
...
 
 	if output.HasProxy() {
-		fmt.Println("Proxy: " + output.Proxy.PublicIP)
+		if len(output.Proxies) > 1 {
+			fmt.Println("Proxies:")
+		} else {
+			fmt.Println("Proxy:")
+		}
+		for _, inst := range output.Proxies {
+			fmt.Println("- " + inst.PublicIP)
```
streamer45, line 63: Any reason for not using the name like we do for other instances?

streamer45 (approved):
Thanks! For my peace of mind, could we create the follow-up ticket to look into DNS?

agnivade:
Ticket: https://mattermost.atlassian.net/browse/MM-60673

deployment/terraform/info.go
agnivade: Thanks, fixed now.

PR #805, opened 2024-09-18
Remove checking DB status while destroying

agnivade:
This has caused more headache than benefit. The problem is that
if you mess up your deployment slightly and lose your cluster,
then you cannot clean up any other resources because it fails
on "DB cluster not found".

Currently, this causes problems while destroying deployments
with db parameter groups. While that's a separate bug, Stu
also ran into this problem when he messed up his setup.

Whereas, if this check didn't exist, it would naturally fail
during terraform destroy. This check just made it fail sooner.

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -48,18 +48,6 @@ func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


streamer45 (approved):
Cool, got this a few times as well and was a little confused.

agarciamontoro (approved):
I had to comment that out in some occasions as well, I agree we can remove it and revisit with other solution.

PR #807, opened 2024-09-20
Monitor intra-cluster traffic

streamer45:
#### Summary

If you recall, in https://community.mattermost.com/core/pl/ow4ttqyksjnmugizjdkdbiz4je, I included a panel showing the intra-cluster traffic consumption for each transport protocol. That was made possible by changing our implementation and adding a transport wrapper. But for a few reasons, the work in https://mattermost.atlassian.net/browse/MM-58564 didn't go as planned.

So today I wanted to explore collecting the same information without making any application changes. Of course, this is possible thanks to [BPF](https://en.wikipedia.org/wiki/Berkeley_Packet_Filter) and the like. So I wrote a small utility called [`netpeek`](https://github.com/streamer45/netpeek) that exposes the needed information through the expected metrics format.

This approach has other advantages, as it lets us track all traffic, including headers and underlying layers (e.g., IP), which could be significant in the case of TCP.

Also, this approach could be easily repurposed in the future to track arbitrary traffic in and out of our instances through a given port.

Let me know what you think.

#### Screenshot

image


deployment/terraform/engine.go
```diff
@@ -61,6 +63,27 @@ func (t *Terraform) runAWSCommand(ctx context.Context, args []string, dst io.Wri
 	return _runCommand(cmd, dst)
 }
 
+func buildGoPackage(ctx context.Context, pkg string, dst io.Writer) (string, error) {
```
agnivade, line 66: Can't we just download the release binary? Compiling code during terraform create doesn't sound ideal.

deployment/terraform/strings.go
```diff
@@ -384,3 +387,18 @@ KC_DATABASE=keycloak`
...
 `
+
+const netpeekServiceFile = `
+[Unit]
+Description=netpeek
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/bin/sh -c '/usr/local/bin/netpeek -iface "$(ip route show to default | awk \'{print $5}\')" -port %d'
```
agarciamontoro, line 398: Why do we need the `/bin/sh` here? Can't we just run the binary directly?

agarciamontoro (commented):
Ooooh, I like this approach! I agree with Agniva: can we download the binary directly? Apart from that, I'm onboard with the idea :tada:

deployment/terraform/strings.go
streamer45: It's needed to make the subcommand that finds out the default network interface work properly.

agarciamontoro: Ah, of course! Thanks for clarifying :nerd_face:

streamer45:
Thanks. I simplified to use a released binary. I also added a flag (default off) to control it since I don't think we'd want it on in all cases, as it can introduce some overhead (it processes every single UDP/TCP packet after all). I can see it being useful during ceiling tests but not as much in day-to-day comparisons.

deployment/terraform/create.go
```diff
@@ -488,6 +491,19 @@ func (t *Terraform) setupAppServer(extAgent *ssh.ExtAgent, ip, siteURL, serviceF
 		}
 	}
 
+	if t.config.EnableNetPeekMetrics {
+		// Stop netpeek service in case it's already deployed
+		cmd = "sudo service netpeek stop"
+		if out, err := sshc.RunCommand(cmd); err != nil {
+			return fmt.Errorf("error running ssh command %q, output: %q: %w", cmd, string(out), err)
+		}
```
agnivade, line 499: Normally a restart is enough right?

agnivade (approved):

docs/config/deployer.md
```diff
@@ -485,6 +485,15 @@ This config is used for tests that require an existing database dump that contai
...
 The path to a file containing a list of credentials for the controllers to use. If present, it is used to automatically upload it to the agents and override the agent's config's own [`UsersFilePath`](config.md/#UsersFilePath).
 
+## EnableNetPeekMetrics
+
+*bool*
+
+If true, enables the collection of fine grained networking metrics through the [netpeek](https://github.com/streamer45/netpeek) utility.
+
+> [!WARNING]
+> These metrics can introduce a not negligible computational overhead on high traffic deployments.
```
agarciamontoro, line 495: Is this something we should measure at some point? Specially if we want to run it in ceiling tests

agarciamontoro (approved):
Great, thank you!

docs/config/deployer.md
streamer45: @agarciamontoro I [measured a bit](https://snapshots.raintank.io/dashboard/snapshot/0TA9RUvi6wyzUaug9H8BdcH6N4o6c6mm?orgId=0&from=1726796371225&to=1726797271225) while testing this PR. The binary also exposes go-process metrics so it's easy to track CPU time if needed.

deployment/terraform/create.go
streamer45: This was a leftover from the previous approach that replaced the binary so required a stop. I'll simplify, thanks.

PR #808, opened 2024-09-24
MM-60590: Add custom tags

agarciamontoro:
#### Summary
In order to better manage AWS resources, we need the ability to add custom tags. This will come in particularly handy for automated load-tests, where we want to tag each test in a unique way (using tags like `PR` or `SHA`), apart from tagging all automated load-tests in a consistent way (using a tag like `Origin` with a value like `automated-load-tests`).
This PR unblocks all this, adding a new `CustomTags` setting to the deployer that can receive an arbitrary amount of key-value pairs that are used to tag *all* resources deployed.

For example, using 
```json
"CustomTags": {
  "Origin": "agm-testing",
  "PR": "123",
  "SHA": "abcdefg"
}
```
we get all resources tagged as specified, apart from the already hard-coded `ClusterName` and `Name`. See how the first agent is tagged:
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60590


config/deployer.sample.toml
```diff
@@ -138,3 +138,5 @@ InstanceEngine = 'aurora-postgresql'
 InstanceType = 'db.r6g.large'
 Password = 'mostest80098bigpass_'
 UserName = 'mmuser'
+
+[CustomTags]
```
streamer45, line 142: Unrelated but I have to ask; is there anyone using this format? Why was it introduced?

deployment/terraform/assets/cluster.tf
```diff
@@ -11,9 +11,12 @@ provider "aws" {
   region  = var.aws_region
   profile = var.aws_profile
   default_tags {
-    tags = {
-      ClusterName = var.cluster_name
-    }
+    tags = merge(
```
streamer45, line 14: nice

streamer45 (approved):
Makes sense 👍

config/deployer.sample.toml
agnivade: Felipe added it. It's just for convenience I think more than anything.

agarciamontoro: Yup, this was a proposal from @fmartingr, which we gladly accepted as a potential UX improvement. I am not using it at the moment, but I want to test it and feel the difference :) Not sure if Felipe is using it currently.

deployment/config.go
```diff
@@ -115,6 +115,21 @@ type Config struct {
...
 	// EnableNetPeekMetrics enables fine grained networking metrics collection through netpeek utility.
 	EnableNetPeekMetrics bool `default:"false"`
+	// CustomTags is an optional list of key-value pairs, which will be used as default
+	// tags for all resources deployed
+	CustomTags TerraformMap
+}
+
+// TerraformMap is a map of string -> string that serializes to the format expected by
+// the Terraform AWS provider when formatted as a string
+type TerraformMap map[string]string
```
agnivade, line 125: 0/5, but this map has nothing to do with Terraform right? I'd rather just say `StringMap`.

agnivade (approved):

deployment/config.go
agarciamontoro: Its String method is Terraform-specific, that's why I called it that

PR #809, opened 2024-09-25
[MM-60675] Allow deploying instances into a given Availability Zone

streamer45:
#### Summary

[Cost explorer](https://us-east-1.console.aws.amazon.com/costmanagement/home?region=us-east-1#/cost-explorer?chartStyle=STACK&costAggregate=unBlendedCost&endDate=2024-09-25&excludeForecasting=true&filter=%5B%5D&futureRelativeRange=CUSTOM&granularity=Monthly&groupBy=%5B%22Operation%22%5D&historicalRelativeRange=CUSTOM&isDefault=true&reportName=New%20cost%20and%20usage%20report&showOnlyUncategorized=false&showOnlyUntagged=false&startDate=2024-09-11&usageAggregate=undefined&useNormalizedUnits=false) shows that a significant part of our monthly cost (~40% in Sept so far) is caused by intraregional traffic, meaning traffic across different AZs in the same region. 

image

According to [AWS pricing](https://aws.amazon.com/ec2/pricing/on-demand/#Data_Transfer_within_the_same_AWS_Region), traffic in and out of the same AZ should be generally free.

PR adds a new `AWSAvailabilityZone` setting to deployer's config that controls into which AZ instances are created.

A few additional notes/thoughts/caveats:

1. I think AWS (or Terraform) is generally smart about this because I couldn't reproduce a case where resources for the same deployment are not deployed on the same AZ. I suspect that as the number of deployed instances grows, that may happen eventually though.
2. ~~For databases, there's a multi-az requirement in place when deploying more than a single writer, which means we cannot force this for all instances~~ (this may not be the case; see discussion below). Still, if we can get one of them to match, it's likely better than keeping them selected completely randomly.
3. Not touching OpenSearch at all here. I'm not sure it's worth the complexity, and again, it would likely be multi-az by default.
4. Currently not enforcing the setting to be set. To be effective and see a cost reduction, we may need to do it or ask people to update their configs.

Most importantly, we should keep monitoring to see if this is giving any real cost reduction because AWS pricing is weird and complex, so it's hard to tell without experimenting a bit.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60675


agnivade:
Perhaps I am not aware of AWS specifics, but is AZ different than a region? We always deploy everything in the same region right? 

> I think AWS (or Terraform) is generally smart about this because I couldn't reproduce a case where resources for the same deployment are not deployed on the same AZ. I suspect that as the number of deployed instances grows, that may happen eventually though.

Yeah I think we should investigate further as to exactly from which service the inter-region cost is coming from. Is that possible? I doubt that EC2 instances would be deployed cross-region when we specifically mention the AWS region.

I can check on Opensearch next time I deploy it, but I don't believe it's multi-AZ by default.

agarciamontoro:
@agnivade: From https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html:

> Availability Zones are multiple, isolated locations within each Region.

agnivade:
Correct, and from what I can see, data transfer within the same region, _across AZs_ are charged lower. i.e. there is no difference between transfer within the same AZ, or different AZs as long as they remain in the same region.

https://aws.amazon.com/ec2/pricing/on-demand/#Data_Transfer

> Data transferred "in" to and "out" from Amazon EC2, Amazon RDS, Amazon Redshift, Amazon DynamoDB Accelerator (DAX), and Amazon ElastiCache instances, Elastic Network Interfaces or VPC Peering connections across Availability Zones in the same AWS Region is charged at $0.01/GB in each direction.

Are we somehow deploying across regions or something? I even checked my S3 bucket and it's in the same region.

agarciamontoro:
I think the important part is:
> Data transferred between Amazon EC2, Amazon RDS, Amazon Redshift, Amazon ElastiCache instances, and Elastic Network Interfaces in the same Availability Zone **is free**.

agarciamontoro (commented):
The code looks good to me, but I'm concerned about the traffic from/to the database. Is it impossible to force the same zone there?

streamer45:
@agarciamontoro @agnivade The charge we see is definitely from the same region (as expected since we don't deploy anything out of `us-east-1` anyway), just different AZs. It could very well be from/to DB but so far I wasn't able to extrapolate more fine grained data. We may need to look in CloudWatch to understand more.

Also, the charge is under EC2 costs, but I understood that Aurora is deployed in EC2 underneath so that's still a possibility. I don't believe it's possible to have read replicas on the same AZs. The cluster itself [requires 3 AZs as input](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/rds_cluster#availability_zones), and if less are given, three are selected randomly.

I think the best we can do for now is at least have one in the same AZ, which could still be a significant saving vs none. For a practical example, looking at a running load-test, I see the app servers are on a different AZ than the matching RDS instances:

image

image

streamer45:
So, AWS docs are super confusing. It seems that Aurora specifically (not standard RDS) may actually support single AZ deployments (or I got extremely lucky!).

image

agnivade:
> So, AWS docs are super confusing.

Wouldn't be the first time :)

docs/config/deployer.md
```diff
@@ -12,14 +12,23 @@ AWS profile to use for the deployment. Also used for all AWS CLI commands run lo
...
 
+## AWSAvailabilityZone
+
+AWS Availability Zone in which to deploy instances. See the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html) for more information.
+
+> [!TIP]
+> Deploying instances in the same Availability Zone can greatly reduce traffic costs. Refer to [data transfer pricing](https://aws.amazon.com/ec2/pricing/on-demand/#Data_Transfer_within_the_same_AWS_Region) for more details.
+
+> [!NOTE]
+> RDS clusters generally have a multi-az requirement which means it's not possible to force all instances (e.g. writer, readers) to be on the same availability zone.
```
agnivade, line 23: Should we remove this given the new insight?

agnivade (approved):
Lots of new learnings for me! Thanks for digging into the docs.

docs/config/deployer.md
streamer45: Yeah, I think we can remove it for now. I believe pure RDS clusters still impose this requirement, but we use Aurora so it shouldn't be an issue worth raising.

streamer45:
@agarciamontoro Let me know if you have concerns. I'd like to get these in soon so we can start monitoring. Will need to make an announcement afterwards so that people update their branch to deploy.

@agnivade If you have to re-create your ceiling test deployment, consider using this branch to see if we can limit expense :)

agarciamontoro (approved):
Thanks, @streamer45!

PR #810, opened 2024-09-26
Use same VPC for all resources

fmartingr:
#### Summary
- [x] Specifies de VPC for all resources based on the `DeployerConfig.ClusterVpcID` variable
- [ ] Normalize `DeployerConfig.ElasticSearchSettings.VpcID` into `DeployerConfig.ClusterVpcID` ?


fmartingr:
> Possibly related: #809. Can a VPC spread across AZs?

IIRC the subnets are tied to AZs, VPCs are spread out, but probably @streamer45 knows more. I will investigate a bit later.

fmartingr:
> > Possibly related: #809. Can a VPC spread across AZs?
> 
> IIRC the subnets are tied to AZs, VPCs are spread out, but probably @streamer45 knows more. I will investigate a bit later.

Yeah it seems that how that works. I'd say this PR is complementary to @streamer45 's

agnivade:
So do we need both or one is fine?

fmartingr:
> So do we need both or one is fine?

Just this one is good for our use case.

deployment/terraform/assets/cluster.tf
```diff
@@ -40,6 +40,8 @@ locals {
 data "aws_subnets" "selected" {
   filter {
     name   = "vpc-id"
+    // TODO: Any specific reason we use a specific VPC for ES only here?
+    // Should we use the `cluster_vpc_id` variable instead and remove `es_vpc`?
```
streamer45, line 44: I remember VPC being a key detail when deploying ES, but that has likely preceded our ability to set one so we may be able to use the same variable. That said, the best thing would be to deploy the changes and see that everything works.

streamer45 (commented):
I see we are only changing the security rules. I'd expect us to set `subnet_id` on the resources we create (e.g. EC2) in order to deploy them in the expected VPC. See https://github.com/mattermost/mattermost-load-test-ng/pull/379 for reference.

sadohert:
> I see we are only changing the security rules. I'd expect us to set `subnet_id` on the resources we create (e.g. EC2) in order to deploy them in the expected VPC. See #379 for reference.

Subnet ID was something that came up yesterday, and I did not include it in my notes (call was running way late and I was way late)

fmartingr:
> > I see we are only changing the security rules. I'd expect us to set `subnet_id` on the resources we create (e.g. EC2) in order to deploy them in the expected VPC. See #379 for reference.
> 
> Subnet ID was something that came up yesterday, and I did not include it in my notes (call was running way late and I was way late)

Testing creating a new env over at #813 this came up so I'm trying to put some changes to make it work without making too many backwards incompatible changes.

sadohert:
> So do we need both or one is fine?

I wonder if 1 is fine, given the operator saw this error:

```
OpenSearch Domain: operation error OpenSearch: CreateDomain, https response error StatusCode: 400, RequestID: fcfe4c9b-36b8-41b0-be1b-ffbd75a208ab, Val     idationException: Subnets and security groups must be in the same VPC.
```

Which I think happened because the es subnet and `ClusterSubnet` were set.

fmartingr:
> > So do we need both or one is fine?
> 
> I wonder if 1 is fine, given the operator saw this error:
> 
> ```
> OpenSearch Domain: operation error OpenSearch: CreateDomain, https response error StatusCode: 400, RequestID: fcfe4c9b-36b8-41b0-be1b-ffbd75a208ab, Val     idationException: Subnets and security groups must be in the same VPC.
> ```
> 
> Which I think happened because the es subnet and `ClusterSubnet` were set.

Yeah, the selected subnet should be in the selected Vpc.

fmartingr:
Replaced by #814

PR #813, opened 2024-09-26
Setup VPC parameter for all resources and allow custom database engine version

fmartingr:
⚠️ This pull request is **not** pointing to `master`.

##### Customizable database engine

- Renames `db_engine_version` to `db_default_engine_version`
- Adds new `db_engine_version` (customizable, string)
- Adds new parameter to `DeployerConfig.TerraformDBSettings.EngineVersion`
  - ℹ️ Remember to match the selected version with the appropriate instance class. ([docs](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/Concepts.DBInstanceClass.SupportAurora.html))
- Logic now uses the above variable is set, defaults otherwise

##### Use same VPC for all resources

- Specifies de VPC for all resources based on the `DeployerConfig.ClusterVpcID` variable
- Elasticsearch still has it's own VPC parameter.
- Added a subnet to all resources. Uses the one defined in `DeploymentConfig.ClusterSubnetID` or a random one from the selected `DeploymentConfig.ClusterVpcID`.

##### Other
- Wrapped command output in a goroutine as well, when the terraform command fails early (validation) the error output wasn't being shown. (ec421b5ad1457c37795c59684d67137b7ee686aa)


fmartingr:
Currently testing this on a new environment.

wiggin77 (approved):
LGTM

streamer45:
My comments at https://github.com/mattermost/mattermost-load-test-ng/pull/810 would still apply.

streamer45:
Nothing blocking from my side in terms of merging to feature branch if the changes work for the current use case.

fmartingr:
Launched a successful test with this PR (without ES, due to failures without a snapshot). Merging this into the feature branch.

PR #816, opened 2024-09-30
Fix malformed Prometheus configuration YAML file

agarciamontoro:
#### Summary
It seems that https://github.com/mattermost/mattermost-load-test-ng/pull/807 stole the `static_configs:` line from the cloudwatch section and gave it to the `netpeek` one :smiling_imp:

#### Ticket Link
--

### DIFF
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -55,6 +55,7 @@ scrape_configs:
     static_configs:
         - targets: [%s]
   - job_name: cloudwatch
+    static_configs:
         - targets: [%s]
   - job_name: netpeek
     static_configs:


agnivade (commented):
I know it's slightly out of scope, but can I ask this string to be templatized like `nginxSiteConfigTmpl`? The `%s`s are starting to get out of hand and becoming very brittle as is apparent with this break.

Hopefully, it's not too much of work.

streamer45 (approved):
I had to merge that a few times to get it right, and, of course, something got messed up. We should probably have some basic parsing tests on these config files. I can look into it, but I don't want to delay the fix.

agarciamontoro:
> I know it's slightly out of scope, but can I ask this string to be templatized like nginxSiteConfigTmpl? The %ss are starting to get out of hand and becoming very brittle as is apparent with this break.

> We should probably have some basic parsing tests on these config files.

Both comment seems sensible to me. I'd say we merge this now and I can tackle those improvements later.

agarciamontoro:
Ticket created: https://mattermost.atlassian.net/browse/MM-60726

agnivade (approved):

PR #818, opened 2024-10-04
MM-60651: Add Loki to collect app, agents, and proxy logs

agarciamontoro:
#### Summary

This PR adds Loki to the metrics instance in order to centralize log analysis. The actual collection of the logs from the app, agent and proxy nodes happen using the [OpenTelemetry collector](https://opentelemetry.io/docs/collector/), specifically the [filelog receiver](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/filelogreceiver/README.md), that is configured to retrieve the corresponding logs in each instance.

This PR also adds a new section to our default dashboard with panels listing all log lines and summarizing their count in each instance type by level, as shown in the screenshot (the proxy errors were forced by setting worker_connections to 500 :smiling_imp:).

Please, review by commit, I've tried to isolate the changes and describe them in each commit message :)

image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60651


deployment/terraform/agent.go
```diff
@@ -139,11 +139,28 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent) error {
 			buf := bytes.NewBufferString("")
 			tpl.Execute(buf, tplVars)
 
+			otelcolConfig, err := fillConfigTemplate(otelcolConfigTmpl, map[string]string{
+				"IncludeFiles": strings.Join([]string{
+					"/home/ubuntu/mattermost-load-test-ng/ltagent.log",
+					"/home/ubuntu/mattermost-load-test-ng/ltcoordinator.log",
+				}, ", "),
```
streamer45, line 146: Do these logs (agent and coordinator) get interleaved or is there an easy way to filter one or the other? In other words, I'd expect a separate service name for coordinator's logs.

deployment/terraform/assets/cluster.tf
```diff
@@ -572,6 +572,16 @@ resource "aws_security_group_rule" "metrics-pyroscope" {
   security_group_id = aws_security_group.metrics[0].id
 }
 
+resource "aws_security_group_rule" "metrics-loki" {
+  count             = var.app_instance_count > 0 ? 1 : 0
```
streamer45, line 576: Conceptually, I'd expect Loki to be needed as long as there's a metrics instance.

deployment/terraform/assets/provisioners/metrics.sh
```diff
@@ -41,6 +41,12 @@ do
       sudo mkdir /opt/yace && \
       wget https://github.com/nerdswords/yet-another-cloudwatch-exporter/releases/download/v0.61.2/yet-another-cloudwatch-exporter_0.61.2_Linux_x86_64.tar.gz && \
       sudo tar -zxf yet-another-cloudwatch-exporter_0.61.2_Linux_x86_64.tar.gz -C /opt/yace && \
+      # Install Loki
+      mkdir -p /etc/apt/keyrings/ && \
+      sudo bash -c 'wget -q -O - https://apt.grafana.com/gpg.key | gpg --dearmor > /etc/apt/keyrings/grafana.gpg' && \
+      sudo bash -c 'echo "deb [signed-by=/etc/apt/keyrings/grafana.gpg] https://apt.grafana.com stable main" | tee /etc/apt/sources.list.d/grafana.list' && \
+      sudo apt-get -y update && \
+      sudo apt-get install -y loki && \
```
streamer45, line 49: Wondering if it would be best to follow what we did for other Grafana packages and install from the `.deb` so we can pin the version and avoid surprises.

deployment/terraform/assets/default_dashboard_tmpl.json
```diff
@@ -2809,10 +3850,9 @@
     "list": [
       {
         "current": {
-          "text": [
-          ],
-          "value": [
-          ]
+          "selected": false,
+          "text": "agent-0:4000",
+          "value": "agent-0:4000"
```
streamer45, line 3855: I'd probably keep these empty, as before.

streamer45 (approved):
Great work, can't wait to try it :)

agnivade:
Sorry for the delay here. Will review tomorrow.

deployment/terraform/agent.go
agarciamontoro: Good point! You can filter them by log name (the panels use that filtering). Do you think that's enough or would you prefer to have a different service name altogether?

deployment/terraform/assets/cluster.tf
agarciamontoro: That makes sense, let me check that instead

deployment/terraform/assets/provisioners/metrics.sh
agarciamontoro: Good point, will take a look

deployment/terraform/assets/provisioners/agent.sh
```diff
@@ -14,6 +14,12 @@ do
         sudo apt-get -y update && \
         sudo apt-get install -y prometheus-node-exporter && \
         sudo apt-get install -y numactl linux-tools-aws linux-tools-aws-lts-22.04 && \
+        # Install OpenTelemetry collector, using ubuntu user to avoid permission issues
+        wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.110.0/otelcol-contrib_0.110.0_linux_amd64.deb && \
+        sudo dpkg -i otelcol-contrib_0.110.0_linux_amd64.deb && \
+        sudo sed -i 's/User=.*/User=ubuntu/g' /lib/systemd/system/otelcol-contrib.service && \
+        sudo sed -i 's/Group=.*/Group=ubuntu/g' /lib/systemd/system/otelcol-contrib.service && \
```
agnivade, line 21: This looks brittle. Would it be worth having our own service file and just copy-pasting it over to the instance? 0/5

deployment/terraform/agent.go
agarciamontoro: What the receiver does is `tail`ing the file specified, so I don't think rolled logs is a problem. I'll see how easy it is to add another service to that, it makes sense logically

deployment/terraform/assets/provisioners/agent.sh
agarciamontoro: I'd say that having our own service file copied brings other forms of brittleness to the table, so I'd keep this as is.

agnivade (approved):
Thanks!

PR #819, opened 2024-10-04
Conditionally create redis

coltoneshaw:
Conditionally creating redis security groups, only if it's enabled and you have app nodes.

### DIFF
--- a/deployment/terraform/assets/cluster.tf
+++ b/deployment/terraform/assets/cluster.tf
@@ -593,7 +593,7 @@ resource "aws_security_group" "redis" {
     security_groups = [aws_security_group.app[0].id, aws_security_group.metrics[0].id]
   }
 
-  count = 1
+  count = var.redis_enabled ? 1 : 0
 }
 
 resource "aws_security_group" "elastic" {


deployment/terraform/assets/cluster.tf
```diff
@@ -593,7 +593,7 @@ resource "aws_security_group" "redis" {
     security_groups = [aws_security_group.app[0].id, aws_security_group.metrics[0].id]
   }
 
-  count = 1
+  count = var.app_instance_count > 0 && var.redis_enabled ? 1 : 0
```
agarciamontoro, line 596: Do we need to check that there are app nodes deployed? Isn't it enough with checking that Redis is enabled?

agnivade: Agree. We should be able to use Redis even with single app node.

coltoneshaw: I agree too, but both security groups in the array are tied to if you have created an app node.

agarciamontoro: Aaaah, you're right! This is basically for the app to be able to communicate with Redis, so I guess it makes sense

agnivade: I am blind. I mistook `> 0` to be `> 1`. We generally don't test with no app nodes. Thanks for looking into these edge casees.

agnivade (approved):
Thanks!

deployment/terraform/assets/cluster.tf
```diff
@@ -593,7 +593,7 @@ resource "aws_security_group" "redis" {
     security_groups = [aws_security_group.app[0].id, aws_security_group.metrics[0].id]
```
agarciamontoro, line 593: It seems `count` is resolved after the check on the number of elements in tuples is done, so this PR still errors to me with "Invalid Index" error, stating that `app_security_group.app is empty tuple`. I found that we can use `count.index` to bypass this:

```suggestion
    security_groups = [aws_security_group.app[count.index].id, aws_security_group.metrics[count.index].id]
```

agarciamontoro (commented):
I've tested this patch in a deployment I'm running now with no app nodes and I still get an error. Is it working for you, @coltoneshaw? More details in the comment below.

coltoneshaw:
@agarciamontoro - Fixed, not sure why that didn't reflect here when i tested.

agnivade:
@coltoneshaw - would you be able to resolve the conflicts?

agarciamontoro (approved):
Thank you, Colton! Can you resolve the conflicts? It seems it's just a matter of running `make assets` again

> Fixed, not sure why that didn't reflect here when i tested.

Maybe you already had a Terraform state where the count was already resolved. That happened to me the first thing I tried.

PR #820, opened 2024-10-07
MM-52099: Be more gentle with comparison failures

agarciamontoro:
#### Summary
Before this PR, when running a comparison, the tool would cancel all tests if any single one of them would fail. This does not need to be like this, since all runs in a comparison are independent of each other: they're completely different deployments. If one of them fails, we can still try to finish the other ones, so that we only need to retry one of them in a second try.

This PR does exactly that: it makes the independent deployments actually independent by removing the cancel channel, which was used to broadcast the cancel to all deployments in the case of a single failure. Removing this makes the code slightly less complex as well.

And while I was here:
- I've removed an ancient TODO that was fixed four years ago (:shrug:)
- I've added the deployment ID to error log lines to know which test specifically failed

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-52099

### DIFF
--- a/cmd/ltctl/comparison.go
+++ b/cmd/ltctl/comparison.go
@@ -67,6 +67,13 @@ func writeResults(results []comparison.Result, writer io.Writer) error {
 	var content string
 
 	for i, res := range results {
+		if len(res.LoadTests) < 2 || res.LoadTests[0].Failed || res.LoadTests[1].Failed {
+			content += "==================================================\n"
+			content += fmt.Sprintf("Deployment %d: No results generated\n", i)
+			content += "==================================================\n\n"
+			continue
+		}
+
 		content += "=================================================="
 		content += "Comparison result:"
 		content += fmt.Sprintf("Report: %s\n", getReportFilename(i, res))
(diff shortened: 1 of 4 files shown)


cmd/ltctl/comparison.go
```diff
@@ -67,6 +67,13 @@ func writeResults(results []comparison.Result, writer io.Writer) error {
 	var content string
 
 	for i, res := range results {
+		if len(res.LoadTests) < 2 || res.LoadTests[0].Failed || res.LoadTests[1].Failed {
```
agnivade, line 70: So if one failed, and other passed, we would not be displaying the one that passed right? Is that intentional? I was thinking we can move this check inside the for loop and just ignore the one that failed.

agarciamontoro: The thing is that if a single test has failed, there are undefined things here, like the `DashboardURL` (which contains the queries from the base and new builds, one of them with an offset to compare both runs in the same graph). Those are generated in `Comparison.getResults`, which contains the same check to skip the generation of those results. We could technically avoid writing the results of these common things and print whatever individual results we have, but what I decided here was to treat a comparison like an indivisible unit: given that we want to compare two things, in my opinion there should be no results if both tests didn't finish successfully.

agnivade (approved):

PR #822, opened 2024-10-08
MM-56823: Validate license only when deploying app nodes

agarciamontoro:
#### Summary
If we are not deploying app nodes (to load-test an already existing Mattermost installation), we don't need to validate the license (first commit). Actually, we don't even need to specify a file in the config at all (second commit).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-56823

### DIFF
--- a/defaults/validate.go
+++ b/defaults/validate.go
@@ -153,14 +153,6 @@ func validate(validation, fieldName string, p, v reflect.Value) error {
 		if s == "" || !isAlphanumeric(s) {
 			return fmt.Errorf("%s is not alphanumeric", fieldName)
 		}
-	case "emptyorfile":
-		s := v.String()
-		if s == "" {
-			return nil
-		}
-		if _, err := os.Stat(s); os.IsNotExist(err) {
-			return fmt.Errorf("%s: %w", fieldName, err)
-		}
 	case "file":
 		s := v.String()
 		if _, err := os.Stat(s); os.IsNotExist(err) {
--- a/defaults/validate.go
+++ b/defaults/validate.go
@@ -11,8 +11,6 @@ import (
 	"strconv"
 	"strings"
 	"unicode"
-
-	"github.com/wiggin77/merror"
 )
 
 var (
@@ -99,7 +97,7 @@ func validate(validation, fieldName string, p, v reflect.Value) error {
 	// Multiple validations separated by a colon (as in "empty;file") behave like an OR clause
 	if strings.Contains(validation, ";") {
 		individualValidations := strings.Split(validation, ";")
-		merr := merror.New()
+		errs := []error{}
 		for _, validation := range individualValidations {
 			err := validate(validation, fieldName, p, v)
 			// Return as soon as one is valid
@@ -107,9 +105,9 @@ func validate(validation, fieldName string, p, v reflect.Value) error {
 				return nil
 			}
 
-			merr.Append(err)
+			errs = append(errs, err)
 		}
-		return fmt.Errorf("%s failed all validations %q: %w", fieldName, individualValidations, merr)
+		return fmt.Errorf("%s failed all validations %q: %w", fieldName, individualValidations, errors.Join(errs...))
 	}
 
 	switch validation {
--- a/defaults/validate.go
+++ b/defaults/validate.go
@@ -94,9 +94,9 @@ func Validate(value interface{}) error {
 }
 
 func validate(validation, fieldName string, p, v reflect.Value) error {
-	// Multiple validations separated by a colon (as in "empty;file") behave like an OR clause
-	if strings.Contains(validation, ";") {
-		individualValidations := strings.Split(validation, ";")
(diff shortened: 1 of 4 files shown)


defaults/validate.go
```diff
@@ -123,6 +123,14 @@ func validate(validation, fieldName string, p, v reflect.Value) error {
 		if s == "" || !isAlphanumeric(s) {
 			return fmt.Errorf("%s is not alphanumeric", fieldName)
 		}
+	case "emptyorfile":
```
streamer45, line 126: Given the license is the only use of the `file` tag I wonder if we could more simply make it accept empty values. It would be cool to allow multiple subtags, so one could define something like `validate:"notempty,file"`.

defaults/validate_test.go
```diff
@@ -371,6 +373,50 @@ func TestValidate(t *testing.T) {
...
+	})
+
+	t.Run("not a path for emptyorfile", func(t *testing.T) {
+		var cfg serverConfiguration
+		Set(&cfg)
+
+		cfg.LicenseFile = "not a file"
+
+		err := Validate(&cfg)
+		require.Error(t, err)
```
streamer45, line 418: nit: I usually find value in using [`EqualError`](https://pkg.go.dev/github.com/stretchr/testify/require#EqualError) rather than simply checking if it errored. It saved me from a few bugs in the past :)

streamer45 (approved):
Nice, and great test coverage!

agnivade (approved):
I really like the `validate:"notempty,file"` idea, but non-blocking. Great improvements.

agarciamontoro:
I added an OR clause: simply concatenate multiple validations separated by a colon. If one of those validations pass, it is considered valid. What we need for the license in this case is `"empty;file"`, since we want it either to be empty, or to be a valid, existing file.

agarciamontoro:
Re-requesting review just to double-check this is what was on your minds. I do like this solution myself :)

defaults/validate.go
```diff
@@ -94,6 +96,22 @@ func Validate(value interface{}) error {
...
+		individualValidations := strings.Split(validation, ";")
+		merr := merror.New()
+		for _, validation := range individualValidations {
+			err := validate(validation, fieldName, p, v)
+			// Return as soon as one is valid
+			if err == nil {
+				return nil
+			}
+
+			merr.Append(err)
```
agnivade, line 110: You can just use the standard library: https://pkg.go.dev/errors#Join

agnivade (approved):
Didn't mean my previous comment to be blocking.

defaults/validate.go
agarciamontoro: Oh, didn't know about that one, thanks!

defaults/validate.go
```diff
@@ -94,6 +94,22 @@ func Validate(value interface{}) error {
 }
 
 func validate(validation, fieldName string, p, v reflect.Value) error {
+	// Multiple validations separated by a colon (as in "empty;file") behave like an OR clause
+	if strings.Contains(validation, ";") {
+		individualValidations := strings.Split(validation, ";")
```
streamer45, line 99: I am thinking that calling `Split` directly and checking if we get more than a single result would be enough?

defaults/validate_test.go
```diff
@@ -18,6 +19,8 @@ func TestValidate(t *testing.T) {
 		MaxUsers      int    `default:"1000" validate:"range:(0,]"`
 		LogLevel      string `default:"ERROR" validate:"oneof:{TRACE, INFO, WARN, ERROR}"`
 		S3URI         string `default:"" validate:"s3uri"`
+		LicenseFile   string `default:"" validate:"empty;file"`
```
streamer45, line 22: I recommend `|` instead of `;`. Then we could add `&` for improved boolean logic :)

streamer45 (approved):
Great work!

defaults/validate_test.go
agarciamontoro: Oooh, I like it! Let's do that :)

defaults/validate.go
agarciamontoro: Mmmm we could, but I feel like
```
if individualValidations := strings.Split(validation, "|"); len(individualValidations) > 1 {
```
is a bit less readable than what we have now, and
```
individualValidations := strings.Split(validation, "|")
if len(individualValidations) > 1 {
```
 is basically the same, but it makes my brain slightly uncomfortable, it's like it's the other way around :joy: 
 
 If you feel strong about this, or if you were thinking about something else, please let me know!

streamer45: Nope, I just prefer to keep the scope level down when feasible. It's a preference really so not at all blocking.

defaults/validate_test.go
streamer45: that was half a joke you know :p

streamer45 (approved):
Merge this please :D

defaults/validate_test.go
agarciamontoro: You have to be careful with your jokes!

PR #823, opened 2024-10-09
MM-60990: Allow attaching a custom IAM Profile to app servers

fmartingr:
#### Summary
Adds a new `AppAttachIAMProfile` deployer configuration setting to be able to attach a custom IAM profile to the application servers.

Useful in scenarios where you want to give application servers access to another AWS resource but you can't do so with the deployer.json settings.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60990


agnivade:
Just checking, have you verified this works with our deployment? Because we always set the access key/secret key to use an S3 bucket. So in this case, they would be empty and the IAM profile would take care of that?

fmartingr:
> Just checking, have you verified this works with our deployment? Because we always set the access key/secret key to use an S3 bucket. So in this case, they would be empty and the IAM profile would take care of that?

I have tested that the feature works (I can deploy an environment and an IAM profile its attached to the EC2 server) but I have not checked that the actual IAM profile works as it should, since thats on AWS side and I would expect that to work as it should. Can do that if you feel more comfortable.

agnivade:
Ah no I don't think that's necessary. So when IAM profile is being used, we are setting empty strings to access key/secret key and the MM client library just works?

fmartingr:
> Ah no I don't think that's necessary. So when IAM profile is being used, we are setting empty strings to access key/secret key and the MM client library just works?

In this case this is only attached to the app servers and I've seen it used to allow access to an already existing bucket without having to populate the key/secret in the deployer.json. The client interacts with the application servers and the application server can connect to the bucket.

agnivade (approved):
Thanks

agarciamontoro (approved):
Looks good to me! And thank you for adding it to the docs and to the sample files :)

PR #825, opened 2024-10-10
Use app_instance_count as proxy for metrics node

agarciamontoro:
#### Summary
We cannot use the length of aws_security_group.metrics to resolve the count, since that's not known until apply. We use the variable app_instance_count as a proxy for it, like the other security group rules.

I didn't catch this in #818 because I didn't try to create a new deployment from scratch after changing it, only to re-create it. At that point, we did know that the metrics security group existed. Let's use this for now, which is proven to work well.


agnivade (approved):
Thanks. These terraform rules are hard to review just by looking at it.

agarciamontoro:
> These terraform rules are hard to review just by looking at it.

I need another fixathon to focus on adding testing to our Terraform code.

agarciamontoro:
Wow, I "assigned" you guys instead of requested your review :

streamer45:
Cool. Apologies for the extra work :sweat:

streamer45 (approved):

PR #832, opened 2024-10-15
MM-60036: License validation

agarciamontoro:
#### Summary
A common error when providing a license is to use a different service environment (dev, test or production) to the one the license was generated for. This error is not surfaced by the call to `validator.ValidateLicense`, so we explicitly test a different environment to the one currently in use to check whether the license would be valid there. If that's the case, then we return a much more helpful error to the user.

Note that the license in the test is not valid for production, only for test (and for 1 user just in case).

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60036


deployment/terraform/create.go
```diff
@@ -109,7 +109,12 @@ func (t *Terraform) Create(initData bool) error {
 	// Validate the license only if we deploy app nodes;
 	// otherwise we don't need a license at all
 	if t.config.AppInstanceCount > 0 {
-		if err := validateLicense(t.config.MattermostLicenseFile); err != nil {
+		licenseData, err := os.ReadFile(t.config.MattermostLicenseFile)
+		if err != nil {
+			return fmt.Errorf("failed to read license file: %w", err)
+		}
+
+		if err := validateLicense(licenseData); err != nil {
```
agnivade, line 117: Reading from the file still seems like the right way for me. We can tweak the tests to move the license contents to a test asset file and read from it.

agarciamontoro: I don't really see any more value in one or the other alternative, to be honest. I usually use "easier for testing" as a proxy for slightly better code, but it doesn't really matter where to get the license data from, a string or an asset file.

So I'll revert and use an asset file, then :)

cpoile (approved):

agnivade (approved):

agnivade:
cc @streamer45 as FYI

PR #833, opened 2024-10-16
[MM-61113] Allow multiple customizable subnets

fmartingr:
#### Summary
- deployer config changes:
  - `ClusterSubnetID` -> `ClusterSubnetIDs`
  - `ElasticsearchSettings.VPCID` removed in favor of `ClusterVpcId`
- All resources are now matched with the specified VPC in the configuration
- Elasticsearch now has specific configuration for AZ Awareness if more than one subnet is set
- Fixed a bug where we wouldn't wait for terraform stdout resulting in incomplete output from commands. 61f26c20c8505eba6e1646dd9beec4082c1ab0e5

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61113


fmartingr:
> Cool! A couple of high-level comments.
> 
> > * deployer config changes:
> >   
> >   * `ClusterVpcId` -> `ClusterVPCIDs`
> 
> I don't see this change in here. I guess you meant `ClusterSubnetID` to `ClusterSubnetIDs`.

Yeah, sorry. Edited.

> Also, is the implicit indexing driven by certain constraints or just simpler/quicker to implement? I am wondering whether a more structured setting could be easier/digestible (and potentially less error-prone) on the user side.

What do you mean by this?

streamer45:
> > Also, is the implicit indexing driven by certain constraints or just simpler/quicker to implement? I am wondering whether a more structured setting could be easier/digestible (and potentially less error-prone) on the user side.
> 
> What do you mean by this?


Having the specific index of `ClusterSubnetIDs ` refer to a resource type feels perhaps a bit overengineered. Why not define an object with N named fields (one per resource type)?

fmartingr:
> Having the specific index of `ClusterSubnetIDs ` refer to a resource type feels perhaps a bit overengineered. Why not define an object with N named fields (one per resource type)?

I had something like that before (I iterated quite a bit during this week) but if I don't specify the subnet from a `data "aws_subnet"` block Terraform considers it part of the environment and tries to delete it on the destroy operation.

fmartingr:
Refactored the code after a talk with @agarciamontoro. 

I wasn't able to reproduce the issue I had that the destroy operation was trying to destroy the subnet if directly specified by a `subnet_id` parameter.

deployment/terraform/engine.go
```diff
@@ -102,10 +102,14 @@ func _runCommand(cmd *exec.Cmd, dst io.Writer) error {
 		}
 	}()
 
-	scanner := bufio.NewScanner(stdout)
-	for scanner.Scan() {
-		mlog.Info(scanner.Text())
-	}
+	go func() {
```
agnivade, line 105: What is the benefit of doing this? Scanning stdout is not really blocking anything. And anyways you need to wait until stderr scanning is finished.

config/deployer.sample.json
```diff
@@ -5,7 +5,17 @@
...
+  "ClusterSubnetIDs": {
+    "Agent": "",
+    "App": "",
+    "Database": [],
+    "ElasticSearch": "",
+    "Job": "",
+    "Keycloak": "",
+    "Metrics": "",
+    "Proxy": "",
+    "Redis": []
```
agnivade, line 17: Why do we need a list for Redis when it's just a single server? Contrary, why we _don't_ need a list for other resources like app/agent/proxy?

deployment/terraform/assets/cluster.tf
```diff
@@ -269,6 +277,13 @@ resource "aws_iam_user_policy" "s3userpolicy" {
 EOF
 }
 
+resource "aws_db_subnet_group" "redis" {
```
agnivade, line 280: The word "db" sounds off here. Can we change to something better?

Additionally, we need index specific tags here like:

```
 tags = {
    Name = "${var.cluster_name}-proxy-${count.index}"
  }
```

deployment/terraform/assets/cluster.tf
```diff
@@ -269,6 +277,13 @@ resource "aws_iam_user_policy" "s3userpolicy" {
...
+  name       = "${var.cluster_name}-redis-group"
```
agnivade, line 281: Maybe `redis-subnet-group`?

deployment/terraform/assets/cluster.tf
```diff
@@ -296,7 +318,7 @@ resource "aws_rds_cluster" "db_cluster" {
   apply_immediately   = true
   engine              = var.db_instance_engine
   engine_version      = var.db_engine_version[var.db_instance_engine]
-
+  db_subnet_group_name = var.db_instance_count > 0 && length(var.cluster_subnet_ids) > 1 ? aws_db_subnet_group.db[0].name : ""
```
agnivade, line 321: We should also have `app_instance_count` in the condition.

deployment/terraform/assets/cluster.tf
```diff
@@ -315,6 +337,7 @@ resource "aws_rds_cluster_instance" "cluster_instances" {
   performance_insights_enabled = var.db_enable_performance_insights
   db_parameter_group_name      = length(var.db_parameters) > 0 ? "${var.cluster_name}-db-pg" : ""
   availability_zone            = var.aws_az
+  db_subnet_group_name         = var.db_instance_count > 0 && length(var.cluster_subnet_ids) > 1 ? aws_db_subnet_group.db[0].name : ""
```
agnivade, line 340: Same as above

deployment/terraform/utils.go
```diff
@@ -240,6 +240,16 @@ func fillConfigTemplate(configTmpl string, data map[string]any) (string, error)
 	return buf.String(), nil
 }
 
+// convertToTerraformVar converts a list parameter to a json encoded string
+func convertToTerraformVar[T any](param T) string {
+	result, err := json.Marshal(param)
+	if err != nil {
+		return ""
```
agnivade, line 247: Worth logging the error?

deployment/terraform/engine.go
fmartingr: Without this I was not getting the entire output from terraform, some details about errors where missing.

config/deployer.sample.json
fmartingr: I could use lists and then index by `count`, but I favoured simplicity in this first PR. if you think that's necessary it's a easy change.

deployment/terraform/assets/cluster.tf
fmartingr: That's the name of the AWS resource, I can't change that.

config/deployer.sample.json
fmartingr: Went ahead and did the change: 46be0e8d9d7e64f71b55dd68d69a19803cb4bcf5

deployment/terraform/engine.go
agnivade: Ok, that's weird. I don't understand how this changes anything though. Probably we can look into this in an isolated scenario. FYI @agarciamontoro @streamer45

config/deployer.sample.json
agnivade: Ah no no, my point was if Redis is a single server, then we don't need a list there. Unless the list is mandatory somehow.

fmartingr: I've made the list change and if we only need one subnet (because there's only one server) we can just set only one subnet in the setting (or leave it empty to use the default from `aws_subnet.selected`). Feels simpler if every setting works the same way.

agnivade: Ok, conceptually it just felt odd to have an array when it's just one resource. But not a big deal.

fmartingr: maybe if we use `aws_elasticache_replication_group` in the future?

deployment/terraform/engine.go
agnivade: Hmm .. ok I understand the bug. https://pkg.go.dev/os/exec#Cmd.StdoutPipe

> [Cmd.Wait](https://pkg.go.dev/os/exec#Cmd.Wait) will close the pipe after seeing the command exit, so most callers need not close the pipe themselves. It is thus incorrect to call Wait before all reads from the pipe have completed.

So using a scanner to read from the pipe is error-prone and probably why it fails a lot of times. The better way is to directly set the StdErr/Stdout fields with a custom io.Writer.

```go
type cmdLogger struct {
}

func (*cmdLogger) Write(in []byte) (int, error) {
	mlog.Info(string(in))
	return len(in), nil
}
```

and then 

```go
cmd.Stdout = &cmdLogger
cmd.Stderr = cmd.Stdout

if err := cmd.Start(); err != nil {
return err
}

if err := cmd.Wait(); err != nil {
return err
}
```

Having the same writer for stdout and stderr is handled properly in the stdlib:
> // If Stdout and Stderr are the same writer, and have a type that can
	// be compared with ==, at most one goroutine at a time will call Write.

@fmartingr - can we try with something like this?

fmartingr:
> @fmartingr - Here: https://github.com/mattermost/mattermost-cloud-monitoring/pull/765/files#diff-af79a285e58611c0f303f2d367ca5d6de36a521a3e8375f3c092f2fda2e0c742R5, I can see `aws_elasticache_subnet_group`. Is there something different that forces us to use only `aws_db_subnet_group`?

I just followed some documentation. I will change the resource type, but it's really weird that it let me use it.

agnivade (approved):
Thanks.

I would like to see https://github.com/mattermost/mattermost-load-test-ng/pull/833#discussion_r1808429565 being tracked separately if you don't want to do this in this PR itself.

deployment/config.go
```diff
@@ -139,6 +139,24 @@ func (t TerraformMap) String() string {
...
+	Metrics       []string `json:"metrics"`
+	Keycloak      []string `json:"keycloak"`
+	Database      []string `json:"database"`
+	Redis         []string `json:"redis"`
+}
+
+// IsAnySet returns true if any of the subnet ids are set.
+func (c *ClusterSubnetIDs) IsAnySet() bool {
+	return len(c.App) > 0 || len(c.Job) > 0 || len(c.Proxy) > 0 || len(c.Agent) > 0 || len(c.ElasticSearch) > 0 || len(c.Metrics) > 0 || len(c.Keycloak) > 0 || len(c.Database) > 0 || len(c.Redis) > 0
+}
```
streamer45, line 158: I think that using something like `reflect.DeepEqual` to avoid every single field and automatically support future ones may be a worthwhile simplification given performance is not really a concern. Not a huge deal though.

streamer45 (approved):
Looks good!

agarciamontoro (changes requested):
Amazing job, and incredibly thorough! I left a couple of comments below. Thank you so much for all your work here, @fmartingr <3

deployment/terraform/utils.go
```diff
@@ -240,6 +240,17 @@ func fillConfigTemplate(configTmpl string, data map[string]any) (string, error)
...
+// convertToTerraformVar converts a list parameter to a json encoded string
+func convertToTerraformVar[T any](param T) string {
+	result, err := json.Marshal(param)
+	if err != nil {
+		mlog.Error("failed to convert parameter to terraform var", mlog.Any("param", param), mlog.Err(err))
+		return ""
+	}
+
+	return string(result)
+}
```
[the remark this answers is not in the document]
fmartingr: As per the code, this converts any parameter since it just marshals into json, but happy to make it just a `String()` function on the type.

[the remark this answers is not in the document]
agarciamontoro: > but happy to make it just a String() function on the type

That's what I did for `TerraformMap`, the type of `CustomTags`. That way you can simply pass it to `Sprintf`. I don't have a super strong opinion on this, but I find it slightly better, since it isolates its function.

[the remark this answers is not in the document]
fmartingr: https://github.com/mattermost/mattermost-load-test-ng/pull/833/commits/4f0a3626163c5fba4245679b2bd9bce2d97f091a

agarciamontoro (approved):
Thank you! Again, amazing work :)

PR #834, opened 2024-10-17
MM-61111: Use default AWS credentials when AWSProfile is unset

agarciamontoro:
#### Summary
Up until now, we relied on the `AWSProfile` to point to the credentials profile that needs to be used to deploy in AWS. That setting even defaulted to `"mm-loadtest"`. However, when running the tool in CI for the automated load-tests, we need to use [the default AWS credentials chain](https://aws.github.io/aws-sdk-go-v2/docs/configuring-sdk/#specifying-credentials). For that, I've made `AWSProfile` optional, adding the corresponding flag/attribute if and only if that setting is set and non-empty.

The PR mostly refactors code around so that we only need to conditionally add the profile to the command (or directly in the .tf file) in one place. The commits are self-contained and hopefully self-explanatory :)

(I've also finally committed the whitespace changes my editor always does when I modify the cluster.tf file)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61111


agarciamontoro:
@toninis, if you're curious, these are the changes I needed to perform in the tool. Pretty straightforward in the end, and it has already been tested in the Github Action.

agnivade (approved):
Nice one

toninis (approved):

deployment/config.go
```diff
@@ -19,8 +19,8 @@ var esDomainNameRe = regexp.MustCompile(`^[a-z][a-z0-9\-]{2,27}$`)
 // Config contains the necessary data
 // to deploy and provision a load test environment.
 type Config struct {
-	// AWSProfile is the name of the AWS profile to use for all AWS commands
-	AWSProfile string `default:"mm-loadtest"`
+	// AWSProfile is the optional name of the AWS profile to use for all AWS commands
+	AWSProfile string `default:""`
```
streamer45, line 23: I suppose we don't really need to define a default anymore, if it's empty.

streamer45 (approved):
Makes sense 👍

deployment/config.go
agarciamontoro: True, but I like it being explicit

PR #837, opened 2024-10-23
MM-61232: Deployment failures when ClusterSubnetIDs is not specified

agarciamontoro:
#### Summary
This PR fixes a bunch of errors experienced when `ClusterSubnetIDs`, introduced in #833, is not specified. The commits are self-explanatory, but in short:
- Remove the now inexistent `ElasticSearchSettings.VpcID` from the sample and doc files.
- Fix the `String()` method in the new `ClusterSubnetIDs` type. It turns out you cannot define it with a pointer receiver if you want fmt to make use of it!
- Unrelated to the original PR, but affecting it: fix the use of `reflect.MakeSlice`. We need to pass a slice type (such as `[]string`), not a basic type (such as `string`).
- Thanks to the previous fix, we now can set the default length of `ClusterSubnetIDs`' slices to 0, so that they are not nil, but empty slices.
- Fix `IsAnySet`, which turned out to be more complex than just doing a DeepEqual. I chose iterating over all fields in the struct, but I'm open to other ideas here.
- Fix a bunch of subnet-related stuff in `cluster.tf`:
  - Check for its length instead of comparing against the empty string
  - Use the proper lists in a couple of places (DB and Redis) - @fmartingr, please double-check this!
  - Modify how we define the `aws_subnets.selected` datasource, conditionally specifying the filters.
  - The change I need @fmartingr to triple-check: instead of using the `aws_subnets.selected` when no subnet lists are specified, simply set `subnet_id` to `null` (this is the same as not specifying `subnet_id` at all in the resource). Without this change, deployments failed because they were using a specific subnet of the selected ones, and it seems it had no inbound access from the internet, so for example the file providers failed. And now that I'm writing this I'm realizing: do we really need the `aws_subnets.selected` data source *at all*? Probably not, we can discuss.
- Also, add some more tests!

Sorry for not having caught these things in the original PR, I should have tested it with an undefined `ClusterSubnetIDs`!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61232


agnivade:
Thanks for the due diligence Alejandro. Nice work.

fmartingr (approved):
Hey @agarciamontoro, huge apologies for this. I caught most of these the other night but in between tests and being late I forgot to create the PR myself. Thanks for catching all and fix the AZ for the aws_subnet is greatly appreciated. 

> The change I need @fmartingr to triple-check: instead of using the aws_subnets.selected when no subnet lists are specified, simply set subnet_id to null (this is the same as not specifying subnet_id at all in the resource). Without this change, deployments failed because they were using a specific subnet of the selected ones, and it seems it had no inbound access from the internet, so for example the file providers failed. And now that I'm writing this I'm realizing: do we really need the aws_subnets.selected data source at all? Probably not, we can discuss.

Regarding this, I will defer to you guys. In my head I'd rather have the same subnet for all resources by default (now that we can specify custom subnets for resources) rather than leaving terraform do. But since you are running this more than I do (or not? 😓) I prefer you to decide what default suits your use case more.

PS: Use TOML. TOML <3 

Apologies again.

deployment/terraform/assets/cluster.tf
```diff
@@ -280,7 +288,7 @@ EOF
 resource "aws_elasticache_subnet_group" "redis" {
   name       = "${var.cluster_name}-redis-subnet-group"
   subnet_ids = (length(var.cluster_subnet_ids.redis) > 0) ? tolist(var.cluster_subnet_ids.redis) : tolist(data.aws_subnets.selected.ids)
```
fmartingr, line 290:
```suggestion
  subnet_ids = (length(var.cluster_subnet_ids.redis) > 1) ? tolist(var.cluster_subnet_ids.redis) : tolist(data.aws_subnets.selected.ids)
```

deployment/terraform/assets/cluster.tf
```diff
@@ -299,13 +307,13 @@ resource "aws_elasticache_cluster" "redis_server" {
   port                 = 6379
   security_group_ids   = [aws_security_group.redis[0].id]
   availability_zone    = var.aws_az
-  subnet_group_name    = var.redis_enabled && length(var.cluster_subnet_ids) > 1 ? aws_elasticache_subnet_group.redis[0].name : ""
+  subnet_group_name    = var.redis_enabled && length(var.cluster_subnet_ids.redis) > 1 ? aws_elasticache_subnet_group.redis[0].name : ""
 }
 
 resource "aws_db_subnet_group" "db" {
   name       = "${var.cluster_name}-db-subnet-group"
   subnet_ids = (length(var.cluster_subnet_ids.database) > 0) ? tolist(var.cluster_subnet_ids.database) : tolist(data.aws_subnets.selected.ids)
```
fmartingr, line 315: Subnet groups require 2+ subnets

```suggestion
  subnet_ids = (length(var.cluster_subnet_ids.database) > 1) ? tolist(var.cluster_subnet_ids.database) : tolist(data.aws_subnets.selected.ids)
```

deployment/config.go
```diff
@@ -144,23 +144,40 @@ func (t TerraformMap) String() string {
...
+	Agent         []string `default_size:"0" json:"agent"`
+	ElasticSearch []string `default_size:"0" json:"elasticsearch"`
+	Metrics       []string `default_size:"0" json:"metrics"`
+	Keycloak      []string `default_size:"0" json:"keycloak"`
+	Database      []string `default_size:"0" json:"database"`
+	Redis         []string `default_size:"0" json:"redis"`
 }
 
 // IsAnySet returns true if any of the subnet ids are set.
 func (c *ClusterSubnetIDs) IsAnySet() bool {
```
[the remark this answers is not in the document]
agarciamontoro: @fmartingr, see this: https://go.dev/play/p/2_aHkpxHFeX In short: a `ClusterSubnetIDs` with non-nil, but empty lists, is different than a `ClusterSubnetIDs{}`, whose lists are all nil.

deployment/terraform/assets/cluster.tf
```diff
@@ -37,15 +37,23 @@ locals {
...
-    values = [var.cluster_vpc_id]
+    values = var.cluster_vpc_id == "" ? [data.aws_vpc.default.id] : [var.cluster_vpc_id]
   }
 
-  filter {
-    name   = "availability-zone"
-    values = [var.aws_az]
+  // Add a filter on the availability zone only if it is explicitly set
+  dynamic "filter" {
+    for_each = var.aws_az != "" ? [1] : []
```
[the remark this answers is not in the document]
agarciamontoro: HT Marcin: https://stackoverflow.com/a/68249905/3248221

agarciamontoro:
@fmartingr, no need to apologize for anything! This is a team effort, and you've been already working incredibly tirelessly on this! Thank you for all your work :)

> Regarding this, I will defer to you guys. In my head I'd rather have the same subnet for all resources by default (now that we can specify custom subnets for resources) rather than leaving terraform do. But since you are running this more than I do (or not? 😓) I prefer you to decide what default suits your use case more.

I'm not opposed to using a single subnet for everything. But we would need to specify one that we can access from outside. I wasn't able to successfully finish a deploy with the selected one because it had no inbound internet access (that's my guess, at least, I didn't check *which* subnet we were using, but the error was always that the host was unreachable). When using no subnet, the host was indeed reachable. Not sure how all this plays out, to be honest. If we specify no subnet, the resource is not attached to one?

> PS: Use TOML. TOML <3

I agree it's better. I need to make an effort and start using it! But do you mean it would have helped here?

fmartingr:
> I'm not opposed to using a single subnet for everything. But we would need to specify one that we can access from outside. I wasn't able to successfully finish a deploy with the selected one because it had no inbound internet access (that's my guess, at least, I didn't check _which_ subnet we were using, but the error was always that the host was unreachable). When using no subnet, the host was indeed reachable. Not sure how all this plays out, to be honest. If we specify no subnet, the resource is not attached to one?

Mhmmm. I always thought the instances have to be attached to a subnet by default. Maybe the `aws_subnet.default` ordering selects a private one, and leaving `subnet_id = nil` finds out the first public one? Unsure how that works.

> > PS: Use TOML. TOML <3
> 
> I agree it's better. I need to make an effort and start using it! But do you mean it would have helped here?

It's being really helpful to me in terms of iteration by allowing me to comment out properties without having to take separate notes of values and adding text comments around.

deployment/config.go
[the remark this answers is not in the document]
fmartingr: owwww, you are right. Good catch.

agarciamontoro:
> Mhmmm. I always thought the instances have to be attached to a subnet by default. Maybe the aws_subnet.default ordering selects a private one, and leaving subnet_id = nil finds out the first public one? Unsure how that works.

Ok, I think I understand it now: resources are indeed attached to the [default subnet](https://docs.aws.amazon.com/vpc/latest/userguide/default-subnet.html) if no `subnet_id` is specified. The default subnet is usually public, unless it is manually modified (not our case, see for example our [default subnet for us-east-1a](https://us-east-1.console.aws.amazon.com/vpcconsole/home?region=us-east-1#SubnetDetails:subnetId=subnet-0df4c0f8057dc62ae)). Now, we can either leave the `subnet_id` out, in which case it seems that resources end up in the same subnet (the default one), or use this advanced resource, [aws_default_subnet](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_subnet), that "adopts" the default subnet into the Terraform plan but does not create it (unless it doesn't exist) or destroy it (as long as its `force_destroy` argument is `false`, which is the default). I think I'd lean towards simply leaving the `subnet_id` unset, although I see some benefit in being explicit as well. In any case, I don't see any benefit to using the aws_subnets.selected data source anymore. I'm looking at its uses after this PR and it's only there for `aws_elasticache_subnet_group` and for `aws_db_subnet_group`, but both are created only if their corresponding subnet lists are specified, in which case they use those, not the ones from `aws_subnets.selected`.

Am I missing something?

fmartingr:
> > Mhmmm. I always thought the instances have to be attached to a subnet by default. Maybe the aws_subnet.default ordering selects a private one, and leaving subnet_id = nil finds out the first public one? Unsure how that works.
> 
> Ok, I think I understand it now: resources are indeed attached to the [default subnet](https://docs.aws.amazon.com/vpc/latest/userguide/default-subnet.html) if no `subnet_id` is specified. The default subnet is usually public, unless it is manually modified (not our case, see for example our [default subnet for us-east-1a](https://us-east-1.console.aws.amazon.com/vpcconsole/home?region=us-east-1#SubnetDetails:subnetId=subnet-0df4c0f8057dc62ae)). Now, we can either leave the `subnet_id` out, in which case it seems that resources end up in the same subnet (the default one), or use this advanced resource, [aws_default_subnet](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/default_subnet), that "adopts" the default subnet into the Terraform plan but does not create it (unless it doesn't exist) or destroy it (as long as its `force_destroy` argument is `false`, which is the default). I think I'd lean towards simply leaving the `subnet_id` unset, although I see some benefit in being explicit as well. In any case, I don't see any benefit to using the aws_subnets.selected data source anymore. I'm looking at its uses after this PR and it's only there for `aws_elasticache_subnet_group` and for `aws_db_subnet_group`, but both are created only if their corresponding subnet lists are specified, in which case they use those, not the ones from `aws_subnets.selected`.
> 
> Am I missing something?

I took a look and yes, that's about right. Since the default subnet can be customized we can just `nil` `subnet_id` if the config does not have subnets defined.

agarciamontoro:
> I took a look and yes, that's about right. Since the default subnet can be customized we can just nil subnet_id if the config does not have subnets defined.

Perfect! Data source removed now :)

deployment/terraform/assets/cluster.tf
```diff
@@ -37,16 +37,8 @@ locals {
...
-  filter {
-    name   = "vpc-id"
-    values = [var.cluster_vpc_id]
-  }
-
-  filter {
-    name   = "availability-zone"
-    values = [var.aws_az]
-  }
+data "aws_vpc" "default" {
```
fmartingr, line 40: Where are you using this?

agarciamontoro: Good catch! It was used in the definition of `aws_subnets.selected`

deployment/config_test.go
```diff
@@ -165,3 +167,95 @@ func TestTerraformMapString(t *testing.T) {
...
+		require.Len(t, cfg.ClusterSubnetIDs.Redis, 0)
+	})
+
+	t.Run("String() of default values", func(t *testing.T) {
+		cfg := Config{}
+		defaults.Set(&cfg)
+
+		expected := `{"app":[],"job":[],"proxy":[],"agent":[],"elasticsearch":[],"metrics":[],"keycloak":[],"database":[],"redis":[]}`
+		//nolint:gosimple
+		actual := fmt.Sprintf("%s", cfg.ClusterSubnetIDs)
```
streamer45, line 249: Wouldn't assigning `cfg.ClusterSubnetIDs.String()` be sufficient or you actually wanted to test from `fmt` methods?

streamer45 (approved):
Thanks for taking care of this! I now feel slightly bad for suggesting `DeepEqual` in the first place :sweat:

deployment/config_test.go
agarciamontoro: I actually wanted to test from `fmt` methods. The bug that prompted this in the first place was not that `String` was buggy, but that it was declared with a pointer receiver. In that case, `fmt` does not call `String`.

agarciamontoro:
> I now feel slightly bad for suggesting DeepEqual in the first place 😓

Haha, no worries!

PR #840, opened 2024-10-28
[MM-60297] Fix `loadtest_http_errors_total` metric

streamer45:
#### Summary

The issue causing the mismatch in 403 errors was due to the extremely high cardinality of the `path` label used in the `loadtest_http_errors_total` counter vector. The errors were tracked correctly by the client, but the `increase` query would return an empty set. 

image

To address this, we replace the IDs in the request path to keep the cardinality low. 

image

If we need that information, we can always find it in the proxy access logs. Speaking of which, I reviewed what we were logging to include more useful info and using JSON for easier filtering/parsing in Loki.

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-60297

### DIFF
--- a/loadtest/user/userentity/user.go
+++ b/loadtest/user/userentity/user.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"net/http"
 	"os"
+	"regexp"
 	"time"
 
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/store"
@@ -38,6 +39,8 @@ const (
 	AuthenticationTypeOpenID     = "openid"
 	AuthenticationTypeSAML       = "saml"
 )
+
+var stripIDsRE = regexp.MustCompile(`\b\w{26}\b`)
 
 // Config holds necessary information required by a UserEntity.
 type Config struct {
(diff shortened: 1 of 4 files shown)


loadtest/user/userentity/user.go
```diff
@@ -39,6 +40,8 @@ const (
 	AuthenticationTypeSAML       = "saml"
 )
 
+var stripIDsRE = regexp.MustCompile(`\b\w{26}\b`)
```
agnivade, line 43: This regex is making me uncomfortable. What about file URLs? And also variations in query params? It feels like it leaves out a lot of corner cases. Can we take a simplistic approach where we make the buckets more coarse by reducing the path further, but we ensure that it's a clean solution?

agarciamontoro: > What about file URLs? And also variations in query params?

Can you give examples of these? I've been using this sort of regex when manually grepping in the logs several times and I've never found an edge case (or I have not realized I did, of course :joy:)

streamer45:
> LGTM! The improvement in the logs is great, thank you!
> 
> > The errors were tracked correctly by the client, but the increase query would return an empty set.
> 
> Why did it return an empty set? Did it look for specific IDs?

It just couldn't process the high number of labels I think. I didn't investigate that much since it was clear that was the problem. For example, doing a `sum by` directly worked fine but that only returned the sum. Applying the increase just didn't work.

loadtest/user/userentity/user.go
streamer45: @agnivade It would literally capture just the ids at this point. Query params would not be full words due to the `?` so they wouldn't get captured unless the value itself is 26 chars (e.g. `/api/v4/something?query=aaaaaaaaaaaaaaaaaaaaaaaaaa`). 

We could add the `/` matching (before and after the word) and replace the resulting captured group if you really wanted that extra check, but we are talking about a metrics label in a load-test tool, not sure how strict we need to be here.

agnivade: Ah nvm, `Path` doesn't contain query params. So all good there. And I checked the file URLs, they also have an id so we should be fine for now.

agnivade (approved):

PR #846, opened 2024-11-11
Misleading error when generating a report with no deployment

kczpl:
#### Summary

The problem is that when running local tests, the report generator errors out with a misleading message instead of explicitly stating that there is no deployment and therefore cannot generate a report. The error message should be clearer.

This pull request changes the error message when there is no ongoing deployment while generating a report. It improves error handling by checking whether the current Terraform state includes the metrics instance and errors out when it doesn't. To implement this, the `ExternalDBSettings.DriverName` configuration was changed to use PostgreSQL by default.

#### Ticket Link

--> https://mattermost.atlassian.net/browse/MM-56949
  Github issue: https://github.com/mattermost/mattermost/issues/29148

### DIFF
--- a/cmd/ltctl/report.go
+++ b/cmd/ltctl/report.go
@@ -92,7 +92,7 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		}
 
 		if !output.HasMetrics() {
-			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag.\nIf you have a local or manually deployed Prometheus server running")
+			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag. If you have a local or manually deployed Prometheus server running")
 		}
 
 		promURL = "http://" + output.MetricsServer.PublicIP + ":9090"
--- a/cmd/ltctl/report.go
+++ b/cmd/ltctl/report.go
@@ -76,7 +76,7 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 	}
 
 	if promURL == "" {
-		fmt.Printf("Flag --prometheus-url is not set. Defaulting to the deployment's Prometheus server...")
+		fmt.Println("Flag --prometheus-url is not set. Defaulting to the deployment's Prometheus server...")
 		config, err := getConfig(cmd)
 		if err != nil {
 			return err
@@ -92,7 +92,7 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		}
 
 		if !output.HasMetrics() {
-			return fmt.Errorf("the output has no metrics, no active deployment found")
+			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag.\nIf you have a local or manually deployed Prometheus server running")
 		}
 
 		promURL = "http://" + output.MetricsServer.PublicIP + ":9090"
--- a/cmd/ltctl/report.go
+++ b/cmd/ltctl/report.go
@@ -92,7 +92,7 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		}
 
 		if !output.HasMetrics() {
-			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag. If you have a local or manually deployed Prometheus server running")
+			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag if you have a local or manually deployed Prometheus server running")
 		}
 
 		promURL = "http://" + output.MetricsServer.PublicIP + ":9090"
--- a/cmd/ltctl/report.go
+++ b/cmd/ltctl/report.go
@@ -90,6 +90,11 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("could not parse output: %w", err)
 		}
+
+		if !output.HasMetrics() {
+			return fmt.Errorf("the output has no metrics, no active deployment found")
+		}
+
 		promURL = "http://" + output.MetricsServer.PublicIP + ":9090"
 	}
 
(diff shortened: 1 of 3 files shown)


agarciamontoro:
Thank you for the PR, @kczpl! Will take a look as soon as I can :)

kczpl:
Thank you! I've also responded in the contributors channel thread.

If my understanding of the issue or approach is incorrect, I'm ready to continue working on this, as I'm still unsure about my solution. :))

cmd/ltctl/report.go
```diff
@@ -90,6 +90,11 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("could not parse output: %w", err)
 		}
+
+		if !output.HasMetrics() {
+			return fmt.Errorf("the output has no metrics, no active deployment found")
```
agarciamontoro, line 95: Here we can return something slightly more useful to the user, in the lines of:
```suggestion
			return fmt.Errorf("no active deployment found, use the --prometheus-url flag if you have a local or manually deployed Prometheus server running")
```

agarciamontoro (changes requested):
Thanks, @kczpl! The code looks good, just a couple of comments:
1. If no deployer.json is configured at all, there is another error that is returned earlier. I'm thinking that we can fix this by setting the default of `TerraformStateDir` as `./ltstatus` (I don't really know why I didn't do that when I added it, to be honest). The error is the following:
```
> go run ./cmd/ltctl report generate --output=base.out --label=base "2022-10-20 08:43:39" "2022-10-20 08:44:53"
Flag --prometheus-url is not set. Defaulting to the deployment's Prometheus server...Error: failed to create terraform engine: not enough permissions to create Terraform state directory "/var/lib/mattermost-load-test-ng".
Here's some alternatives you can try:
	1. Change the TerraformStateDir setting in config/deployer.json to a directory you have permissions over (recommended).
	2. Manually create the currently configured directory "/var/lib/mattermost-load-test-ng" and change its owner to your current user.
	3. Run this and all next commands as root (not recommended).
exit status 1
```
2. It seems we're missing a newline when we return the `Flag --prometheus-url is not set. Defaulting to the deployment's Prometheus server...` error, so the new error is a bit buried Can you add it? Here's what I see:
```
> go run ./cmd/ltctl report generate --output=base.out --label=base "2022-10-20 08:43:39" "2022-10-20 08:44:53"
File base.out exists. Overwrite ? (Y/n) y
Flag --prometheus-url is not set. Defaulting to the deployment's Prometheus server...Error: the output has no metrics, no active deployment found
exit status 1
```

cmd/ltctl/report.go
kczpl: sure thing! just added that

cmd/ltctl/report.go
```diff
@@ -90,6 +90,11 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("could not parse output: %w", err)
 		}
+
+		if !output.HasMetrics() {
+			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag.\nIf you have a local or manually deployed Prometheus server running")
```
agnivade, line 95: No need to break the line. It's a single error message. Let the user's console break the line for us :)

deployment/config.go
```diff
@@ -93,7 +93,8 @@ type Config struct {
 	Report                report.Config
 	// Directory under which the .terraform directory and state files are managed.
 	// It will be created if it does not exist
-	TerraformStateDir string `default:"/var/lib/mattermost-load-test-ng" validate:"notempty"`
+	// ./ltstate is the default value used when `deployer.json`` is configured at all
+	TerraformStateDir string `default:"./ltstate" validate:"notempty"`
```
agnivade, line 97: Fine with the change, but is this somehow related to the PR?

cmd/ltctl/report.go
kczpl: Makes sense :)) I've just pushed the removal

cmd/ltctl/report.go
```diff
@@ -90,6 +90,11 @@ func RunGenerateReportCmdF(cmd *cobra.Command, args []string) error {
 		if err != nil {
 			return fmt.Errorf("could not parse output: %w", err)
 		}
+
+		if !output.HasMetrics() {
+			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag. If you have a local or manually deployed Prometheus server running")
```
agarciamontoro, line 95: This is actually the same sentence! Sorry I wasn't clear before, but what I meant with the message is that if you have deployed Prometheus yourself (not with the tool), then you need to use the `--prometheus-url` flag. I think this is clear, but I'm open to other suggestions:
```suggestion
			return fmt.Errorf("no active deployment found, use the `--prometheus-url` flag if you have a local or manually deployed Prometheus server running")
```

deployment/config.go
agarciamontoro, line 97: I think this comment is not needed, since the tag is clear in my opinion:
```suggestion
	TerraformStateDir string `default:"./ltstate" validate:"notempty"`
```

agarciamontoro (changes requested):
Just a couple of nits in the comments, but other than that it looks good!

deployment/config.go
agarciamontoro: @agnivade: Not the original goal, but to get to the correct state we wanted, we need it.

cmd/ltctl/report.go
kczpl: I should have removed the capital letter obviously while removing `\n`.
 In my opinion, this message is quite clear:
`no active deployment found, use the `--prometheus-url` flag if you have a local or manually deployed Prometheus server running`

deployment/config.go
agnivade: Ah, understood!

agnivade (approved):
Nicely done @kczpl !

agarciamontoro (approved):
Thank you, @kczpl, this is great! I've already tested it, so we can get this merged :)

kczpl:
@agarciamontoro @agnivade 
thanks guys! I'm looking forward to more open issues in this repo

PR #847, opened 2024-11-11
MM-60286: Fix deletion of DB param group

agnivade:
We need to use the create_before_destroy
lifecycle configuration to allow Terraform
to properly change settings of the parameter
group: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/db_parameter_group#create_before_destroy-lifecycle-configuration.

While here, we also remove installation
of inux-tools-aws-lts-22.04.
More context on that: https://community.mattermost.com/core/pl/a7krecfgjpyoty8q7edy61k4ao

Bonus fix: we create the redis security
group conditionally as well. Just one of those
small oversights.

https://mattermost.atlassian.net/browse/MM-60286


agnivade:
@streamer45 / @agarciamontoro - Just checking if you haven't forgotten about this one :D

agarciamontoro (commented):
I don't really understand why we need any of those two changes, let me ask a couple of questions:
1. I guess we need `create_before_destroy` because the DB param group is being somehow modified? But the problem happened when it was being destroyed, right?
2. Why do we need the 30-second duration *at all*? That feels like the sort of hacks Terraform is specifically designed to avoid, right?

agnivade:
1. Somehow the DB param group gets modified during destroy. I don't know how that seems to be the case.
2. This is to enforce a strict ordering of the deletion. I think we can even make it 1 second and it shouldn't matter. Although the param group should only get destroyed after the DB is destroyed, but there's something which does not let that happen cleanly. I don't know what. I could run another test with this sleep removed to see if it works without that or not, because I was testing with multiple variations and this one worked. But I can check whether removing the sleep also works fine or not.

agarciamontoro:
> Somehow the DB param group gets modified during destroy. I don't know how that seems to be the case.

Hm, interesting.

> But I can check whether removing the sleep also works fine or not.

@agnivade, I'd really appreciate it. If we can avoid that hack it would be good.

agnivade:
Tested, and we don't need the sleep!

streamer45 (approved):
Cool!

agarciamontoro (approved):
Yay, thanks!

PR #848, opened 2024-11-14
Scheduled post

harshilsharma63:
#### Summary
Adding scheduled posts support to load test tool.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-61486


agnivade (commented):
It looks like some debug `fmt.Println` statements are still left over from debugging. Let's clean them up.

loadtest/store/memstore/random.go
```diff
@@ -448,3 +449,70 @@ func (s *MemStore) RandomDraftForTeam(teamId string) (string, error) {
...
+			selectedInnerMap = innerMap
+			break
+		}
+		outerIndex++
+	}
+
+	// Check if the selected inner map is empty
+	if len(selectedInnerMap) == 0 {
+		return nil, errors.New("no posts available in selected category")
+	}
```
agarciamontoro, line 480: I think we shouldn't error out if there are no posts in a single inner map. In that case, we should pick another one. Maybe what we can do is: build a slice of all keys whose inner map contain at least one element. Then, get a random element from that slice, and then get a random post from that inner map. Something like the following (not tested):
```go
var keys []string
for key, innerMap := range s.scheduledPosts {
	if len(innerMap) > 0 {
		keys = append(keys, key)
	}
}

if len(keys) == 0 {
	return nil, errors.New("no posts available")
}

selectedInnerMap := s.scheduledPosts[keys[rand.Intn(len(keys))]

File: loadtest/store/memstore/random.go [hunk 2]
```diff
@@ -448,3 +449,70 @@ func (s *MemStore) RandomDraftForTeam(teamId string) (string, error) {
...
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	for teamId := range s.scheduledPosts {
+		if _, ok := s.scheduledPosts[teamId][scheduledPostID]; ok {
+			delete(s.scheduledPosts[teamId], scheduledPostID)
+			break
+		}
+	}
+}
```
agarciamontoro, line 507: Do we know the team ID when deleting the post? We could skip the outer loop if we do.

loadtest/user/userentity/scheduled_posts.go
```diff
@@ -0,0 +1,104 @@
...
+	createdScheduledPost, _, err := ue.client.CreateScheduledPost(context.Background(), scheduledPost)
+	if err != nil {
+		fmt.Println("CreateScheduledPost: CreateScheduledPost error", err)
+		return err
+	}
+
+	id := scheduledPost.ChannelId
+	if createdScheduledPost.RootId != "" {
+		id = scheduledPost.RootId
+	}
```
agarciamontoro, line 28: Is this `ChannelId`? Does this mean that we can schedule a single post per channel? This happens in `UpdateScheduledPost` as well.

loadtest/user/userentity/scheduled_posts.go
```diff
@@ -0,0 +1,104 @@
...
+
+	for teamIdInResponse := range scheduledPostsByTeam {
+		for _, scheduledPost := range scheduledPostsByTeam[teamIdInResponse] {
+			err := ue.store.SetScheduledPost(teamID, scheduledPost.Id, scheduledPost)
+			if err != nil {
+				fmt.Println("GetTeamScheduledPosts: SetScheduledPost error", err)
+				return err
+			}
+		}
+	}
```
agarciamontoro, line 101: I didn't understand this code, so I took a look at the server code and saw that we use `directChannels` as a special team ID here to get posts scheduled in DMs. Is this something that can break some assumptions elsewhere in this code? I suppose no, but wanted to double-check.
Nit: also, should we add a `ue.store.SetScheduledPostsInTeam` so that we can pass it the whole slice of posts?

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	scheduledPost := &model.ScheduledPost{
+		Draft: model.Draft{
+			Message:   message,
+			ChannelId: channel.Id,
+			RootId:    rootId,
+			CreateAt:  model.GetMillis(),
+		},
+		ScheduledAt: loadtest.RandomFutureTime(17_28_00_000, 10),
```
agarciamontoro, line 58: What does `17_28_00_000` mean?

loadtest/utils.go
```diff
@@ -87,3 +89,19 @@ func nextPowerOf2(val int) int {
 	val++
 	return val
 }
+
+// RandomFutureTime generates a random time between (now + xMilliseconds) to (now + xDays) in Unix milliseconds.
```
agarciamontoro, line 93: Nit: can we use the name of the variables themselves? It took me a while to understand the comment
```suggestion
// RandomFutureTime generates a random time between (now + millisecondsFromNow) to (now + uptoDaysFromNow) in Unix milliseconds.
```

loadtest/store/memstore/store.go
```diff
@@ -52,6 +52,7 @@ type MemStore struct {
 	featureFlags        map[string]bool
 	report              *model.PerformanceReport
 	channelBookmarks    map[string]*model.ChannelBookmarkWithFileInfo
+	scheduledPosts      map[string]map[string]*model.ScheduledPost
```
agnivade, line 55: Also, please update the `(s *MemStore) Clear()` method to clear this field as well.

loadtest/store/memstore/random.go
```diff
@@ -448,3 +449,70 @@ func (s *MemStore) RandomDraftForTeam(teamId string) (string, error) {
...
+			selectedPost = post
+			break
+		}
+		innerIndex++
+	}
+
+	return selectedPost, nil
+}
+
+func (s *MemStore) DeleteScheduledPost(scheduledPostID string) {
```
agnivade, line 497: The delete and update methods should not be in random.go. Let's move this to memstore/store.go.

agnivade (changes requested):

loadtest/control/simulcontroller/controller.go
```diff
@@ -288,6 +288,30 @@ func getActionList(c *SimulController) []userAction {
...
+			run:              c.deleteScheduledPost,
+			frequency:        0.1,
+			minServerVersion: semver.MustParse("10.3.0"),
+		},
+		{
+			name:             "SendScheduledPost",
+			run:              c.sendScheduledPost,
+			frequency:        0.1,
+			minServerVersion: semver.MustParse("10.3.0"),
+		},
```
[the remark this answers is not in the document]
harshilsharma63: There is no real data as this is a new feature. I've created https://mattermost.atlassian.net/browse/MM-61907 to update these frequencies in the future.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,184 @@
...
+		return control.UserActionResponse{Info: "no scheduled posts found"}
+	}
+
+	if err := u.DeleteScheduledPost(scheduledPost.Id); err != nil {
+		fmt.Println("deleteScheduledPost: DeleteScheduledPost error", err)
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	u.Store().DeleteScheduledPost(scheduledPost.Id)
+	return control.UserActionResponse{Info: fmt.Sprintf("scheduled post deleted with id %v", scheduledPost.Id)}
```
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func (c *SimulController) sendScheduledPost(u user.User) control.UserActionResponse {
```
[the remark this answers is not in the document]
harshilsharma63: Done, renamed to `sendScheduledPostNow`. This runs the manual Send Now option available on scheduled posts.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,184 @@
...
+		fmt.Println("sendScheduledPost: CreatePost error", err)
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	if err := u.DeleteScheduledPost(scheduledPost.Id); err != nil {
+		fmt.Println("sendScheduledPost: DeleteScheduledPost error", err)
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	u.Store().DeleteScheduledPost(scheduledPost.Id)
```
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,184 @@
...
+
+	return control.UserActionResponse{Info: fmt.Sprintf("scheduled post sent with id %v", scheduledPost.Id)}
```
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/store/memstore/random.go
```diff
@@ -448,3 +449,70 @@ func (s *MemStore) RandomDraftForTeam(teamId string) (string, error) {
 
 	return draftIDs[rand.Intn(len(draftIDs))], nil
 }
+
+func (s *MemStore) GetRandomScheduledPost() (*model.ScheduledPost, error) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	// Seed the random generator
+	rand.Seed(time.Now().UnixNano())
```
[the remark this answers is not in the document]
harshilsharma63: Got it, done.

loadtest/store/memstore/random.go
harshilsharma63: Done.

loadtest/user/userentity/scheduled_posts.go
harshilsharma63: Ah damn, I need to save an array of scheduled posts, not just one scheduled post. Fixed it now.

So yeah, we can have multiple scheduled posts per channel or per thread, this if fixed now.

loadtest/user/userentity/scheduled_posts.go
harshilsharma63: Only the scheduled posts code touches that data, so we are safe with this. This hardcoded key isn't used anywhere else, so it doesn't create problems with other parts of the code.

loadtest/control/simulcontroller/scheduled_posts.go
harshilsharma63: Its 2 days in milliseconds. I've created a const for this to make it more readable.

loadtest/utils.go
harshilsharma63: Done.

loadtest/store/memstore/store.go
harshilsharma63: Done.

loadtest/store/memstore/random.go
harshilsharma63: Done.

harshilsharma63:
@agnivade @agarciamontoro can you please re-review this? I've made the fixes.

agnivade (commented):
I will leave the store logic to Alejandro.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,184 @@
...
+		},
+		ScheduledAt: loadtest.RandomFutureTime(17_28_00_000, 10),
+	}
+
+	if rand.Float64() < 0.02 {
+		if err := control.AttachFilesToDraft(u, &scheduledPost.Draft); err != nil {
+			fmt.Println("createScheduledPost: AttachFilesToDraft error", err)
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+	}
```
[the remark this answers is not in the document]
agarciamontoro: Well, I didn't mean the number `0.02` specifically, but a constant for when to attach files to a post, for which we use 2%. We use that exact meaning elsewhere (probably less than 5 places, but more than 1, I'd say)

loadtest/utils.go
```diff
@@ -87,3 +89,19 @@ func nextPowerOf2(val int) int {
 	val++
 	return val
 }
+
+// RandomFutureTime generates a random time between (now + millisecondsFromNow) to (now + uptoDaysFromNow) in Unix milliseconds.
+func RandomFutureTime(millisecondsFromNow int64, uptoDaysFromNow int) int64 {
```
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/user/userentity/scheduled_posts.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (ue *UserEntity) GetTeamScheduledPosts(teamID string) error {
+	scheduledPostsByTeam, _, err := ue.client.GetUserScheduledPosts(context.Background(), teamID, true)
+	if err != nil {
+		return err
+	}
+
+	for teamIdInResponse := range scheduledPostsByTeam {
+		for _, scheduledPost := range scheduledPostsByTeam[teamIdInResponse] {
```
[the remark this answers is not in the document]
harshilsharma63: I'd still need two loops. scheduledPostsByTeam is a `map[string][]*model.ScheduledPosts`, and I need to iterate the individual scheduled posts from that array. I could use `for _, v := range scheduledPostsByTeam` but then I'd need another loop to iterate items from `v`.

loadtest/store/store.go
```diff
@@ -200,6 +204,10 @@ type MutableUserStore interface {
 	// SetDrafts stores the given drafts.
 	SetDrafts(teamId string, drafts []*model.Draft) error
 
+	// scheduled posts
+	SetScheduledPost(teamId string, scheduledPost *model.ScheduledPost) error
+	GetRandomScheduledPost() (*model.ScheduledPost, error)
```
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/control/simulcontroller/scheduled_posts.go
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/user/userentity/scheduled_posts.go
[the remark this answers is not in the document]
agnivade: Ah right. Indeed.

agnivade (approved):
Looking good!

loadtest/store/memstore/store.go
```diff
@@ -1356,6 +1359,53 @@ func (s *MemStore) DeleteChannelBookmark(bookmarkId string) error {
...
+func (s *MemStore) UpdateScheduledPost(teamId string, scheduledPost *model.ScheduledPost) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	if _, ok := s.scheduledPosts[teamId]; !ok {
+		s.scheduledPosts[teamId] = make(map[string][]*model.ScheduledPost)
+	}
+
+	s.scheduledPosts[teamId][scheduledPost.Id] = append(s.scheduledPosts[teamId][scheduledPost.Id], scheduledPost)
+}
```
agarciamontoro, line 1411: Why do we append to `scheduledPosts` instead of updating the post itself? Also, shouldn't we use the channel or thread ID instead of `scheduledPost.Id` in the inner map?

loadtest/store/memstore/store.go
```diff
@@ -1356,6 +1359,53 @@ func (s *MemStore) DeleteChannelBookmark(bookmarkId string) error {
...
 	return nil
 }
+
+func (s *MemStore) DeleteScheduledPost(scheduledPostID string) {
+	s.lock.Lock()
+	defer s.lock.Unlock()
+
+	for teamId := range s.scheduledPosts {
+		if _, ok := s.scheduledPosts[teamId][scheduledPostID]; ok {
+			delete(s.scheduledPosts[teamId], scheduledPostID)
```
agarciamontoro, line 1396: Same here: do we index the inner map by the post's ID or by the channel/thread ID?

loadtest/user/userentity/scheduled_posts.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = ue.store.SetScheduledPost(teamId, updatedScheduledPost)
+	if err != nil {
+		return err
+	}
+
+	ue.Store().UpdateScheduledPost(teamId, updatedScheduledPost)
```
agarciamontoro, line 45: Do we need to set it and then update it? Also, we should check the error from UpdateScheduledPost

loadtest/store/memstore/store.go
```diff
@@ -1356,6 +1359,53 @@ func (s *MemStore) DeleteChannelBookmark(bookmarkId string) error {
...
+			break
+		}
+	}
+}
+
```
agarciamontoro, line 1401: Should we error if the post doesn't exist?

loadtest/user/userentity/scheduled_posts.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (ue *UserEntity) DeleteScheduledPost(scheduledPostId string) error {
+	_, _, err := ue.client.DeleteScheduledPost(context.Background(), scheduledPostId)
+	if err != nil {
+		return err
+	}
+
+	ue.Store().DeleteScheduledPost(scheduledPostId)
+	return nil
```
agarciamontoro, line 57: Right now DeleteScheduledPost doesn't return an error, but maybe it should (I left a comment for that). If we decide that's a good decision, then we need to return its error.

loadtest/utils.go
```diff
@@ -87,3 +89,11 @@ func nextPowerOf2(val int) int {
...
 	return val
 }
+
+func RandomFutureTime(deltaStart, maxUntil time.Duration) int64 {
+	now := time.Now()
+	diff := maxUntil - deltaStart
+	offset := deltaStart.Milliseconds() + rand.Int63n(diff.Milliseconds())
+	randomFutureTime := now.Add(time.Duration(offset) * time.Millisecond)
+	return randomFutureTime.UnixMilli()
+}
```
agarciamontoro, line 99: Let's add a small comment here (I think I got the interval right, but please review it yourself, it's confusing :joy:):
```suggestion
// RandomFutureTime returns a random Unix timestamp, in milliseconds, in the interval
// [now+deltaStart, now+deltaStart+maxUntil]
func RandomFutureTime(deltaStart, maxUntil time.Duration) int64 {
	now := time.Now()
	diff := maxUntil - deltaStart
	offset := deltaStart.Milliseconds() + rand.Int63n(diff.Milliseconds())
	randomFutureTime := now.Add(time.Duration(offset) * time.Millisecond)
	return randomFutureTime.UnixMilli()
}
```

agarciamontoro (changes requested):
Thanks, @harshilsharma63, and sorry for the delay in the review! Some more comments below.

loadtest/user/userentity/scheduled_posts.go
harshilsharma63: I don't think its of any use returning error from delete method, it doesn't matter

loadtest/store/memstore/store.go
harshilsharma63: It doesn't really matter.

loadtest/utils.go
harshilsharma63: Done, updated.

loadtest/store/memstore/store.go
harshilsharma63: Fixed.

loadtest/user/userentity/scheduled_posts.go
harshilsharma63: Just update should be fine, fixed this.

loadtest/store/memstore/store.go
harshilsharma63: Fixed.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+	if scheduledPost == nil {
+		return control.UserActionResponse{Info: "no scheduled posts found"}
+	}
+
+	if err := u.DeleteScheduledPost(scheduledPost.Id); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("scheduled post deleted with id %v", scheduledPost.Id)}
```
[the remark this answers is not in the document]
harshilsharma63: Fixed.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	scheduledPost.Message = message
+	scheduledPost.ScheduledAt = loadtest.RandomFutureTime(time.Hour*24*2, time.Hour*24*10)
+
+	if err := u.UpdateScheduledPost(channel.TeamId, scheduledPost); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("scheduled post updated in channel id %v", channel.Id)}
```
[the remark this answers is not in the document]
harshilsharma63: Fixed.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+		if err := control.AttachFilesToDraft(u, &scheduledPost.Draft); err != nil {
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+	}
+
+	if err := u.CreateScheduledPost(channel.TeamId, scheduledPost); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("scheduled post created in channel id %v", channel.Id)}
```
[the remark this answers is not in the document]
harshilsharma63: Fixed.

loadtest/user/userentity/scheduled_posts.go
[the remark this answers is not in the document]
harshilsharma63: Done.

loadtest/utils.go
```diff
@@ -90,6 +90,8 @@ func nextPowerOf2(val int) int {
 	return val
 }
 
+// RandomFutureTime returns a random Unix timestamp, in milliseconds, in the interval
+// [now+deltaStart, now+maxUntil]
```
agarciamontoro, line 94: The comment is actually correct, thanks! But it surfaced a bug in this function: if `maxUntil` is smaller than `deltaStart` (let's say, you call `RandomFutureTime(10*time.Second, 9*time.Second)`, then `rand.Int63n` panics because it gets a negative number). We should fix the function so that the interval is actually `[now+deltaStart, now+deltaStart+maxUntil]`, and add unit testing for this function.

agarciamontoro (changes requested):
Thanks, @harshilsharma63! I've left a comment in the `RandomFutureTime` function, and I have an additional ask: I do think we should error out in `MemStore`'s `DeleteScheduledPost` and `UpdateScheduledPost` if the post is not found, and log an error if that's the case. It shouldn't happen and may point to other problems in the load-test.

loadtest/utils.go
harshilsharma63: Done.

agnivade:
Both Alejandro and Claudio are out. I will take another quick look and ask for a stamp approval.

agnivade:
I gave it a test run and found some minor issues. We should be good to go after that.

**1. Incorrect error/info logging when there are no scheduled posts.**

Log line is:

```
error [2024-12-23 14:51:17.856 +05:30] control/simulcontroller.(*SimulController).sendScheduledPostNow loadtest/control/simulcontroller/scheduled_posts.go:135 no scheduled posts available caller="loadtest/loadte
st.go:60" controller_id=4 user_id=anoxyyicofbozjrig7og95i6py
...
error [2024-12-23 14:51:34.667 +05:30] control/simulcontroller.(*SimulController).deleteScheduledPost loadtest/control/simulcontroller/scheduled_posts.go:113 no scheduled posts available caller="loadtest/loadtes
t.go:60" controller_id=10 user_id=bpjjn1f67jg8xnya68m41ki6dc
```

Note that the log level is error, but it's just a case of scheduled posts not being available. And we are correctly doing it here:

```go
scheduledPost, err := u.Store().GetRandomScheduledPost()
if err != nil {
}
if scheduledPost == nil {
return control.UserActionResponse{Info: "no scheduled posts found"}
}
```

But the issue is that `GetRandomScheduledPost()` already returns an error if no scheduled posts are found.

The right way is to return an error variable and check for it. See for example `RandomTeam`:

```go
	if len(teams) == 0 {
		return model.Team{}, ErrTeamStoreEmpty
	}
```

and then we check for the error at call site:

```go
team, err := u.Store().RandomTeam(store.SelectMemberOf | store.SelectNotCurrent)
if errors.Is(err, memstore.ErrTeamStoreEmpty) {
return control.UserActionResponse{Info: "no other team to switch to"}
} else if err != nil {
}
```

We should apply the same logic while calling `GetRandomScheduledPost` as well.

**2. Invalid scheduled at time:**

```
error [2024-12-23 14:52:18.420 +05:30] control/simulcontroller.(*SimulController).createScheduledPost loadtest/control/simulcontroller/scheduled_posts.go:63 Invalid scheduled at time., id=m9txr5p37bf63kg4xwq7upjxcy caller="loadtest/loadtest.go:60" controller_id=9 user_id=b1ysdcjc5prodfef1386jdjawy
```

This is what I see in the server logs:

```
debug [2024-12-23 15:01:16.543 +05:30] Invalid scheduled at time.                    caller="web/context.go:120" path=/api/v4/posts/schedule request_id=xzx84obdgjgdugma4setb8a1ye ip_addr=127.0.0.1 user_id=eh1k7cuhmtntfx8y9ziihatmkr method=POST err_where=ScheduledPost.IsValid http_code=400 error="ScheduledPost.IsValid: Invalid scheduled at time., id=qoafc3rp3tf5zjsqpcck9ew55w"
```

If we fix the above 2 issues, we should be good to go.

agarciamontoro:
Thanks for taking a look, @agnivade!

harshilsharma63:
I'll address the change next week and I'm getting the edit file functionality ready for feature complete by Monday

agarciamontoro:
Got it, @harshilsharma63, and thank you for your work here!

agnivade:
@harshilsharma63 - Just checking to see the status on this. I think there are just a few minor things pending, and then we should be good to merge this. So hopefully not a lot of effort required.

harshilsharma63:
> @harshilsharma63 - Just checking to see the status on this. I think there are just a few minor things pending, and then we should be good to merge this. So hopefully not a lot of effort required.

I'm working to finish scheduled posts on mobile before the feature complete date so have deferred this until that. Should be done in Feb first half.

agnivade:
Thanks. Ideally, we should not be releasing features which are not load tested. @streamer45 - wondering if you can help bump up the priority on this so that Harshil is free to wrap up the work?

loadtest/control/actions.go
```diff
@@ -1034,6 +1034,15 @@ func DraftsEnabled(u user.User) (bool, UserActionResponse) {
 
 func ChannelBookmarkEnabled(u user.User) (bool, UserActionResponse) {
 	allow := u.Store().FeatureFlags()["ChannelBookmarks"]
+	return allow, UserActionResponse{}
+}
+
+func ScheduledPostsEnabled(u user.User) (bool, UserActionResponse) {
+	allow, err := strconv.ParseBool(u.Store().ClientConfig()["ScheduledPosts"])
+	if err != nil {
+		fmt.Println("Error parsing ScheduledPosts config", err)
```
streamer45, line 1043: Unnecessary log?

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+
+	channel, err := u.Store().CurrentChannel()
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	var rootId = ""
+	if rand.Float64() < 0.25 {
+		post, err := u.Store().RandomPostForChannel(channel.Id)
+		if err == nil {
```
streamer45, line 28: We should handle the error case.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+				rootId = post.Id
+			}
+		}
+	}
+
+	if err := sendTypingEventIfEnabled(u, channel.Id); err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	message, err := createMessage(u, channel, false)
```
streamer45, line 41: I assume we can schedule replies as well. A case to consider adding.

loadtest/utils_test.go
```diff
@@ -0,0 +1,97 @@
...
+	end := start.Add(maxUntil)
+
+	randomTime := RandomFutureTime(deltaStart, maxUntil)
+
+	if randomTime < start.Unix() || randomTime > end.Unix() {
+		t.Errorf("RandomFutureTime() = %v, want between %v and %v", randomTime, start.Unix(), end.Unix())
+	}
+}
+
+func TestRandomFutureTimeZeroDuration(t *testing.T) {
```
streamer45, line 26: nit: I'd expect these to be subcases of `TestRandomFutureTime` rather than top-level tests.

loadtest/utils_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	now := time.Now()
+	start := now.Add(deltaStart)
+	end := start.Add(maxUntil)
+
+	randomTime := RandomFutureTime(deltaStart, maxUntil)
+
+	if randomTime < start.Unix() || randomTime > end.Unix() {
+		t.Errorf("RandomFutureTime() = %v, want between %v and %v", randomTime, start.Unix(), end.Unix())
+	}
```
streamer45, line 23: We may want to use the `require` package for these checks.

loadtest/utils.go
```diff
@@ -87,3 +89,24 @@ func nextPowerOf2(val int) int {
 	val++
 	return val
 }
+
+// RandomFutureTime returns a random Unix timestamp, in milliseconds, in the interval
+// [now+deltaStart, now+deltaStart+maxUntil]
+func RandomFutureTime(deltaStart, maxUntil time.Duration) int64 {
```
streamer45, line 95: I'd find it more versatile if we returned a `time.Time` and then let the caller format it as needed. Otherwise, if we can only foresee a timestamp usage, we could rename this to reflect its behavior better.

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	scheduledPost := &model.ScheduledPost{
+		Draft: model.Draft{
+			Message:   message,
+			ChannelId: channel.Id,
+			RootId:    rootId,
+			CreateAt:  model.GetMillis(),
+		},
+		ScheduledAt: loadtest.RandomFutureTime(time.Hour*24*2, time.Hour*24*10),
```
streamer45, line 53: I'd really welcome named constants to make it straightforward to read, e.g. `TwoDays`, `TenDays` :)

loadtest/control/simulcontroller/scheduled_posts.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (c *SimulController) updateScheduledPost(u user.User) control.UserActionResponse {
+	if ok, resp := control.ScheduledPostsEnabled(u); resp.Err != nil {
+		return resp
+	} else if !ok {
+		return control.UserActionResponse{Info: "scheduled posts not enabled"}
+	}
+
+	scheduledPost, err := u.Store().GetRandomScheduledPost()
```
streamer45, line 76: Are we ensuring `GetRandomScheduledPost()` always returns non-deleted posts, if any?

streamer45 (commented):
Looks good overhaul. Left mostly minor suggestions.

One additional comment is that I don't see any websocket addition here, and wondering how we are keeping the state in sync across multiple clients (e.g. deleting a draft).

agnivade:
@harshilsharma63 - Just wanted to check on this. Would you be able to find some time to address the review comments?

harshilsharma63:
@agnivade will do this tomorrow. Noted.

loadtest/control/simulcontroller/scheduled_posts.go
harshilsharma63: Done, but this doesn't matter as all thats happening from isReply param is changing the avg word length, nothing else The part that send scheduled post as a reply is above where I set the rootId. But fixed this anyways as it is now semantically correct.

loadtest/control/actions.go
harshilsharma63: yeah, removed.

loadtest/control/simulcontroller/scheduled_posts.go
harshilsharma63: Done.

loadtest/utils_test.go
harshilsharma63: Done.

loadtest/utils_test.go
harshilsharma63: Done.

loadtest/utils.go
harshilsharma63: There's just one use case of this function right now and I don't want to spend excessive time refactoring this to accommodate all possible cases. This is fine for now can can be updated later as needed. I've already refactored this twice.

loadtest/control/simulcontroller/scheduled_posts.go
harshilsharma63: We'd end up creating constants for every day. Skipping this for now.

loadtest/control/simulcontroller/scheduled_posts.go
harshilsharma63: While there is no explicit check for it, just like in other random functions, I am removing scheduled posts from store when deleting or posting them, so it will not return a deleted scheduled post.

harshilsharma63:
@agnivade I've fixed the comments and invalid time error.

agnivade:
Thanks!

streamer45:
@harshilsharma63 Thoughts on my [top comment](https://github.com/mattermost/mattermost-load-test-ng/pull/848#pullrequestreview-2575764250) above? On the websocket part.

harshilsharma63:
> One additional comment is that I don't see any websocket addition here, and wondering how we are keeping the state in sync across multiple clients (e.g. deleting a draft).

@streamer45 I didn't know we had to do it. Can you share an example where this is being done and I'll refer it?

agnivade:
Basically look at the simulcontroller/websocket.go file. We need to add event handlers for scheduled post events and keep the load-test agent state up to date. Although we don't do it for a lot of other events. But I agree that we should start doing it.

streamer45:
Yeah, the idea is to mimic a real client's behavior as much as possible including any synchronization logic that may happen as a result of receiving a websocket event. Usually, I'd expect a Redux handler to be in place on the web implementation, which on this side can be converted to use the local store. See how we handle the posted event as a practical example:

https://github.com/mattermost/mattermost-load-test-ng/blob/4954198fea1907f09f9edbbc7887ebd15de8d655/loadtest/user/userentity/websocket.go#L70-L96

https://github.com/mattermost/mattermost-load-test-ng/blob/4954198fea1907f09f9edbbc7887ebd15de8d655/loadtest/control/simulcontroller/websocket.go#L32-L62

You can notice we have two main places to handle ws events, based on what needs to happen. If we just need to save the data to the store, then we modify the internal `userentity/websocket.go` file. If we need to execute some actions as a result of receiving the event, then we'd be touching the `simulcontroller/websocket.go`.

agnivade:
@streamer45 - what are your thoughts on splitting the ws event handler to another PR? This PR has been up for 4 months and quite big already. I am thinking if we could merge this and send another PR after this for the ws handler. That way atleast we start getting some feature coverage for scheduled posts.

streamer45:
> @streamer45 - what are your thoughts on splitting the ws event handler to another PR? I am thinking if we could merge this and send another PR after this for the ws handler. That way atleast we start getting some feature coverage for scheduled posts.

Not opposed. Mostly I'd like to understand (from someone who developed the functionality) what ws events we need to handle and what the handling looks like. Do we make additional API requests in response? Is it just about saving data to the store?

These questions help me understand what the performance impact of leaving these out could be.

> This PR has been up for 4 months and quite big already.

I'd encourage us to understand more about that (I have a few thoughts, e.g. the choice of 3 reviewers). Maybe a quick retro would be beneficial since we want to make this process as seamless as possible.

I'll remove myself from review to unblock and let @agarciamontoro do a final pass. Thanks for your patience :)

harshilsharma63:
@streamer45 

> Mostly I'd like to understand (from someone who developed the functionality) what ws events we need to handle and what the handling looks like. Do we make additional API requests in response? Is it just about saving data to the store?

just to keep state updated, thats the only purpose of WS handling.

> I'd encourage us to understand more about that (I have a few thoughts, e.g. the choice of 3 reviewers). Maybe a quick retro would be beneficial since we want to make this process as seamless as possible.

The PR has been up for 4 months because I started working on another task and couldn't stop working on it to get back to this.

agnivade:
> I'd encourage us to understand more about that (I have a few thoughts, e.g. the choice of 3 reviewers). 

To be clear, I wasn't explicitly looking for your review when I said the following :stuck_out_tongue: 

> @streamer45 - wondering if you can help bump up the priority on this so that Harshil is free to wrap up the work?

I was hoping you could have a chat with the leads so that Harshil gets enough bandwidth to prioritise this load-test work. Otherwise, he was constantly getting pulled into developing other features. But you did a review, so no harm done :smile: 

@agarciamontoro - could you give this a final pass? Thanks.

@harshilsharma63 - As Claudio outlined here: https://github.com/mattermost/mattermost-load-test-ng/pull/848#issuecomment-2679175864, we can set up event handlers for those ws events and update the store of the load-test agent to keep the state in sync. Let's schedule that work as a separate PR.

harshilsharma63:
@agarciamontoro can you give this a final pass? I'll merge it after your review. I'll create a ticket to address websocket work.

loadtest/control/simulcontroller/controller.go
```diff
@@ -288,6 +292,30 @@ func getActionList(c *SimulController) []userAction {
...
+			run:              c.deleteScheduledPost,
+			frequency:        0.1,
+			minServerVersion: semver.MustParse("10.3.0"),
+		},
+		{
+			name:             "SendScheduledPost",
+			run:              c.sendScheduledPostNow,
+			frequency:        0.1,
+			minServerVersion: semver.MustParse("10.3.0"),
+		},
```
agarciamontoro, line 318: These frequencies seem to be quite lower in real servers. Following https://github.com/mattermost/mattermost-load-test-ng/blob/master/docs/coverage-frequency.md and taking a look at Community and the largest customer in Cloud, the ratio of these API handlers against CreatePost doesn't even show up in the time windows with the most traffic. The largest ratio over the last 7 days is 0.04, but that's on very low-traffic time windows.
In short: I'd apply a frequency of 0.001 (which is the one we use for minimally-used endpoints) to all of these actions.

agarciamontoro (changes requested):
Looking good! Thank you, @harshilsharma63 :) I left a comment to update the frequencies. Thanks!

loadtest/control/simulcontroller/controller.go
harshilsharma63: Done.

agarciamontoro (approved):
Thank you for all your work here, @harshilsharma63! Testing will happen in the release process, that we're starting today, so we can merge this already :)

harshilsharma63:
Thanks @agarciamontoro !

agnivade:
Only took 3 months for the PR to be merged :)

harshilsharma63:
> Only took 3 months for the PR to be merged :)

Well aged PR 🤌

PR #852, opened 2024-11-22
[MM-59994] Fix `ethtool` metrics

streamer45:
#### Summary

We need to restart the service so that it picks up the updated environment file with the enabled collector.

#### Ticket

https://mattermost.atlassian.net/browse/MM-59994


agarciamontoro (approved):
Oh, that makes sense indeed.

agnivade (approved):
Oh I see, so it was never working. Since the flag was already added before my PR, I (incorrectly) assumed that it was working. Thanks!

PR #853, opened 2024-11-25
MM-61944: Decouple vacuum step from Terraform.Create and use it in Comparison

agarciamontoro:
#### Summary
Last week, @streamer45 realized that we're not running the `vacuum` step in comparison runs. Actually, we do run it, but we do it *before* loading the database dump, which is of course completely useless.

This PR refactors that step (and the rest of post-processing we do after the DB is ready: clear licenses data and run extra SQL commands) out from `Terraform.Create`, so we can call it when needed in both usages of that function: in the usual `deployment create` command and in the `comparison run` command.
 
There's still a lot of work to decouple everything going on in `Terraform.Create`, but this is a baby step in that direction.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61944

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/mattermost/mattermost-load-test-ng/defaults"
 	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform"
+	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
 	"github.com/mattermost/mattermost-load-test-ng/logger"
 	"github.com/mattermost/mattermost/server/public/model"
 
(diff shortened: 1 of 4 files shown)


agarciamontoro:
Actually, @fmartingr, can you take a look as well? Specially to the extra SQL commands stuff, since that's something you have more context on. The ordering of steps is still correct, right?

fmartingr (approved):
LGTM. I think that at some point we need to get the `InitData` value from the deployer.json or a flag.

cmd/ltctl/main.go
```diff
@@ -28,12 +29,20 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
 
-	initData := config.DBDumpURI == ""
-	err = t.Create(initData)
+	extAgent, err := ssh.NewAgent()
```
streamer45, line 32: Curious, is there any advantage in passing this agent around vs creating it inside these methods (`Create` and `PostProcessDatabase`)?

streamer45 (approved):
Thanks! I agree it's messy and could use some more love.

agnivade:
Thanks Felipe and Claudio for taking care of the review!

agarciamontoro:
They were quick!

PR #856, opened 2024-11-25
chore: get aws client with region set up in config

fmartingr:
#### Summary

We have the AWS Region parameter but that can conflict with the default values or the profile values (specifically if region is not set there).

This pull request adds the region option when creating the AWS Client to match the one in the configuration.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61279

### DIFF
--- a/deployment/terraform/utils.go
+++ b/deployment/terraform/utils.go
@@ -342,12 +342,17 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
 // GetAWSConfig returns the AWS config, using the profile configured in the
 // deployer if present, and defaulting to the default credential chain otherwise
 func (t *Terraform) GetAWSConfig() (aws.Config, error) {
+	regionOpt := awsconfig.WithRegion(t.config.AWSRegion)
+
 	if t.config.AWSProfile == "" {
-		return awsconfig.LoadDefaultConfig(context.Background())
+		return awsconfig.LoadDefaultConfig(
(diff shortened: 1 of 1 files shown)


agarciamontoro (commented):
How does this behave when no region nor profile is set? We need to use the default credential chain in the automated load-tests, so it's important that we keep that same behaviour.

fmartingr:
> How does this behave when no region nor profile is set? We need to use the default credential chain in the automated load-tests, so it's important that we keep that same behaviour.

I have not tested this scenario, thought that the region field was mandatory (since it has a default value) but it can be emptied. Have we emptied the field explicitly in the CI configuration files? How does the automated load test get the credentials?

agarciamontoro:
Oh, I checked and I have that set up in the automated load-test config templates as well. Could you run a quick test to verify all is good with that set of files? https://github.com/mattermost/mattermost-load-test-ng/tree/master/examples/config/ci

fmartingr:
> Oh, I checked and I have that set up in the automated load-test config templates as well. Could you run a quick test to verify all is good with that set of files? https://github.com/mattermost/mattermost-load-test-ng/tree/master/examples/config/ci

Just created a deployment from this branch used those settings. The envionment was created without errors.

streamer45 (approved):
Make sense :+1:

agarciamontoro (approved):
Thanks, @fmartingr, looks great :)

PR #857, opened 2024-11-25
fix: escape keycloak password to improve shell compat

fmartingr:
#### Summary
Use a naive approach to password escaping for keycloak if the tilde character is included in the password.

An improvement for this would be to limit the characters in the field validation.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61278

### DIFF
--- a/deployment/terraform/keycloak.go
+++ b/deployment/terraform/keycloak.go
@@ -171,7 +171,12 @@ func (t *Terraform) setupKeycloak(extAgent *ssh.ExtAgent) error {
 	}
 
 	// Authenticate as admin to execute keycloak commands
-	_, err = sshc.RunCommand(fmt.Sprintf(`%s/kcadm.sh config credentials --server http://127.0.0.1:8080 --user "%s" --password '%s' --realm master`, keycloakBinPath, t.config.ExternalAuthProviderSettings.KeycloakAdminUser, strings.ReplaceAll(t.config.ExternalAuthProviderSettings.KeycloakAdminPassword, "'", "\\'")))
+	_, err = sshc.RunCommand(
+		fmt.Sprintf(`%s/kcadm.sh config credentials --server http://127.0.0.1:8080 --user %q --password %q --realm master`,
+			keycloakBinPath,
+			t.config.ExternalAuthProviderSettings.KeycloakAdminUser,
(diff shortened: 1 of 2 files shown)


deployment/terraform/keycloak.go
```diff
@@ -171,7 +171,7 @@ func (t *Terraform) setupKeycloak(extAgent *ssh.ExtAgent) error {
 	}
 
 	// Authenticate as admin to execute keycloak commands
-	_, err = sshc.RunCommand(fmt.Sprintf("%s/kcadm.sh config credentials --server http://127.0.0.1:8080 --user %s --password %s --realm master", keycloakBinPath, t.config.ExternalAuthProviderSettings.KeycloakAdminUser, t.config.ExternalAuthProviderSettings.KeycloakAdminPassword))
+	_, err = sshc.RunCommand(fmt.Sprintf(`%s/kcadm.sh config credentials --server http://127.0.0.1:8080 --user "%s" --password '%s' --realm master`, keycloakBinPath, t.config.ExternalAuthProviderSettings.KeycloakAdminUser, strings.ReplaceAll(t.config.ExternalAuthProviderSettings.KeycloakAdminPassword, "'", "\\'")))
```
agnivade, line 174: I thought `%q` was meant to solve this scenario? Does that not work? :thinking:

agnivade (approved):

fmartingr:
I tried a bunch of things before leaving as it is and entirely depends on the special characters being used. I think it would be better to just validate the field to be alphanumeirc only, so I will make the necessary modifications to this PR, moving to draft.

agnivade (approved):

agarciamontoro (commented):
What if we connect this to a real server? The password may not fulfill this criteria, right?

fmartingr:
> What if we connect this to a real server? The password may not fulfill this criteria, right?

Load test code is not prepared to connect to an external keycloak server at the moment, so that's a non-issue. Furthermore this requirement is in place to avoid errors when running commands via SSH in the keycloak server, which I believe it would not apply if we were to provide an external server.

agarciamontoro:
> Load test code is not prepared to connect to an external keycloak server at the moment, so that's a non-issue

Thanks for the context, I thought otherwise! Then it all looks good to me :)

agarciamontoro (approved):

PR #860, opened 2024-11-29
MM-60630: Automated loadtests configuration templates

agarciamontoro:
#### Summary
This PR adds the needed config templates for CI to run the automated load-tests.

It also formats cluster.tf (there's always some formatting to do there) and removes the `lts` version of the `linux-aws-tools` in the proxy, which was still there.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-60630


examples/config/ci/config.json
```diff
@@ -0,0 +1,46 @@
...
+    "AdminEmail": "sysadmin@sample.mattermost.com",
+    "AdminPassword": "Sys@dmin-sample1"
+  },
+  "UserControllerConfiguration": {
+    "Type": "simulative",
+    "RatesDistribution": [
+      {
+        "Rate": 5.4,
+        "Percentage": 1
+      }
```
streamer45, line 14: So, did we settle on the single rate for these? Is the multi-rate used anywhere at this point?

examples/config/ci/coordinator.json
```diff
@@ -0,0 +1,102 @@
...
+        "Query": "(avg(rate(node_netstat_Tcp_RetransSegs{instance=~\"proxy:9100\"}[1m])) / avg(rate(node_netstat_Tcp_OutSegs{instance=~\"proxy:9100\"}[1m]))) * 100",
+        "Threshold": 0.5,
+        "MinIntervalSec": 60,
+        "Alert": true
+      }
+    ]
+  },
+  "NumUsersInc": 8,
+  "NumUsersDec": 8,
+  "RestTimeSec": 1,
```
streamer45, line 91: Just noting this doesn't make too much sense if `UpdateIntervalMs` is still at 2s.

examples/config/ci/deployer.json
```diff
@@ -0,0 +1,153 @@
+{
+  "AWSProfile": "",
+  "AWSRegion": "us-east-1",
+  "AWSAvailabilityZone": "",
```
streamer45, line 4: Eventually, I think we'd want to try and use the same AZ, if possible, to contain costs unless it becomes a nightmare of flakiness. Trying other regions may help with that.

examples/config/ci/deployer.json
```diff
@@ -0,0 +1,153 @@
...
+  "AWSAMI": "ami-003d3d03cfe1b0468",
+  "ClusterName": "undefined",
+  "ClusterVpcID": "",
+  "AppInstanceCount": 2,
+  "AppInstanceType": "c7i.large",
+  "MetricsInstanceType": "t3.xlarge",
+  "AgentInstanceCount": 6,
+  "AgentInstanceType": "c7i.xlarge",
+  "ElasticSearchSettings": {
```
streamer45, line 13: Do we have plans to use ES? We have 20 million posts after all. My rationale here is the more we test, the higher the chance of spotting regressions in the PRs.

examples/config/ci/deployer.json
```diff
@@ -0,0 +1,153 @@
...
+    "KeycloakDBDumpURI": "",
+    "GenerateUsersCount": 0
+  },
+  "MattermostDownloadURL": "https://latest.mattermost.com/mattermost-enterprise-linux",
+  "MattermostLicenseFile": "/home/runner/mattermost-load-test",
+  "MattermostConfigPatchFile": "",
+  "AdminEmail": "sysadmin@sample.mattermost.com",
+  "AdminUsername": "sysadmin",
+  "AdminPassword": "Sys@dmin-sample1",
+  "LoadTestDownloadURL": "https://github.com/mattermost/mattermost-load-test-ng/releases/download/v1.23.0-rc1/mattermost-load-test-ng-v1.23.0-rc1-linux-amd64.tar.gz",
```
streamer45, line 83: Will this need constant updating?

examples/config/ci/deployer.json
```diff
@@ -0,0 +1,153 @@
...
+  },
+  "PyroscopeSettings": {
+    "EnableAppProfiling": true,
+    "EnableAgentProfiling": true,
+    "BlockProfileRate": 0
+  },
+  "CustomTags": {
+    "Origin": "ci-automated-loadtests",
+    "PR": "undefined",
+    "SHA": "undefined"
```
streamer45, line 151: Are these filled in later? Why the undefined literal vs an empty string?

streamer45 (commented):
Good stuff, left some thoughts, nothing truly blocking though,

examples/config/ci/deployer.json
agarciamontoro: What we don't have for testing ES is an index for the 20M posts. I agree we should add it at some point, but this first iteration was planned without it.

examples/config/ci/deployer.json
agarciamontoro: This set of config will need maintaining from our side, yes. I can add a step in the release process to review this. Also, I need to change this to the final release, not the RC-1.

examples/config/ci/deployer.json
agarciamontoro: Yes, [they are](https://github.com/mattermost/delivery-platform/pull/222/files#diff-b489de800b182237e408bba9e7d5cd5fba6b55439057d750be59bd94416d62a1R148-R152)! It doesn't really matter what we put here, it could be another literal or an empty string, nothing special about `undefined`, just what I first wrote :)

examples/config/ci/deployer.json
agnivade: Something for the future. ;)

examples/config/ci/deployer.json
agnivade: Let's set it to empty. We should actually move to toml where we can add comments.

agnivade (approved):

examples/config/ci/deployer.json
agarciamontoro: Yup, but we need some more work for toml. I have some changes in a git stash forgotten somewhere.

agnivade: Indeed. I forced myself to spend some time and moved all local load-test configs to toml, and now life is great!

examples/config/ci/config.json
streamer45: Should we make it our default, then? Either that or a plan to review it, I think, would be sensible.

examples/config/ci/coordinator.json
streamer45: Just beware that if we change the logic underneath (e.g. https://mattermost.atlassian.net/browse/MM-61922), then we could get different results from one version to the next.

examples/config/ci/deployer.json
streamer45: Could we track it in Jira at least?

streamer45 (approved):

examples/config/ci/coordinator.json
agarciamontoro: Yup, that's important. There's still room for increasing this, though, I tested it with 12 users/s (which would actually be 6 users/s because `UpdateIntervalMs` was still at 2s) many many times. I'll keep an eye out, though.

examples/config/ci/deployer.json
agarciamontoro: I'd like to have a second epic on automated load-tests at some point. Added it to business-as-usual for now: https://mattermost.atlassian.net/browse/MM-62062

examples/config/ci/config.json
agarciamontoro: This number was used to match the rate of our customer's `createPost`. We should definitely review all frequencies at some point, but do we want to create a ticket for this now?

PR #861, opened 2024-11-29
MM-62026: Fix comparison build filenames

agarciamontoro:
#### Summary
This PR makes sure that we use a unique filename for each build file we need in comparisons by:
1. Using a new `getBuildFilename` function to get the filename of each build given the configuration. The function adds the config's label to the filename.
2. Ensuring the labels for both the base and new builds are different.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62026

### DIFF
--- a/comparison/config.go
+++ b/comparison/config.go
@@ -109,6 +109,11 @@ func (c *Config) IsValid() error {
 			return err
 		}
 	}
+
+	if c.BaseBuild.Label == c.NewBuild.Label {
+		return fmt.Errorf("the labels for the base and new build must be unique")
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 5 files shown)


comparison/config.go
```diff
@@ -109,6 +109,11 @@ func (c *Config) IsValid() error {
 			return err
 		}
 	}
+
+	if c.BaseBuild.Label == c.NewBuild.Label {
+		return fmt.Errorf("the labels for the base and new build must be unique")
+	}
```
streamer45, line 115: (Claudio in 2021) Who on earth would ever label base and new build the same?

streamer45 (approved):

PR #863, opened 2024-11-29
feat: execute commands async to speed up tasks

fmartingr:
#### Summary
After I started testing using a huge number of users in the same team/channel combination I experienced slowness on some of the tasks that stop/start the mattermost servers to run jobs. Those commands were being executed in serial, so I changed the function so the commands run in parallel.

### DIFF
--- a/deployment/terraform/dump.go
+++ b/deployment/terraform/dump.go
@@ -268,9 +268,7 @@ func (t *Terraform) executeCommands(commands []deployment.Cmd) error {
 		}
 
 		wg.Wait()
-		go func() {
-			close(errors)
-		}()
+		close(errors)
 
 		errorsFound := false
 		for e := range errors {
--- a/deployment/terraform/dump.go
+++ b/deployment/terraform/dump.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform/ssh"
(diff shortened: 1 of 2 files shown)


deployment/terraform/dump.go
```diff
@@ -257,12 +252,36 @@ func (t *Terraform) executeDatabaseCommands(extraCommands []deployment.Cmd) erro
...
+
 		for _, client := range c.Clients {
-			mlog.Debug("Running cmd", mlog.String("cmd", c.Value))
-			if out, err := client.RunCommand(c.Value); err != nil {
-				return fmt.Errorf("failed to run cmd %q: %w %s", c.Value, err, out)
-			}
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				mlog.Debug("Running cmd", mlog.String("cmd", c.Value))
```
agnivade, line 263: Can we also log the publicIP of the instance here to differentiate between the log lines?

deployment/terraform/dump.go
```diff
@@ -257,12 +252,36 @@ func (t *Terraform) executeDatabaseCommands(extraCommands []deployment.Cmd) erro
...
+				if out, err := client.RunCommand(c.Value); err != nil {
+					errors <- fmt.Errorf("failed to run cmd %q: %w %s", c.Value, err, out)
```
agnivade, line 265: Pass the IP of the instance, otherwise it will be difficult to debug on which machine it failed.

fmartingr: Please check: https://github.com/mattermost/mattermost-load-test-ng/pull/863/commits/4517d3859521acd1976b05ba632358fd8639d032

I prefer the hostname but that requires more changes.

agnivade (approved):

deployment/terraform/dump.go
```diff
@@ -257,12 +252,36 @@ func (t *Terraform) executeDatabaseCommands(extraCommands []deployment.Cmd) erro
...
+				if out, err := client.RunCommand(c.Value); err != nil {
+					errors <- fmt.Errorf("failed to run cmd %q on %s: %w %s", c.Value, client.IP, err, out)
+				}
+			}()
 		}
+
+		wg.Wait()
+		go func() {
+			close(errors)
+		}()
```
agarciamontoro, line 273: Why do we run this in a goroutine?

agarciamontoro (changes requested):
This looks great, thanks! Just a question below on the need of an additional goroutine.

deployment/terraform/dump.go
fmartingr: These are the remains of me doing more complex code with a `doneChan` and all that jazz, I simplified it in this way to at least maintain the "the receiver should not close the channel" mantra. Should work either way I guess.

agarciamontoro: It would work either way, indeed. I'd prefer to remove the goroutine myself, but will leave it up to you :)

agarciamontoro (approved):

PR #864, opened 2024-12-04
MM-62075: Ignore error from aws s3 rm command

agarciamontoro:
#### Summary
When destroying a cluster that deploys an S3 bucket, we try to speed up its deletion by calling `aws s3 rm s3://my-bucket` concurrently while we run `terraform destroy`. This was added to fix https://mattermost.atlassian.net/browse/MM-47263, since we were seeing too many timeouts due to the S3 bucket not being properly destroyed in time. Emptying the bucket while other resources are being destroyed has worked to fix the timeout issue, but it introduces a new error: if the bucket is destroyed by Terraform before the `aws s3 rm` command finishes, this command, and thus the whole `{comparison,deployment} destroy` command, fails.

This is a spurious failure: all resources, including the S3 bucket, have been destroyed (that's exactly why the `aws s3 rm` command fails, actually), so we can simply ignore it.

In the case this command would error in any other way that we are not accounting for as of today, the `{comparison,deployment} destroy` command would still fail, since it would notice that the S3 bucket has not been destroyed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-62075

### DIFF
--- a/deployment/terraform/destroy.go
+++ b/deployment/terraform/destroy.go
@@ -5,7 +5,6 @@ package terraform
 
 import (
 	"context"
-	"fmt"
 	"strings"
 
 	"github.com/mattermost/mattermost/server/public/shared/mlog"
@@ -28,7 +27,6 @@ func (t *Terraform) Destroy() error {
 	// See https://mattermost.atlassian.net/browse/MM-47263
 	emptyBucketCtx, emptyBucketCancel := context.WithCancel(context.Background())
 	defer emptyBucketCancel()
-	emptyBucketErrCh := make(chan error, 1)
 	go func() {
 		if t.output.HasS3Bucket() {
 			mlog.Info("emptying S3 bucket s3://" + t.output.S3Bucket.Id)
@@ -38,13 +36,13 @@ func (t *Terraform) Destroy() error {
 				"s3://" + t.output.S3Bucket.Id,
 				"--recursive",
 			}
-			if err := t.runAWSCommand(emptyBucketCtx, emptyS3BucketArgs, nil); err != nil {
-				emptyBucketErrCh <- fmt.Errorf("failed to run local cmd \"aws %s\": %w", strings.Join(emptyS3BucketArgs, " "), err)
-				return
-			}
+			// We intentionally ignore potential errors from this command,
+			// since it introduces spurious failures when the bucket is
+			// destroyed before the command finishes.
+			// See https://mattermost.atlassian.net/browse/MM-62075
+			_ = t.runAWSCommand(emptyBucketCtx, emptyS3BucketArgs, nil)
 			mlog.Info("emptied S3 bucket s3://" + t.output.S3Bucket.Id)
 		}
-		emptyBucketErrCh <- nil
 	}()
 
 	var params []string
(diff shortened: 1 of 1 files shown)


toninis:
Check https://github.com/mattermost/delivery-platform/pull/222#issuecomment-2517224229

agarciamontoro:
Thanks for taking a look, @toninis! Let me answer your comment here:

> I think you should use force_destroy option . Check [here](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/s3_bucket#force_destroy-1)

The problem here was that the whole thing was timing out because it was incredibly slow. I don't like the solution of using a concurrent `aws s3 rm`, but we found it sped things up quite a lot. That said, I believe we've never tried the `force_destroy` option, so maybe that would solve it as well. I'm a bit concerned about doing that *now*, because I'd need to run more tests and delay this further.

That's why I'm proposing getting this PR merged. If this one gets accepted, I'll open a ticket to test the `force_destroy` option and properly test it with time.

toninis (approved):

deployment/terraform/destroy.go
```diff
@@ -38,13 +36,13 @@ func (t *Terraform) Destroy() error {
...
 			}
-			if err := t.runAWSCommand(emptyBucketCtx, emptyS3BucketArgs, nil); err != nil {
-				emptyBucketErrCh <- fmt.Errorf("failed to run local cmd \"aws %s\": %w", strings.Join(emptyS3BucketArgs, " "), err)
-				return
-			}
+			// We intentionally ignore potential errors from this command,
+			// since it introduces spurious failures when the bucket is
+			// destroyed before the command finishes.
+			// See https://mattermost.atlassian.net/browse/MM-62075
+			_ = t.runAWSCommand(emptyBucketCtx, emptyS3BucketArgs, nil)
```
agnivade, line 43: Is it possible at all to know what error is returned when the bucket is destroyed before the command finishes, and ignore that exact error rather than all errors? If that is too complicated, this is fine as well. Just curious.

agarciamontoro: We could manually grep the returned error, looking for something like `failed to empty`. I don't know if that same string would show up in other errors, though. That would need some more testing.

agarciamontoro:
Oh, it turns out we already have`force_destroy`! https://github.com/mattermost/mattermost-load-test-ng/blob/deadfc95fd030523dacd68fe38fee8632cd2e7f8/deployment/terraform/assets/cluster.tf#L228 It was added way before the original timeout issue was filed and the `aws s3 rm` fix was added. cc/@toninis

toninis:
> Oh, it turns out we already have`force_destroy`!
> 
> https://github.com/mattermost/mattermost-load-test-ng/blob/deadfc95fd030523dacd68fe38fee8632cd2e7f8/deployment/terraform/assets/cluster.tf#L228
> 
> It was added way before the original timeout issue was filed and the `aws s3 rm` fix was added. cc/@toninis

That answers my question . Because the rm command should not fail with file does not exist unless another delete operation is in place. Perfect 🆗

agnivade (approved):

cpoile (approved):
Thanks for the detailed summary!

deployment/terraform/destroy.go
agnivade: I actually ran into this today. It seems like the error is NoSuchBucket:

```
info  [2024-12-06 14:26:13.837 +05:30] delete failed: s3://agnivalocallt.s3bucket/users/baw7cx384i8ppdka1h1zj3ntdr/profile.png An error occurred (NoSuchBucket) when calling the DeleteObject operation: The specified bucket does not exist caller="terraform/engine.go:71"
info  [2024-12-06 14:26:13.851 +05:30] fatal error: An error occurred (NoSuchBucket) when calling the ListObjectsV2 operation: The specified bucket does not exist caller="terraform/engine.go:71"

```

PR #868, opened 2024-12-11
[MM-61281] allow customizable public ip allocation

fmartingr:
#### Summary
Right now the terraform code forces public ip address on agents and proxy servers by default, adding a new configuration option to allow disabling that configuration attribute in the terraform configuration.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61281


agarciamontoro (approved):
Makes sense to me!

(I'm ever so slightly concerned that there will come a time when we have a config knob for each single Terraform field :joy:)

fmartingr:
> Makes sense to me!
> 
> (I'm ever so slightly concerned that there will come a time when we have a config knob for each single Terraform field 😂)

Yeah... what I would like is to group configuration by server type, instead of having that much of root level attributes. Should be easier to navigate, but is a big change at the moment.

agnivade (commented):
Man, this is becoming a maze of config options :P

Can we update deployer.sample.json and deployer.sample.toml?

Do we also need to update examples/ci/deployer.json every time we add a new field in deployer config @agarciamontoro ?

fmartingr:
> Can we update deployer.sample.json and deployer.sample.toml?

Oh boy I keep forgetting **something**. I thought I got it right this time 🙃

agnivade (approved):

agarciamontoro:
> Do we also need to update examples/ci/deployer.json every time we add a new field in deployer config @agarciamontoro ?

I have an item in the load-test tool release playbook to update it just before the release, no need to be on top of this every time.

PR #870, opened 2024-12-11
[MM-61280] disable init data if providing external database

fmartingr:
#### Summary
Avoid running the database initialization commands if providing an external database since the `IngestDump` only works for terraform created databases. External databases are usually already populated and ready to use as they come from backups.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61280

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -34,7 +34,7 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create SSH agent: %w", err)
 	}
 
-	initData := config.DBDumpURI == ""
+	initData := config.DBDumpURI == "" || config.ExternalDBSettings.DataSource != ""
 	if err = t.Create(extAgent, initData); err != nil {
 		return fmt.Errorf("failed to create terraform env: %w", err)
 	}
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -399,7 +399,7 @@ func (c *Config) IsValid() error {
 	}
 
 	if c.ExternalDBSettings.DataSource != "" && c.DBDumpURI != "" {
-		return errors.New("both ExternalDBSettings.DataSource and DbdumpURI are set, only one can be set")
+		return fmt.Errorf("both ExternalDBSettings.DataSource and DBDumpURI are set, only one can be set")
 	}
 
 	if err := c.validateElasticSearchConfig(); err != nil {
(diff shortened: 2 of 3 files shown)


cmd/ltctl/main.go
```diff
@@ -34,7 +34,7 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create SSH agent: %w", err)
 	}
 
-	initData := config.DBDumpURI == ""
+	initData := config.DBDumpURI == "" || config.ExternalDBSettings.DataSource != ""
```
streamer45, line 37: Wondering if this should be an invalid case at the config level. Does having both defined make sense?

fmartingr: No, because it will try to upload the dump to the database using the `TerraformDBSettings` credentials and getting the database URL from terraform's output `DBWriter`.

agarciamontoro: I don't understand this either. With this change, if both `DBDumpURI and `DataSource` are non-empty, `initData` is true, right? Shouldn't this be
```suggestion
	initData := config.DBDumpURI == "" && config.ExternalDBSettings.DataSource == ""
```

fmartingr: That looks better 👍

streamer45: I'd still welcome an addition to `func (c *Config) IsValid()` if having both set is a clear invalid case to avoid unexpected side effects. Not blocking though.

streamer45 (approved):

cmd/ltctl/main.go
fmartingr: https://github.com/mattermost/mattermost-load-test-ng/pull/870/commits/9aa44f4400579c82786440b7f67bae3b3f283823

deployment/config.go
```diff
@@ -398,6 +398,10 @@ func (c *Config) IsValid() error {
 		return fmt.Errorf("load-test download url is not in correct format: %q", c.LoadTestDownloadURL)
 	}
 
+	if c.ExternalDBSettings.DataSource != "" && c.DBDumpURI != "" {
+		return errors.New("both ExternalDBSettings.DataSource and DbdumpURI are set, only one can be set")
```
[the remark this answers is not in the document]
fmartingr: Fixed in https://github.com/mattermost/mattermost-load-test-ng/pull/870/commits/3fcdec404f23b180c3afcb3d4e22594e99542e4c

streamer45 (approved):
Thanks!

PR #871, opened 2024-12-11
[MM-61914] remove usage of unused errorsChan in ltkeycloak command

fmartingr:
#### Summary
Removed the `errorsChan` channel in the `ltkeycloak` command. The errors were being sent **and logged**, but the channel was never consumed, getting the command stuck if some errors appeared. In the feature branch we just increased the channel size to the number of users being imported but it's better to just remove it and allow regular logging to work this out.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61914


agnivade (approved):
Oops, nasty one.

PR #873, opened 2024-12-13
[MM-61910] refactor: log users by email

fmartingr:
#### Summary
Refactors the way the users log in to use login by email instead of by username. The problem with this was that when providing a txt file with a list of email/passwords the login flow used the username extracted from the email and that may not be accurate for non-synthetic dumps.

This pull request changes the login flow to use `GetLoginByEmail` instead of `GetLoginByUsername` but still returns the internal `user` object with the extracted username from the email.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61910

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -494,15 +494,13 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
 		password := split[1]
 		// Quick and dirty hack to extract username from email.
 		// This is not terribly important to be correct.
-		username := strings.Split(email, "@")[0]
-		username = strings.Replace(username, "+", "-", -1)
 		authService := userentity.AuthenticationTypeMattermost
 
 		// Check if the user has a custom authentication type. Custom authentication types are
 		// specified by prepending the email with the authentication type followed by a colon.
 		// Example: "openid:email1@xample.com"
-		if strings.Contains(username, ":") {
-			split := strings.Split(username, ":")
+		if strings.Contains(email, ":") {
+			split := strings.Split(email, ":")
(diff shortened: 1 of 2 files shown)


api/agent.go
```diff
@@ -494,15 +494,13 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
 		password := split[1]
 		// Quick and dirty hack to extract username from email.
 		// This is not terribly important to be correct.
```
agnivade, line 496: This needs to be moved lower as well.

api/agent.go
```diff
@@ -494,15 +494,13 @@ func getUserCredentials(usersFilePath string, _ *loadtest.Config) ([]user, error
...
-		username := strings.Split(email, "@")[0]
-		username = strings.Replace(username, "+", "-", -1)
```
agnivade, line 498: Why are we not using this anymore?

fmartingr: Because users are directly logging in via email, so we don't need to _guess_ the username from the email.

agnivade: I meant the `strings.Replace` line. The strings.Split line is there in L517. I actually don't recall why we needed to do the `strings.Replace`. But are you sure that we don't need it?

fmartingr: I am not, so let me put that back again.

fmartingr: https://github.com/mattermost/mattermost-load-test-ng/pull/873/commits/a39e6f9f6e2e8e5ed7b3910c7f94a28f2fde796c

agnivade (approved):
Thanks!

PR #874, opened 2024-12-13
MM-60336: Allow ability to create empty OS

agnivade:
https://mattermost.atlassian.net/browse/MM-60336


agnivade:
This works, but I had to set `EnableCloudwatchLogs` to false. Any idea what's up with that? I kept running into some Terraform error. 

Is https://github.com/mattermost/mattermost-load-test-ng/commit/5cf2fc2c74ab8e41557d46aa8f7f4e925ca88154 related somehow?

fmartingr (approved):

agnivade:
```
info  [2024-12-13 22:17:03.783 +05:30] ╷
│ Error: Missing resource instance key
│
│   on elasticsearch.tf line 130, in resource "aws_opensearch_domain" "es_server":
│  130:       cloudwatch_log_group_arn = aws_cloudwatch_log_group.es_log_group.arn
│
│ Because aws_cloudwatch_log_group.es_log_group has "count" set, its
│ attributes must be accessed on specific instances.
│
│ For example, to correlate with indices of a referring resource, use:
│     aws_cloudwatch_log_group.es_log_group[count.index]
╵ caller="terraform/engine.go:71"
Error: failed to create terraform env: exit status 1
```

agarciamontoro (approved):
Looks good to me, thanks :)

agarciamontoro:
@fmartingr, do we need `aws_cloudwatch_log_group.es_log_group[count.index]` in there, then? It makes sense

fmartingr:
> @fmartingr, do we need `aws_cloudwatch_log_group.es_log_group[count.index]` in there, then? It makes sense

Since we only create one I think it should work with `aws_cloudwatch_log_group.es_log_group[0]`

agarciamontoro:
@fmartingr, it seems that in some cases (not sure what the necessary conditions are), `count` is resolved after the check on the number of elements done. We saw that in https://github.com/mattermost/mattermost-load-test-ng/pull/819 as well. I'm not 100% sure this is the same error, though.

agarciamontoro:
@fmartingr, huh, then it must be another thing, sorry. But I'm still a bit worried about that error Agniva saw. Have you been able to reproduce it? Agniva's out this week.

agnivade:
I will retry with `[0]` and report back.

agnivade:
Setting `0` worked.

PR #878, opened 2025-01-13
[MM-61911] yace: use deployer config region

fmartingr:
#### Summary
Use the region in the deployer configuration inside the YACE template, instead of forcing it to `us-east-1`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61911


agnivade (approved):
Aha! I think this is why I wasn't getting any system metrics for my west coast cluster. Nice one :clap:

agarciamontoro (approved):

PR #880, opened 2025-01-14
MM-62533: Improvements for ExternalDBSettings

agnivade:
- We bump the postgres version to 14.9 as that is the version
after which "write forwarding" is available on a Global Aurora
Cluster.
- We also add the ClusterIdentifier for ExternalDBSettings
because we need to attach the DB security group to the DB cluster
like we do for a restored DB backup.

https://mattermost.atlassian.net/browse/MM-62533


deployment/terraform/create.go
```diff
@@ -153,18 +153,24 @@ func (t *Terraform) Create(extAgent *ssh.ExtAgent, initData bool) error {
...
+	if t.config.TerraformDBSettings.ClusterIdentifier != "" || t.config.ExternalDBSettings.ClusterIdentifier != "" {
 		if len(t.output.DBSecurityGroup) == 0 {
 			return errors.New("No DB security group created")
 		}
+		var identifier string
+		if t.config.TerraformDBSettings.ClusterIdentifier != "" {
+			identifier = t.config.TerraformDBSettings.ClusterIdentifier
+		} else {
+			identifier = t.config.ExternalDBSettings.ClusterIdentifier
+		}
```
streamer45, line 167: This is where I'd love it if Golang could implement some syntactic sugar.

streamer45 (approved):

agnivade:
> I understand that this applies only when using an AWS cluster, and if you use any other DB, the new ClusterIdentifier can simply be unset. Is that correct?

Correct.

PR #881, opened 2025-01-16
[MM-61277] Allow selecting between private and public connection types

fmartingr:
#### Summary
Adds a new `ConnectionType` deployer configuration option to allow the user to choose between using public/private IPs and DNS. This is necessary when a load test is deployed under private subnets.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61277


fmartingr:
@agnivade @agarciamontoro Can you guys take a look at this and let me know if the approach makes sense?

agarciamontoro:
Self-requested a review so I don't forget!

deployment/terraform/output.go
```diff
@@ -87,11 +87,40 @@ type Output struct {
...
 type Instance struct {
-	PrivateIP  string `json:"private_ip"`
-	PublicIP   string `json:"public_ip"`
-	PublicDNS  string `json:"public_dns"`
-	PrivateDNS string `json:"private_dns"`
-	Tags       Tags   `json:"tags"`
+	PrivateIP      string `json:"private_ip"`
+	PublicIP       string `json:"public_ip"`
+	PublicDNS      string `json:"public_dns"`
+	PrivateDNS     string `json:"private_dns"`
```
agnivade, line 93: Let's make these private fields to prevent mistakes of accessing them in the future.

deployment/terraform/output.go
```diff
@@ -87,11 +87,40 @@ type Output struct {
...
+	}
+	return i.GetConnectionIP()
+}
+
+func (i Instance) GetConnectionDNS() string {
+	if i.GetConnectionType() == "private" {
+		return i.PrivateDNS
+	}
+	return i.GetConnectionDNS()
 }
```
agarciamontoro, line 124: Are these self-referencing?

docs/config/deployer.md
```diff
@@ -621,3 +621,9 @@ Set the pprof block profile rate. This value applies to both agent and Mattermos
 *map[string]string*
 
 Optional map of key-value pairs, used to tag all deployed resources in AWS. Check [AWS documentation on tags](https://docs.aws.amazon.com/whitepapers/latest/tagging-best-practices/what-are-tags.html) for more information and best practices.
+
+## ConnectionType
+
+*string*
+
+Defines how instances should be accessed. Can be either "public" (default) or "private". When set to "private", instances will only be accessible through private IP addresses within the VPC.
```
agarciamontoro, line 629: These docs may be slightly ambiguous to someone that doesn't know anything about this. We should be more specific with "how instances should be accessed": who is accessing them? The user? The other instances?

agarciamontoro (commented):
Thanks for this! The approach makes sense to me. Left a couple of comments below.

deployment/terraform/output.go
agnivade: Oh yeah .. isn't this infinite recursion? :thinking:

deployment/terraform/output.go
agnivade: I realized that making these fields private would stop JSON unmarshal. In that case, I'd suggest to have a separate struct to unmarshal `InstanceToParse` and then create a new `Instance` by calling the Set methods.

fmartingr: I don't think this is necessary, there may be situations where we want to use the private or public ips directly. For example for intra-server communication we favor Private IPs over Public ones. One example is on the proxy IP in the `getServerURL` method. I'm currently reviewing usages and making sure I'm not breaking the services. Will re-request review once done and we can discuss in there, there is some tooling that I don't know how it would behave with this change.

agnivade: > For example for intra-server communication we favor Private IPs over Public ones. One example is on the proxy IP in the getServerURL method. I'm currently reviewing usages and making sure I'm not breaking the services. 

Ok, this is tricky. I'd prefer to have clear APIs that cannot be misused.

deployment/terraform/keycloak.go
```diff
@@ -356,7 +356,7 @@ func (t *Terraform) setupKeycloakAppConfig(sshc *ssh.Client, cfg *model.Config)
 		return fmt.Errorf("error uploading saml-idp.crt: %s - %w", out, err)
 	}
 
-	keycloakUrl := keycloakScheme + "://" + t.output.KeycloakServer.PublicDNS + ":8080"
+	keycloakUrl := keycloakScheme + "://" + t.output.KeycloakServer.PrivateDNS + ":8080"
```
agarciamontoro, line 359: This has to be always private?

deployment/terraform/output.go
```diff
@@ -87,11 +87,40 @@ type Output struct {
...
+	}
+	i.connectionType = connType
+}
+
+func (i Instance) GetConnectionType() string {
+	if i.connectionType == "" {
+		return "public"
+	}
+	return i.connectionType
+}
```
agarciamontoro, line 110: Should we set `connectionType` to `public` in `SetConnectionType` instead of to the empty string, and then simply return what we have in `GetConnectionType`?

deployment/terraform/output_test.go
```diff
@@ -0,0 +1,101 @@
...
+				PublicIP:   "203.0.113.4",
+				PrivateDNS: "ip-10-0-0-4.internal",
+				PublicDNS:  "ec2-203-0-113-4.compute-1.amazonaws.com",
+			},
+			connectionType: "",
+			expectedType:   "public",
+			expectedIP:     "203.0.113.4",
+			expectedDNS:    "ec2-203-0-113-4.compute-1.amazonaws.com",
+		},
+	}
```
agarciamontoro, line 70: Nit (disregard if you find it too nitty): Could we move the `Instance{}` block to a single definition and use the same in all tests? Just to make reading the test slightly easier.

docs/config/deployer.md
```diff
@@ -52,6 +51,16 @@ The ID of the subnet associated to the resources.
...
+## ConnectionType
+
+*string*
+
+Defines how instances should be accessed. Can be either `public` (default) or `private`. 
+
+This defines how the deployer connects to the servers and the connection IPs exposed to the user of the load-test tool in the different CLI utilities:
+- When set to `private` the deployer will connect to the servers using their private IP addresses, which means the **user must be connected to the same network as the servers via jumpshost/VPN**.
+- If set to `public` the deployer will connect using public IP addresses, which means the user must have access to the internet and **the servers must be on a VPC and subnet with internet access**.
+
```
agarciamontoro, line 63: Thank you so much for the thoroughness on the docs <3 There's a small typo on `jumpshost`:
```suggestion
## ConnectionType

*string*

Defines how instances should be accessed. Can be either `public` (default) or `private`. 

This defines how the deployer connects to the servers and the connection IPs exposed to the user of the load-test tool in the different CLI utilities:
- When set to `private` the deployer will connect to the servers using their private IP addresses, which means the **user must be connected to the same network as the servers via jump host/VPN**.
- If set to `public` the deployer will connect using public IP addresses, which means the user must have access to the internet and **the servers must be on a VPC and subnet with internet access**.

```

agarciamontoro (approved):
Looks good to me now! Just a few nits below and a typo. I haven't tested this, but if you want me to, I can give it a try before merging.

deployment/terraform/keycloak.go
fmartingr: This is the URL to be used from Mattermost to the Keycloak server, so it can be always private since the servers should reside in the same network.

deployment/terraform/output.go
fmartingr: If we end up in a code path that does not call `SetConnectionType` it will fail becasue it will be empty (as default string value). Made the change anyway because we should not end up in that situation anyways.

fmartingr:
> Looks good to me now! Just a few nits below and a typo. I haven't tested this, but if you want me to, I can give it a try before merging.

Hey, I made two small tests myself, but if you want to check it out just in case I missed something I'm not stopping you. The more 👀 the better.

agnivade (commented):
Given that we still have the fields exposed, and needed in cases for private communication, I am still a bit concerned that this highly increases the chances of a mistake happening in the future.

Anything we can do to prevent the user making this mistake, or atleast guide them in the right direction?

fmartingr:
> Given that we still have the fields exposed, and needed in cases for private communication, I am still a bit concerned that this highly increases the chances of a mistake happening in the future.
> 
> Anything we can do to prevent the user making this mistake, or atleast guide them in the right direction?

By that you mean adding a few docstrings to `Instance` or some change in the code?

agnivade:
I was thinking of some changes in the code. But if adding docs is the best we can do, then let's do that atleast.

fmartingr:
> I was thinking of some changes in the code. But if adding docs is the best we can do, then let's do that atleast.

Don't know how to properly solve that at the moment and I don't want this PR to become bigger, added some comments, please check them out: https://github.com/mattermost/mattermost-load-test-ng/pull/881/commits/2ba1e8f7695c93f5477ab0333703f632a0f3a551

We can discuss further modifications for this on a follow up effort.

agarciamontoro:
Tested this using `private` first, then re-running `deployment create` after having changed it to `public` and all looks good to me! Thanks, @fmartingr :)

agnivade (approved):
Looks good. Sorry this took longer than usual. I fear that the tool is becoming too hard to maintain because of the need to support so many options.

PR #883, opened 2025-01-24
Remove OpenTelemetry collector's debug exporter

agarciamontoro:
#### Summary
The debug exporter logs whatever payload the collector sends to Loki to the syslog, so it's easier to... well, debug the collector itself :) This was really useful when setting the whole logging system up, but it's no longer needed, and if there's any unrelated error (like an agent being down), `/var/log/syslog` can quickly end up consuming the whole disk (which is what motivated this change). The logs will still be sent to Loki if there's such an error, so it's not like we're removing useful info for debugging potential issues.

#### Ticket Link
--

### DIFF
--- a/deployment/terraform/strings.go
+++ b/deployment/terraform/strings.go
@@ -594,16 +594,12 @@ exporters:
     endpoint: "http://{{.MetricsIP}}:3100/otlp"
     tls:
       insecure: true
-  debug:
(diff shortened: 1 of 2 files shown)


agnivade (approved):

cpoile (approved):
👍

PR #885, opened 2025-02-04
Check for ExternalDBSettings while starting a load test

agnivade:
In case of ExternalDB there will be no cluster identifier
from a Terraform deployment.

### DIFF
--- a/deployment/terraform/db_operations.go
+++ b/deployment/terraform/db_operations.go
@@ -91,15 +91,27 @@ func (t *Terraform) DBStatus() (string, error) {
 		return "", err
 	}
 
-	if output.DBCluster.ClusterIdentifier == "" && t.config.ExternalDBSettings.ClusterIdentifier == "" {
-		return "", errors.New("DB cluster identifier not found")
+	// Neither Terraform DB exists, nor any External DB is used.
+	if output.DBCluster.ClusterIdentifier == "" && t.config.ExternalDBSettings.DataSource == "" {
+		return "", errors.New("DB cluster identifier not found or no external DB is used.")
 	}
 
+	// If an external non-AWS DB is used.
+	if t.config.ExternalDBSettings.DataSource != "" && t.config.ExternalDBSettings.ClusterIdentifier == "" {
+		return "available", nil
+	}
+
+	var identifier string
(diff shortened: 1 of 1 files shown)


agnivade:
LMK how it looks now. I am distinguishing the presence of a `ClusterIdentifier` with an AWS instance, and the presence of a `DataSource` with whether External DB is being used or not.

deployment/terraform/db_operations.go
```diff
@@ -91,15 +91,27 @@ func (t *Terraform) DBStatus() (string, error) {
 		return "", err
 	}
 
-	if output.DBCluster.ClusterIdentifier == "" {
-		return "", errors.New("DB cluster identifier not found")
+	// Neither Terraform DB exists, nor any External DB is used.
+	if output.DBCluster.ClusterIdentifier == "" && t.config.ExternalDBSettings.DataSource == "" {
+		return "", errors.New("DB cluster identifier not found or no external DB is used.")
```
agarciamontoro, line 96:
```suggestion
		return "", errors.New("DB cluster identifier not found and no external DB is used.")
```

deployment/terraform/db_operations.go
```diff
@@ -91,15 +91,27 @@ func (t *Terraform) DBStatus() (string, error) {
...
 	}
 
+	// If an external non-AWS DB is used.
+	if t.config.ExternalDBSettings.DataSource != "" && t.config.ExternalDBSettings.ClusterIdentifier == "" {
+		return "available", nil
+	}
```
agarciamontoro, line 102: Not for this PR, but: is there a way we can test the connection to the datasource here? It would need to be from one of the app nodes, but maybe it's worth doing it to make sure all is working as expected.

agarciamontoro (approved):
This looks better! I'm slightly concerned it's getting too complex, but that's a problem of the flexible configuration we have, of course, not of this PR. Thanks for this!

deployment/terraform/db_operations.go
agnivade: At which phase do you want to test it? This DBStatus method is purely geared towards AWS instances so that we can turn them off to save money. I think we could test this in the deployment phase, but it should automatically fail if the servers won't start.

cpoile (approved):

PR #886, opened 2025-02-06
Added missing keycloak setting

coltoneshaw:
#### Summary
Instance type was missing from the example config, just added it.


agnivade (commented):
We should also update the deployer.saml.toml file accordingly.

PR #891, opened 2025-02-21
feat: Add support for AWS role assumption in deployment configuration

toninis:
#### Summary

This change introduces an optional AWSRoleARN field to the deployment Config struct, allowing users to specify an IAM role to assume during AWS operations. The GetAWSConfig method in the Terraform utility has been updated to support role assumption using AWS SDK v2 credentials.

#### Ticket Link

https://mattermost.atlassian.net/browse/CLD-8926

### DIFF
--- a/deployment/config.go
+++ b/deployment/config.go
@@ -24,6 +24,8 @@ var esDomainNameRe = regexp.MustCompile(`^[a-z][a-z0-9\-]{2,27}$`)
 type Config struct {
 	// AWSProfile is the optional name of the AWS profile to use for all AWS commands
 	AWSProfile string `default:""`
+	// AWSRoleARN is the optional ARN of an IAM role to assume for AWS operations
+	AWSRoleARN string `default:""`
 	// AWSRegion is the region used to deploy all resources.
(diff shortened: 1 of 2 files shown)


agarciamontoro (commented):
Thank you! Can you add the `AWSRoleARN` we need to the `examples/config/ci/deployer.json` file so that the pipeline gets it (that's the set of files used in the pipeline)? I'll change https://github.com/mattermost/delivery-platform/pull/251 to use this branch to test it.

fmartingr (approved):
LGTM! Should we add this to the sample files as well?

agarciamontoro:
I was planning to investigate and fix the other error on the pipeline after this PR, but I didn't have time :( I have just started [a new pipeline](https://github.com/mattermost/delivery-platform/actions/runs/13573914297) with these changes just in case there's some flakiness and it's not a permissions error (it doesn't look like one, to be honest). I'm out tomorrow, so I'll let this working tonight and get back to it from Monday on.

agnivade:
@agarciamontoro - should we close this?

PR #892, opened 2025-02-21
Bump strings to v1.25.0 in master

agarciamontoro:
#### Summary
Just that :man_shrugging: Still waiting for https://mattermost.atlassian.net/browse/CLD-7570 :sob: 

#### Ticket Link
--


streamer45 (approved):

rahimrahman (approved):
Approving as an observer.  

What I'm understanding is it just uses the newer version that has already been compiled.

agarciamontoro:
We're basically updating the strings that point to the latest version (released this week) in sample files, test files and in default values. That's why I want a latest URL that always points to the latest release. That way we can hardcode that URL and forget about these annoying PRs.

PR #895, opened 2025-03-04
MM-63351: Use latest URL

agarciamontoro:
#### Summary
Now that https://mattermost.atlassian.net/browse/CLD-7570 is resolved, we have https://latest.mattermost.com/mattermost-load-test-ng-linux, which will always point to the latest release of the tool. This PR removes the hardcoded versions in the URLs in favour of this new URL, and removes the Makefile rule `prepare-release`, which is no longer needed.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63351

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,33 +100,7 @@ BRANCH_EXISTS=$(shell git rev-parse $(BRANCH_NAME) >/dev/null 2>&1; echo $$?)
 PR_URL=https://github.com/mattermost/mattermost-load-test-ng/compare/master...$(BRANCH_NAME)?quick_pull=1&labels=2:+Dev+Review
 CURR_BRANCH=$(shell git branch --show-current)
 
-prepare-release: ## Release step 1: Prepare the PR needed before releasing a new version, identified by the envvar NEXT_VER.
-ifndef NEXT_VER
-	@echo "Error: NEXT_VER must be defined"
-else
-ifeq ($(TAG_EXISTS), 0)
-	@echo "Error: tag ${NEXT_VER} already exists"
-else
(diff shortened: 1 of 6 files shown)


agarciamontoro:
I'm cherry-picking to `release-1.24` as soon as it's merged so that I can test the new, streamlined, release process with `v1.24.1` :nerd_face:

agarciamontoro:
> what would be the best procedure to figure out which version of the tool is deployed

Oh, that's a good question. We don't currently have a straightforward way to know the version that's deployed, as far as I know. We could add a new `/version` endpoint that returns the SHA of the commit used to build the binary and add a new command to query it. And log it as well when the server starts.

I'm curious: How would you figure it out now, without the changes in this PR?

streamer45:
> > what would be the best procedure to figure out which version of the tool is deployed
> 
> Oh, that's a good question. We don't currently have a straightforward way to know the version that's deployed, as far as I know. We could add a new `/version` endpoint that returns the SHA of the commit used to build the binary and add a new command to query it. And log it as well when the server starts.
> 
> I'm curious: How would you figure it out now, without the changes in this PR?

I'd be looking at the `deployer.json` used :p I'd expect us to log which version gets downloaded at the very least.

agarciamontoro:
> I'd be looking at the deployer.json used :p I'd expect us to log which version gets downloaded at the very least.

Yeah, we cannot even rely on the name in the agent machines (even now), since any reference to the original URL is gone. I just created [a ticket for this](https://mattermost.atlassian.net/browse/MM-63354), it shouldn't be too hard to write something that makes our lives easier.

agnivade (approved):

PR #896, opened 2025-03-04
Use journald app logs in opentelemetry collector

streamer45:
#### Summary

As the last comparison failure showed, it's rather annoying to figure out whether the app panicked. PR adds support to include these messages by using the full output of the `journalctl` unit rather the MM log file.

With this in place, we should be able to add a Grafana alert in case a panic actually happens, potentially creating a webhook in MM or something (a future improvement).

#### Screenshot

new_logs


agnivade (approved):
Great improvement!

deployment/terraform/create.go
```diff
@@ -1014,13 +1012,13 @@ func (t *Terraform) updateAppConfig(siteURL string, sshc *ssh.Client, jobServerE
...
 	cfg.LogSettings.EnableConsole = model.NewPointer(true)
-	cfg.LogSettings.ConsoleLevel = model.NewPointer("ERROR")
+	cfg.LogSettings.ConsoleLevel = model.NewPointer("WARN")
 	cfg.LogSettings.EnableFile = model.NewPointer(true)
 	cfg.LogSettings.FileLevel = model.NewPointer("WARN")
 	cfg.LogSettings.EnableSentry = model.NewPointer(false)
 
 	cfg.NotificationLogSettings.EnableConsole = model.NewPointer(true)
-	cfg.NotificationLogSettings.ConsoleLevel = model.NewPointer("ERROR")
+	cfg.NotificationLogSettings.ConsoleLevel = model.NewPointer("WARN")
```
agarciamontoro, line 1021: Yeah, good point :+1:

deployment/terraform/strings.go
```diff
@@ -549,7 +549,12 @@ RestartSec=1
 WantedBy=multi-user.target
 `
 
-const otelcolOperatorAppAgent = `
+const otelcolOperatorApp = `
+      - type: move
+        from: body.MESSAGE
+        to: body`
+
+const otelcolOperatorAgent = `
```
agarciamontoro, line 557: I'm kinda regretting this decision of writing this modularly instead of just writing a long string of text for each :joy:

agarciamontoro (approved):
Oooooh, this is nice! Is there a way to get the severity out of the journald entries? So that the log entries are coloured in Grafana in orange/red for warns/errors?

agarciamontoro:
> Judging by the screenshot above, I think this is handled out of the box for the usual MM logs

That's what this setting does (although truth be told, I didn't test it *without* it, so maybe it can autodetect?)
```yaml
        severity:
          parse_from: attributes.level`
```

But yeah, we could leverage a lot of features from the collector that we're ignoring right now. In any case, this is already a really good step forward, thanks for this!

PR #898, opened 2025-03-05
[AI assisted] MM-63354: Add logic to get build info

agarciamontoro:
#### Summary
This PR, done by a frustrated human (me) using aider for most of the tasks, does the following:
1. It adds a new package with a couple of functions to retrieve build information using [`debug.ReadBuildInfo`](https://pkg.go.dev/runtime/debug#ReadBuildInfo), which is a reliable way to get that info since, it seems, Go 1.18.
2. It adds a new `/version` endpoint to the server that returns that specific info.
3. It modified the starting log of the server to add that info as well.

The commits from aider were a total mess, so I re-did them all, making reviewable units. So feel free to review by commit :)

The experience of using aider has been a bit frustrating, although the integration test it came up with was a good start (I still needed to steer it in the right direction, but anyway, I kind of liked what it did there). ~~Not sure how it will behave in CI, so let's see~~ It passed, yay!

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63354


agarciamontoro:
Just tested that the new command works on a real deployment:
```
ubuntu@ip-172-27-137-145:~/alejandro/agm898$ go run ./cmd/ltctl deployment version
Agent Version Information:
Commit: 5c574322f045a1003cb1fcf90d12244b3c4518cd (modified)
Build Time: 2025-03-05 19:57:08
```

The logging does work, but we don't send the API logs (`ltapi.log`) to Loki, so we cannot see them in Grafana. However, they're there:

```
ubuntu@ip-172-27-132-40:~$ cat mattermost-load-test-ng/ltapi.log
{"timestamp":"2025-03-05 19:59:25.060 Z","level":"info","msg":"API server starting","caller":"ltapi/main.go:55","port":4000,"commit":"5c574322f045a1003cb1fcf90d12244b3c4518cd","buildTime":"2025-03-05T19:57:08Z","modified":true}
{"timestamp":"2025-03-05 20:11:56.544 Z","level":"info","msg":"API server starting","caller":"ltapi/main.go:55","port":4000,"commit":"5c574322f045a1003cb1fcf90d12244b3c4518cd","buildTime":"2025-03-05T19:57:08Z","modified":true}
{"timestamp":"2025-03-05 20:13:56.439 Z","level":"info","msg":"API server starting","caller":"ltapi/main.go:55","port":4000,"commit":"5c574322f045a1003cb1fcf90d12244b3c4518cd","buildTime":"2025-03-05T19:57:08Z","modified":true}
```

I've created [a ticket](https://mattermost.atlassian.net/browse/MM-63360) to track the Loki issue, it's something that was broken before this.

version/integration_test.go
```diff
@@ -0,0 +1,79 @@
...
+	// Create a temporary directory for our test binary
+	tempDir, err := os.MkdirTemp("", "ltapi-test")
+	require.NoError(t, err, "Failed to create temp directory")
+	defer os.RemoveAll(tempDir)
+
+	// Build the ltapi binary in the temp directory
+	apiServerPath := filepath.Join(tempDir, "ltapi")
+
+	// Build the binary directly with go build
+	buildCmd := exec.Command("go", "build", "-o", apiServerPath, "../cmd/ltapi")
```
agnivade, line 40: Is the need to compile the binary coming from the debug.ReadBuildInfo requirement? Is simply testing the version string not enough? It feels like by compiling we are testing `debug.ReadBuildInfo` itself, and not really the load-test code.

agarciamontoro: The code I added in version.go is quite simple, not more than a proxy for `debug.ReadBuildInfo`.  I added some unit tests for it, but I felt like they were not testing the real thing, and that's why I asked aider to add an integration test that tested the logic in a real environment. And yes, for `debug.ReadBuildInfo` to work, the binary needs to be built (and needs to be built in a git repo, actually).

agnivade (approved):

streamer45:
@agarciamontoro Is this a more idiomatic way of what we've been doing for ages through [injected `LDFLAGS`](https://github.com/mattermost/mattermost/blob/e549aaffb04c969a518e422be914a2addf91e737/server/Makefile#L104-L108) at compile time?

agarciamontoro:
@streamer45: for a basic use, I see this is cleaner than embedding values to variables at build time. The only downside is that the information stored in the binary is limited:
```
vcs: the version control system for the source tree where the build ran
vcs.revision: the revision identifier for the current commit or checkout
vcs.time: the modification time associated with vcs.revision, in RFC3339 format
vcs.modified: true or false indicating whether the source tree had local modifications
```

For this use-case, I'd say it's more than enough, but I'm open to discussing this.

streamer45:
@agarciamontoro Thanks. I am mostly trying to understand more as I notice we are introducing a slight inconsistency. Do we see a future where we use `debug.ReadBuildInfo` in MM server (or other services we ship)?

agarciamontoro:
@streamer45: It is indeed a slight inconsistency, but I wouldn't say it's too bad even if we don't end up using `ReadBuildInfo` anywhere else. I don't see us using this in MM in the near future, since we use more information there than the one provided by `ReadBuildInfo` (e.g., SHA for enterprise, flag for enterprise-ready).

The motivation of not using compile-time values here was mostly keeping things simple, as we don't need to add environment variables at compile time for this to work.

agnivade:
Indeed I see a future where we convert MM to be using debug.ReadBuildInfo instead. The compile time variables were used before this function was added in the standard library, and now there's no point in using the old method.

version/version.go
```diff
@@ -0,0 +1,69 @@
...
+	"fmt"
+	"runtime/debug"
+	"time"
+)
+
+// VersionInfo contains version information about the binary
+type VersionInfo struct {
+	Commit    string    `json:"commit"`
+	BuildTime time.Time `json:"build_time"`
+	Modified  bool      `json:"modified"`
```
streamer45, line 16: non-blocking: I'd include the Go runtime version since it can be somewhat important at times.

version/integration_test.go
```diff
@@ -0,0 +1,79 @@
...
+	require.NoError(t, err, "Failed to read response body")
+
+	// Unmarshal the response into a VersionInfo struct
+	var info version.VersionInfo
+	err = json.Unmarshal(body, &info)
+	require.NoError(t, err, "Failed to unmarshal response body")
+
+	// Verify the commit is a valid SHA (40 hex characters)
+	require.NotEqual(t, "unknown", info.Commit, "Commit should not be 'unknown'")
+	require.Regexp(t, regexp.MustCompile(`^[0-9a-f]{40}$`), info.Commit, "Commit should be a 40-character hex SHA")
```
streamer45, line 75: Surely we can figure out the current commit hash? :p

streamer45 (approved):
Nice!

version/version.go
agarciamontoro: https://github.com/mattermost/mattermost-load-test-ng/pull/898/commits/29bab705aac1af3d1e86a90befec6beb562b5f90

version/integration_test.go
agarciamontoro: https://github.com/mattermost/mattermost-load-test-ng/pull/898/commits/f8a72ccdebcf7ec17ad56736e6a85a3c5d88f6d8

PR #904, opened 2025-03-26
MM-63542: Add load-test coverage for new RHS loading pattern

agnivade:
Our existing viewThread implementation was incorrect. It used
to load pages of posts in a thread as a user scrolls up or down.

That is the ideal way, but our webapp actually loads the entire
thread. That is the first change done here.

Then, we apply the performance optimization where for subsequent
loads of a thread, we use the lastUpdateAt of a thread and pass
that to the API call.

https://mattermost.atlassian.net/browse/MM-63542

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/gliderlabs/ssh v0.1.1
 	github.com/grafana/alloy/syntax v0.1.0
 	github.com/graph-gophers/graphql-go v1.6.0
-	github.com/mattermost/mattermost/server/public v0.1.11-0.20250317155315-c95968c380be
+	github.com/mattermost/mattermost/server/public v0.1.11-0.20250319140854-a87c4ec9be66
 	github.com/mattermost/mattermost/server/v8 v8.0.0-20250317155315-c95968c380be
 	github.com/opensearch-project/opensearch-go/v4 v4.4.0
 	github.com/pelletier/go-toml/v2 v2.2.3
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/gliderlabs/ssh v0.1.1
 	github.com/grafana/alloy/syntax v0.1.0
 	github.com/graph-gophers/graphql-go v1.6.0
-	github.com/mattermost/mattermost/server/public v0.1.11-0.20250319140854-a87c4ec9be66
+	github.com/mattermost/mattermost/server/public v0.1.12-0.20250424064128-131cf039bbd5
 	github.com/mattermost/mattermost/server/v8 v8.0.0-20250317155315-c95968c380be
 	github.com/opensearch-project/opensearch-go/v4 v4.4.0
 	github.com/pelletier/go-toml/v2 v2.2.3
@@ -139,6 +139,7 @@ require (
 	github.com/yudai/golcs v0.0.0-20170316035057-ecda9a501e82 // indirect
 	github.com/yudai/pp v2.0.1+incompatible // indirect
 	github.com/yuin/goldmark v1.7.8 // indirect
+	golang.org/x/mod v0.24.0 // indirect
 	golang.org/x/net v0.37.0 // indirect
 	golang.org/x/sys v0.31.0 // indirect
 	golang.org/x/text v0.23.0 // indirect
--- a/loadtest/control/simulcontroller/actions.go
+++ b/loadtest/control/simulcontroller/actions.go
@@ -1729,63 +1729,81 @@ func (c *SimulController) viewThread(u user.User) control.UserActionResponse {
 		thread = *threads[0]
 	}
 
-	postIds, hasNext, err := u.GetPostThreadWithOpts(thread.PostId, "", model.GetPostsOptions{
+	// Get the lastUpdateAt
+	opts := model.GetPostsOptions{
 		CollapsedThreads: true,
 		Direction:        "down",
-		PerPage:          25,
-	})
-	if err != nil {
-		return control.UserActionResponse{Err: control.NewUserError(err)}
 	}
-	if len(postIds) == 0 {
-		return control.UserActionResponse{Info: "viewthread: no posts available to view in thread"}
-	}
-	newestPostId := postIds[len(postIds)-1]
-	newestPost, err := u.Store().Post(newestPostId)
-	if err != nil && !errors.Is(err, memstore.ErrPostNotFound) {
-		return control.UserActionResponse{Err: control.NewUserError(err)}
-	}
-	var newestCreateAt int64
-	if errors.Is(err, memstore.ErrPostNotFound) {
-		newestCreateAt = thread.Post.CreateAt
-	} else {
-		newestCreateAt = newestPost.CreateAt
+	if thread.LastUpdateAt != 0 {
+		opts.UpdatesOnly = true
+		opts.FromUpdateAt = thread.LastUpdateAt
 	}
 
-	// scrolling between 1 and 3 times
-	numScrolls := rand.Intn(3) + 1
-	for i := 0; i < numScrolls && hasNext; i++ {
-		postIds, hasNext, err = u.GetPostThreadWithOpts(thread.PostId, "", model.GetPostsOptions{
-			CollapsedThreads: true,
-			Direction:        "down",
-			PerPage:          25,
-			FromPost:         newestPostId,
-			FromCreateAt:     newestCreateAt,
-		})
+	// We always load the entire thread the first time.
+	// For subsequent loads, we only load the next updated posts.
+	hasNext := true
+	var allPostIds []string
+	for hasNext {
+		var postIds []string
+		postIds, hasNext, err = u.GetPostThreadWithOpts(thread.PostId, "", opts)
 		if err != nil {
+			// We are not checking for memstore.ErrPostNotFound because
+			// u.GetPostThreadWithOpts already stores the returned posts in the store.
+			// So the only case where an error might come is if list.Order doesn't
+			// have posts in list.Posts which is an error anyways.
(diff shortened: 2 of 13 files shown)


go.mod
```diff
@@ -33,7 +33,7 @@ require (
 	github.com/gliderlabs/ssh v0.1.1
 	github.com/grafana/alloy/syntax v0.1.0
 	github.com/graph-gophers/graphql-go v1.6.0
-	github.com/mattermost/mattermost/server/public v0.1.11-0.20250317155315-c95968c380be
+	github.com/mattermost/mattermost/server/public v0.1.11-0.20250319140854-a87c4ec9be66
```
agnivade, line 36: This points to the PR branch for now: https://github.com/mattermost/mattermost/pull/30486

loadtest/control/simulcontroller/actions.go
```diff
@@ -1729,72 +1729,90 @@ func (c *SimulController) viewThread(u user.User) control.UserActionResponse {
...
+			// We are not checking for memstore.ErrPostNotFound because
+			// u.GetPostThreadWithOpts already stores the returned posts in the store.
+			// So the only case where an error might come is if list.Order doesn't
+			// have posts in list.Posts which is an error anyways.
+			return control.UserActionResponse{Err: control.NewUserError(err)}
+		}
+		if latestUpdateAt < p.UpdateAt {
+			latestUpdateAt = p.UpdateAt
+		}
+	}
```
streamer45, line 1804: I think this could probably become its own util function.

loadtest/user/userentity/actions.go
```diff
@@ -1385,6 +1393,11 @@ func (ue *UserEntity) UpdateThreadFollow(teamId, threadId string, state bool) er
 	return err
 }
 
+// UpdateThreadLastUpdateAt updates the lastUpdateAt of the given thread
```
streamer45, line 1396: Do we really need this wrapper to simply set a store value? This layer was mostly meant for API calls (+ store as needed).

loadtest/store/memstore/store.go
```diff
@@ -1068,8 +1069,22 @@ func (s *MemStore) SetThread(thread *model.ThreadResponse) error {
 	return nil
 }
 
+// SetThreadLastUpdateAt stores the lastUpdateAt value of the given root post.
```
streamer45, line 1072: Do we need to call this in response to a websocket event to keep everything in sync like a real client would?

go.mod
streamer45: What would happen when this new logic runs against a MM server without those changes?

loadtest/control/simulcontroller/actions.go
```diff
@@ -1729,72 +1729,90 @@ func (c *SimulController) viewThread(u user.User) control.UserActionResponse {
...
-			FromPost:         newestPostId,
-			FromCreateAt:     newestCreateAt,
-		})
+	// We always load the entire thread the first time.
+	// For subsequent loads, we only load the next updated posts.
+	hasNext := true
+	var allPostIds []string
+	for hasNext {
+		var postIds []string
+		postIds, hasNext, err = u.GetPostThreadWithOpts(thread.PostId, "", opts)
```
streamer45, line 1748: How many posts would this fetch per call?

streamer45 (commented):
Changes make sense. Left a few comments.

Do we have a sense of how fetching the entire thread could impact our unbounded tests?

loadtest/user/userentity/actions.go
agnivade: I resorted to this because of our UserStore/MutableUserStore abstraction. The u.Store() returns store.Userstore which does not include writable methods. Do we want to change that?

go.mod
agnivade: We probably need to version-guard some of the logic. Good point on that.

loadtest/store/memstore/store.go
agnivade: Hmm .. even the real client does not do that :sweat_smile: Maybe it should. I'll look into it.

loadtest/control/simulcontroller/actions.go
agnivade: Oops, that's a mistake. I need to specify the perPage param here. By default, it doesn't set anything. The webapp sets 60. I need to match that.

loadtest/user/userentity/actions.go
streamer45: Oh man, that's something I had forgotten about. I guess I really tried to avoid mutations from the higher level. Maybe a small comment with this rationale is sufficient for now. If we start seeing a rising need for mutating in the controller then we may reconsider and expose the whole thing.

agnivade: Haha okay.

loadtest/store/memstore/store.go
agnivade: Ok, after giving it some more thought, keeping the websocket events in sync would be problematic. The reason is that this lastUpdateAt is a computed property on top of the redux store. So it's not directly stored in redux, but in src/reducers/views/threads.ts. And new reactions are also supposed to update this value.

Now, when a websocket event for a new reaction comes in - it just updates the redux state inside the reaction map, and the UI re-renders accordingly.

```ts
case PostTypes.RECEIVED_REACTION: {
        const reaction = action.data as Reaction;
        const nextReactions = {...(state[reaction.post_id] || {})};
        nextReactions[reaction.user_id + '-' + reaction.emoji_name] = reaction;

        return {
            ...state,
            [reaction.post_id]: nextReactions,
        };
    }
```

To update the lastUpdateAt from here would be difficult, because we need to find the threadId of the thread this post belongs to, and then find the latest updateAt of the thread, and then dispatch the action.

It seems like an anti-pattern to me to update a view state from a websocket event.

The only downside of not doing this would be that additional posts might be redundantly fetched. I'd say that it's still an improvement from before, where we would _always_ fetch the full thread.

WDYT?

streamer45: Thanks for looking into this. As long as we are not simulating less load than a real client, I think it's fine to keep it simple. Maybe we should just document (e.g., link to this ticket) somewhere so that we remember why we didn't go over this in more detail.

agnivade: >  As long as we are not simulating less load than a real client, I think it's fine to keep it simple.

We are not. This is the same logic as the client.

I will link this convo in the ticket.

streamer45 (approved):
Cool! Looking forward to some test results.

loadtest/control/simulcontroller/actions.go
```diff
@@ -1729,72 +1729,86 @@ func (c *SimulController) viewThread(u user.User) control.UserActionResponse {
...
-	for i := 0; i < numScrolls && hasNext; i++ {
-		postIds, hasNext, err = u.GetPostThreadWithOpts(thread.PostId, "", model.GetPostsOptions{
-			CollapsedThreads: true,
-			Direction:        "down",
-			PerPage:          25,
-			FromPost:         newestPostId,
-			FromCreateAt:     newestCreateAt,
-		})
+	// We always load the entire thread the first time.
+	// For subsequent loads, we only load the next updated posts.
```
agarciamontoro, line 1744: Where do the "subsequent loads" happen?

loadtest/store/memstore/store_test.go
```diff
@@ -653,14 +654,36 @@ func TestThreads(t *testing.T) {
...
+			},
+		}
+		err := s.SetThreads([]*store.ThreadResponseWrapped{thread})
+		require.NoError(t, err)
+		th, err := s.Thread(id)
+		require.NoError(t, err)
+		require.Equal(t, *thread, *th)
+	})
+
+	t.Run("SetThreads", func(t *testing.T) {
```
agarciamontoro, line 669: Should this be called `SetThreadLastUpdateAt`?

agarciamontoro (approved):
Looking good! Thank you for the changes, maintaining parity with the webapp is hard work. I left some comments, but all of them are non-blocking, so approving already :)

loadtest/control/simulcontroller/actions.go
agnivade: They will happen when `u.Store().RandomThread()` returns the same threadId again.

loadtest/control/simulcontroller/utils.go
```diff
@@ -227,3 +227,18 @@ func getPostFromEvent(ev *model.WebSocketEvent) (*model.Post, error) {
...
+		if err != nil {
+			return 0, err
+		}
+		if latestUpdateAt < p.UpdateAt {
+			latestUpdateAt = p.UpdateAt
+		}
+	}
+
+	return latestUpdateAt, nil
+}
```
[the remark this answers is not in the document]
agnivade: This is already checked here: https://github.com/mattermost/mattermost-load-test-ng/pull/904/files#diff-f4f5c4c3b7a3b3210ca6a5dba9b843c5529d94fe903a94a643fd9fb0e90cf0f5R1757. So it cannot be empty.

loadtest/user/userentity/utils.go
```diff
@@ -32,6 +32,9 @@ func postsMapToSlice(postsMap map[string]*model.Post) []*model.Post {
 func postListToSlice(list *model.PostList) []*model.Post {
 	posts := make([]*model.Post, len(list.Order))
 	for i, id := range list.Order {
+		// This assumes that list.Posts will always contain
+		// posts in list.Order. Need to improve. Otherwise, there is a
+		// possibility of nil pointers getting inserted.
```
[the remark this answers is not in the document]
agnivade: No, this is just a possibility I observed. In the server, the code is fine. So nothing to worry about yet.

PR #905, opened 2025-04-11
MM-63743: Add flag to maintain metrics instance

agarciamontoro:
#### Summary
This PR proposes a way to destroy everything but the metrics instance. If the `--do-not-destroy-metrics-instance` flag is passed, then:
1. The command runs `terraform rm <list of resources>`, where `<list of resources>` is a list containing the metrics instance and the security groups and rules that are needed.
2. Then the command runs `terraform destroy` as usual.

This has an important consequence: the resources that are passed to `rm` are completely ignored by Terraform from that moment on, meaning that the user will need to clean them up manually at some point. To make this very clear, I've added a confirmation step explaining the consequences of using the flag, which stops the command if the user does not answer `y`.

I'm open to discussions on how to approach this differently, but the idea of this flag is that it should only be used by people that know what they are doing, and it is of course a workaround while we find the time to tackle https://mattermost.atlassian.net/browse/MM-63249, which would be the proper solution.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63743

### DIFF
--- a/cmd/ltctl/main.go
+++ b/cmd/ltctl/main.go
@@ -60,7 +60,20 @@ func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to create terraform engine: %w", err)
 	}
 
-	return t.Destroy()
+	maintainMetrics, err := cmd.Flags().GetBool("do-not-destroy-metrics-instance")
+	if err != nil {
+		return fmt.Errorf("failed getting the --do-not-destroy-metrics-instance flag: %w", err)
+	}
+
+	resourcesToMaintain := []string{}
+	if maintainMetrics {
+		resourcesToMaintain = append(resourcesToMaintain, "aws_instance.metrics_server[0]")
+		resourcesToMaintain = append(resourcesToMaintain, "aws_security_group.metrics[0]")
+		resourcesToMaintain = append(resourcesToMaintain, "aws_security_group_rule.metrics-egress[0]")
+		resourcesToMaintain = append(resourcesToMaintain, "aws_security_group_rule.metrics-grafana[0]")
+	}
+
+	return t.Destroy(resourcesToMaintain...)
 }
 
 func RunInfoCmdF(cmd *cobra.Command, args []string) error {
@@ -272,11 +285,6 @@ func main() {
 			RunE:  RunCreateCmdF,
 		},
 		{
-			Use:   "destroy",
-			Short: "Destroy the current load-test deployment",
-			RunE:  RunDestroyCmdF,
-		},
-		{
 			Use:   "info",
 			Short: "Display information about the current load-test deployment",
 			RunE:  RunInfoCmdF,
@@ -307,6 +315,15 @@ func main() {
 			RunE:  RunVersionCmdF,
 		},
 	}
+
+	destroyCmd := &cobra.Command{
+		Use:     "destroy",
+		Short:   "Destroy the current load-test deployment",
+		PreRunE: checkDoNotDestroyMetricsInstanceFlag,
+		RunE:    RunDestroyCmdF,
+	}
+	destroyCmd.Flags().Bool("do-not-destroy-metrics-instance", false, "Destroy everything but the metrics instance, so you can still analyze already stored data.")
+	deploymentCommands = append(deploymentCommands, destroyCmd)
 
 	deploymentCmd.AddCommand(deploymentCommands...)
 	rootCmd.AddCommand(deploymentCmd)
@@ -485,10 +502,12 @@ func main() {
 	}
 
 	destroyComparisonCmd := &cobra.Command{
-		Use:   "destroy",
-		Short: "Destroy the current load-test comparison environment",
-		RunE:  DestroyComparisonCmdF,
+		Use:     "destroy",
+		Short:   "Destroy the current load-test comparison environment",
+		PreRunE: checkDoNotDestroyMetricsInstanceFlag,
+		RunE:    DestroyComparisonCmdF,
 	}
+	destroyComparisonCmd.Flags().Bool("do-not-destroy-metrics-instance", false, "Destroy everything but the metrics instance, so you can still analyze already stored data.")
 
 	comparisonCmd.AddCommand(runComparisonCmd, destroyComparisonCmd, collectComparisonCmd)
 	rootCmd.AddCommand(comparisonCmd)
--- a/cmd/ltctl/utils.go
+++ b/cmd/ltctl/utils.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+func askForConfirmation(prompt string) (bool, error) {
+	reader := bufio.NewReader(os.Stdin)
+	fmt.Print(prompt)
+	answer, err := reader.ReadString('\n')
+	if err != nil {
+		return false, fmt.Errorf("unable to read answer from user: %w", err)
+	}
+
+	return strings.ToLower(answer) != "y", nil
+}
+
+// checkDoNotDestroyMetricsInstanceFlag asks for confirmation if the
+// --do-not-destroy-metrics-instance flag is passed
+func checkDoNotDestroyMetricsInstanceFlag(cmd *cobra.Command, args []string) error {
+	maintainMetrics, err := cmd.Flags().GetBool("do-not-destroy-metrics-instance")
+	if err != nil {
+		return fmt.Errorf("failed getting the --do-not-destroy-metrics-instance flag: %w", err)
+	}
+
+	if maintainMetrics {
+		confirmed, err := askForConfirmation("CAUTION! The --do-not-destroy-metrics-instance flag will keep the metrics instance alive by removing it from Terraform state. This means that you will need to manually clean it up when you are done with it. Do you want to continue? [y/n] ")
+		if err != nil {
+			return err
+		}
+
+		if !confirmed {
+			return fmt.Errorf("Aborting this destroy. Remove the --do-not-destroy-metrics-instance flag if you want to destroy everything.")
+		}
+	}
+
+	return nil
+}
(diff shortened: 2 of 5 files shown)


cmd/ltctl/main.go
```diff
@@ -60,7 +60,20 @@ func RunDestroyCmdF(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("failed getting the --do-not-destroy-metrics-instance flag: %w", err)
+	}
+
+	resourcesToMaintain := []string{}
+	if maintainMetrics {
+		resourcesToMaintain = append(resourcesToMaintain, "aws_instance.metrics_server[0]")
+		resourcesToMaintain = append(resourcesToMaintain, "aws_security_group.metrics[0]")
+		resourcesToMaintain = append(resourcesToMaintain, "aws_security_group_rule.metrics-egress[0]")
+		resourcesToMaintain = append(resourcesToMaintain, "aws_security_group_rule.metrics-grafana[0]")
+	}
```
agnivade, line 74: This looks very brittle to me.

agnivade (commented):
If this is mainly applicable for core staff members, could an approach be simply to set agent/app/db instances to zero (or 1) and run `deployment create` again?

That should effectively achieve what we are trying to do here.

agarciamontoro:
> simply to set agent/app/db instances to zero

Oh, I had not thought about that, actually! We cannot set the number of agent instances to 0, but we can do it with the rest, and the cost of one agent instance is probably negligible. I like that! What we can do is hide all that complexity under the flag as well, so that users can still do `destroy ----do-not-destroy-metrics-instance`, and then just a simple `destroy` to get rid of what's left.

I'll try that and come back :)

agnivade:
Perfect :)

agarciamontoro:
I just tested your approach and it works flawlessly, @agnivade! Thanks for the suggestion :) Will push the changes once I clean up the code and make the same modifications to the `comparison destroy` command.

agarciamontoro:
Oh, one nice side-effect of this approach that I had not thought about is that now you can reuse the metrics instance! As everything is still under Terraform state, you can do
- `deployment create` at the start of your day
- `deployment destroy --do-not-destroy-metrics-instance` at the end of your day
- `deployment create` at the start of the next day

And it will reuse the same metrics instance, with no loss of the metrics gathered in the first deployment. This is cool.

deployment/terraform/create.go
```diff
@@ -307,8 +304,10 @@ func (t *Terraform) Create(extAgent *ssh.ExtAgent, initData bool) error {
 		}
 	}
 
-	if err := t.setupLoadtestAgents(extAgent, initData); err != nil {
-		return fmt.Errorf("error setting up loadtest agents: %w", err)
+	if t.config.AgentInstanceCount > 0 {
```
agnivade, line 307: Nice one

deployment/config.go
```diff
@@ -498,3 +500,19 @@ func ReadConfig(configFilePath string) (*Config, error) {
 
 	return &cfg, nil
 }
+
+// DestroyAllButMetrics overrides all created resources setting their instance
+// counts to 0 or their enable flags to false, so that everything is destroyed
+// in the next Create except for the metrics instance and related resources.
+// Note that this list should be kept up-to-date when new resources are added
+// to the Terraform files.
```
agnivade, line 508: This line is what bothers me. What I was thinking is that the user would simply edit the deployment config, and do a `deployment create`. But this seems like we are internally overwriting the config ourselves? Is there a way we can keep the config file also up-to-date?

agarciamontoro: I don't think we should update the config file, actually. That's defined by the user, and as a user I expect that to never change, only if *I* change it. What we do here is overwriting the config, yes, but for this command and this command only. That's why if you run `deployment create` once again, you would get the same deployment you had before: because your config file did not change. In my opinion, that's expected.

The user simply editing the config themselves would be a way to approach this, but I see that a lot more cumbersome for the user, and if you want to get to your previous deployment, you need to make a backup of your config files.

On the risk of this list getting outdated: yes, there is a risk, that's the biggest drawback here. However, think of the worst case scenario: we never update this list ever again. Even in that case, using this flag will still destroy the app nodes, the DB nodes, the agents, Elasticsearch, Keycloak, Redis and the job servers. It would keep any other future resource alive, yes, but I think removing the app nodes and the DB nodes would be actually good enough. 

Also, note that this flag is actually a temporary workaround until we get to migrate all metrics to the internal Grafana infrastructure. When we get there, this flag will hopefully become useless.

agnivade: Yep understood. Can we just log a warning message saying that the user has to manually update the deployer.json/toml themselves?

agnivade:
Also, we should update the PR message now that the code has changed :)

deployment/config.go
agarciamontoro: I'm not sure I understand the goal of such a message, could you give an example of what you have in mind?

agarciamontoro:
`comparison` command tested as well. We do need to modify the config before, so I reverted that commit. The only missing part is what [I shared in the team channel](https://hub.mattermost.com/pde/pl/ypn57bnwbfgzxkxa6ekesm7d7o).

deployment/config.go
agnivade: Something like: "Using this flag will preserve the metrics instance. So next time you run deployment create, it will reuse the instance instead of creating a new one."

agarciamontoro: Done in https://github.com/mattermost/mattermost-load-test-ng/pull/905/commits/6b0b7a875dc9c0037c37f9644b701f31eb38ec0e! The message will be :
```
CAUTION!
The --do-not-destroy-metrics-instance flag will keep the metrics instance and associated resources alive, which means that:
  1. If you recreate your deployment after this command finishes, the metrics instance will be reused instead of creating a new one.
  2. You will need to run `destroy` *without* this flag afterwards to clean everything up.
Do you want to continue? [y/n] 
```

agnivade (approved):
Thanks!

cmd/ltctl/main.go
```diff
@@ -49,12 +49,39 @@ func RunCreateCmdF(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
+func destroyAllButMetrics(config deployment.Config) error {
+	// Override all created resources so we destroy everything but the metrics
+	// instance.
+	config.DestroyAllButMetrics()
```
streamer45, line 55: Now that I see it in action, I find the naming a bit confusing. Could we revisit? Something like `MarkForDestroyAllButMetrics` perhaps.

deployment/terraform/create.go
```diff
@@ -154,11 +154,8 @@ func (t *Terraform) Create(extAgent *ssh.ExtAgent, initData bool) error {
 	}
 
 	// If we are restoring from a DB backup, or using an external database, then we need to hook up
-	// the security group to it.
-	if t.config.TerraformDBSettings.ClusterIdentifier != "" || t.config.ExternalDBSettings.ClusterIdentifier != "" {
-		if len(t.output.DBSecurityGroup) == 0 {
-			return errors.New("No DB security group created")
```
streamer45, line 160: Could you share some context on removing this?

streamer45 (commented):
Nice work, left some comments.

cmd/ltctl/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return fmt.Errorf("failed getting the --do-not-destroy-metrics-instance flag: %w", err)
+	}
+
+	if maintainMetrics {
+		msg := "CAUTION!\n"
+		msg += "The --do-not-destroy-metrics-instance flag will keep the metrics instance and associated resources alive, which means that:\n"
+		msg += "  1. If you recreate your deployment after this command finishes, the metrics instance will be reused instead of creating a new one.\n"
+		msg += "  2. You will need to run `destroy` *without* this flag afterwards to clean everything up.\n"
+		msg += "Do you want to continue? [y/n] "
```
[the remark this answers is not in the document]
agarciamontoro: I like it better because you don't have weird indentation in the code. Using backticks would be something like:
```
	if maintainMetrics {
		msg := `CAUTION!
The --do-not-destroy-metrics-instance flag will keep the metrics instance and associated resources alive, which means that:
  1. If you recreate your deployment after this command finishes, the metrics instance will be reused instead of creating a new one.
  2. You will need to run 'destroy' *without* this flag afterwards to clean everything up.
Do you want to continue? `
		confirmed, err := askForConfirmation(msg)
		if err != nil {
			return err
		}

		if !confirmed {
			return fmt.Errorf("Aborting this destroy. Remove the --do-not-destroy-metrics-instance flag if you want to destroy everything.")
		}
	}
```
It's just personal preference over code style.

cmd/ltctl/main.go
agarciamontoro: That's a good name, thanks! I couldn't come up with one that I liked, the `Mark` verb is key :ok_hand: Done in https://github.com/mattermost/mattermost-load-test-ng/pull/905/commits/1882968fec784af7dc4346de652c29f3c0a15969.

agarciamontoro:
Pausing this until I fix the issues with the security groups, which somehow become stale and so the metrics instance cannot get (some) metrics from the app, agent and proxy nodes

streamer45 (approved):

deployment/config.go
```diff
@@ -45,6 +45,8 @@ type Config struct {
 	AppInstanceType string `default:"c7i.xlarge" validate:"notempty"`
 	// IAM role to attach to the app servers
 	AppAttachIAMProfile string `default:""`
+	// EnableMetricsInstance enables deploying a metrics instance
+	EnableMetricsInstance bool `default:"true"`
```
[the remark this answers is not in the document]
agarciamontoro: I just tested it and it defaults to true, so nothing will change for the end user.

agarciamontoro:
The issue with the security groups can be reproduced without this PR, on current `master`, so these changes are not the culprit. Merging now, and tracking that issue on https://mattermost.atlassian.net/browse/MM-64198.

PR #907, opened 2025-04-22
MM-62559 Add custom profile attributes to loadtests

sbishel:
#### Summary
Adds data and API calls for Custom Profile Attributes

#### Ticket Link

  Fixes https://mattermost.atlassian.net/browse/MM-62559


cmd/ltagent/init.go
```diff
@@ -208,6 +209,17 @@ func genAdmins(config *loadtest.Config, userPrefix string) error {
...
+	if resp.Err != nil {
+		return resp.Err
+	}
+	if cpaEnabled {
+		err = loadtest.CreateCustomAttributeFields(sysadmin)
+		if err != nil {
+			return err
+		}
+	}
+
```
sbishel, line 222: This will create the Fields in the database. I'm expecting this code to run once and only once. Otherwise, I guess we need to add these to the database dump.

loadtest/control/actions.go
```diff
@@ -85,6 +85,17 @@ func Login(u user.User) UserActionResponse {
...
+	if resp.Err != nil {
+		return resp
+	}
+	if cpaEnabled {
+		err = u.GetCPAFields()
+		if err != nil {
+			return UserActionResponse{Err: NewUserError(err)}
+		}
+	}
+
```
sbishel, line 98: Whenever a user logs in, we need retrieve the CPA Fields.

loadtest/control/gencontroller/controller.go
```diff
@@ -103,6 +103,15 @@ func (c *GenController) Run() {
 		c.createTeam,
 	}
 
+	cpaEnabled, resp := control.CustomProfileAttributesEnabled(c.user)
+	if resp.Err != nil {
+		c.sendFailStatus("Failed to retreive CustomProfileAttributesEnabled")
+		return
+	}
+	if cpaEnabled {
+		initActions = append(initActions, c.createCPAValues)
```
sbishel, line 112: Creates the CPA Values for a new users. I'm expecting this code to run once per user creation.

loadtest/control/simulcontroller/actions.go
```diff
@@ -2007,6 +2007,22 @@ func (c *SimulController) openUserProfile(u user.User) control.UserActionRespons
...
+	if cpaEnabled {
+		// attempt to retrieve from store
+		attributes := u.Store().GetCPAValues(post.UserId)
+		if len(attributes) == 0 {
+			// Retrieve custom profile attribute values for the user.
+			if attributes, err = u.GetCPAValues(post.UserId); err != nil {
+				return control.UserActionResponse{Err: control.NewUserError(err)}
+			}
+		}
+	}
```
sbishel, line 2024: This works just as it does in the webapp. Attempts to retrieve from redux, if they haven't been retrieved, retrieve them from server.

sbishel (commented):
I haven't been able to try to run this yet. So almost guaranteed not to work. I needed to create data. The fields for the system and values for each user created. I'm not sure I did that properly, so I tried to comment what I was "trying" to do.

agarciamontoro (changes requested):
Sorry this took so long for me to review, @sbishel! The general approach is on point, I left a couple of comments below, and I have some more general comments here:

1. To generate data, updating the `gencontroller` is the way to go. But then we need to use it against a dump to create the data needed. For example, the DB dump with 100M posts was created using it (and then tweaking a couple of things here and there). As you point out in your comment, calling `CreateCustomAttributeFields` in the `genAdmins` function would call it everytime a test is executed. Maybe it's better to just generate the dump once.
2. I understand that right now, opening the user profile is the only place where the CPA values are retrieved. Is that right?

I lack some more context, so for next steps, I'm open to jump on a call to discuss specifics of the PR, and to try and run this for you / try to debug with you why you cannot deploy a cluster at all. Thoughts?

Thank you for the comments throughout the code, btw, it helped me understand your goals.

loadtest/control/gencontroller/controller.go
sbishel: `createCPAValues` creates various values.

In a real server this is either done manually by the user in UserPreferences. Or via LDAP/SAML sync. Or combination of both.

loadtest/user/userentity/custom_profile_attributes.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+// PatchUserCPA patches (or creates) a given users custom profile attributes.
+func (ue *UserEntity) PatchCPAValues(userId string, values map[string]json.RawMessage) error {
+	_, _, err := ue.client.PatchCPAValues(context.Background(), values)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
sbishel: No, PatchCPAValues only allow patching of the Session User's Values. I removed `userId` from the parameter list.

sbishel:
@agarciamontoro - I did update based on your comments, but still unsure of how to handle the db dump and update that. 
I'm open to finishing up with a call to answer questions. I'll schedule something this week.

agarciamontoro (changes requested):
## Data generation: gencontroller

As per our call, we decided that Scott was going to send some SQL statements to me to update our current dumps, so that we can add the fields and values to existing users there.

On top of that, we still need to update the `gencontroller` to make sure that whenever we create a new DB dump using it, it gets the CPA fields and values that we need.

Both for generating fields and values we can add regular actions to the `gencontroller`, since we will be using the global `gencontroller` state, which makes sure that each agent doesn't create more than the configured number of entities in `gencontroller.json` (this state is local to an agent process, but it is global to all users running there; this is known and taken into account when generating data).

So the first step here would be to add a new setting to the `gencontroller.Config` struct: `NumCPAFields`. Users can configure this setting each time they run the `gencontroller` to control how many fields will be created in the dump. As with every new config setting, we need to add it to `gencontroller.sample.json` and document it in `gencontroller.md`.

Then, we need to add a new `StateTargetCPAFields` constant in `gencontroller/state.go` and add it to the `targets` in the `init()` function.

Then, we can code our two new actions.

### Generating fields

For the fields, the new action should be run *before* the regular actions, so we can create a similar block to the one creating all channels. This action should be run in a `c.runActions` loop where the `done` function is something like
```go
func() bool { return st.get(StateTargetCPAFields) >= c.config.NumCPAFields }
```

Now, this new action should behave as follows:
1. First check if the user is an admin. If it's not, exit without an error.
2. If it is an admin, call `st.inc(StateTargetCPAFields, c.config.NumCPAFields)`.
3. Then, finally call `CreateCPAField` to create a single field, with a random `Name` and a random `Type` (or just hard-coded `text` there if it's the only type supported for now).

We could probably run this action once, but running it in a `runActions` loop make sure that even if we configure more fields than users we are simulating, then we will get them all created.

### Generating values

For the values, the logic should be similar: the new action should be run *before* the regular actions, but *after* the fields have been created. This time, we don't actually need to keep any state, and we don't even need to run the action in a loop, since we want to run this action once for each user, so that the values are created for the calling user just once. 

## Load test: simulcontroller

For `simulcontroller`, the current logic for getting the CPA values is perfect, but what I'm missing is new logic to patch them, mimicking the action where a real user edits their own values in their profile. We could add that as a new standalone action with its own frequency. You can get inspiration from any action in this controller, such as `updateCustomStatus`, for example.

-----

I think this covers all we discussed, but please ping me if I forgot anything or something here is not clear enough. Thanks for your patience!

docs/config/gencontroller.md
```diff
@@ -30,6 +30,12 @@ The number of private channels to be created.
 
 The number of public channels to be created.
 
+## NumCPAFields
+
+*int64*
+
+The number of Custom Profile Attribute fieldsto be created.
```
agarciamontoro, line 37: Typo:
```suggestion
The number of Custom Profile Attribute fields to be created.
```

loadtest/control/gencontroller/actions.go
```diff
@@ -71,6 +72,51 @@ func (c *GenController) createTeam(u user.User) (res control.UserActionResponse)
...
+
+	cpaField := &model.PropertyField{
+		Name: control.PickRandomWord() + "_" + control.PickRandomWord(),
+		Type: model.PropertyFieldTypeText,
+	}
+	// Only sysadmin can create CPA fields
+	field, err := c.sysadmin.CreateCPAField(cpaField)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
```
agarciamontoro, line 93: I would rather skip this action for non-admins here, I think it would be semantically more correct. So something like:
```suggestion
	// Only sysadmin can create CPA fields
	isSysAdmin, err := u.IsSysAdmin()
	if err != nil {
		return UserActionResponse{Err: NewUserError(err)}
	}
	
	if !isSysAdmin {
		return UserActionResponse{Warn: "not an admin user, unable to create a CPA field"}
	}

	field, err := c.sysadmin.CreateCPAField(cpaField)
	if err != nil {
		return control.UserActionResponse{Err: control.NewUserError(err)}
	}
```

loadtest/control/gencontroller/actions.go
```diff
@@ -71,6 +72,51 @@ func (c *GenController) createTeam(u user.User) (res control.UserActionResponse)
...
+
+	return control.UserActionResponse{Info: fmt.Sprintf("created CPA fields %s", field.ID)}
```
agarciamontoro, line 95: Typo:
```suggestion
	return control.UserActionResponse{Info: fmt.Sprintf("created CPA field %s", field.ID)}
```

loadtest/control/gencontroller/actions.go
```diff
@@ -71,6 +72,51 @@ func (c *GenController) createTeam(u user.User) (res control.UserActionResponse)
...
+}
+
+func (c *GenController) createCPAValues(u user.User) (res control.UserActionResponse) {
+	fields := u.Store().GetCPAFields()
+	values := make(map[string]json.RawMessage)
```
agarciamontoro, line 100: We should return with a warning if no fields returned by `GetCPAFields`

loadtest/control/gencontroller/actions.go
```diff
@@ -71,6 +72,51 @@ func (c *GenController) createTeam(u user.User) (res control.UserActionResponse)
...
+
+		values[field.ID] = value
+	}
+
+	err := u.PatchCPAValues(values)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("created randomized CPA values for user %s", u.Store().Id())}
```
agarciamontoro, line 117: Nit (everything is randomized in this codebase):
```suggestion
	return control.UserActionResponse{Info: fmt.Sprintf("created CPA values for user %s", u.Store().Id())}
```

loadtest/control/gencontroller/config.go
```diff
@@ -22,6 +22,8 @@ type Config struct {
 	NumChannelsPrivate int64 `default:"20" validate:"range:[0,]"`
 	// The target number of public channels to be created.
 	NumChannelsPublic int64 `default:"20" validate:"range:[0,]"`
+	// The target number of custom profile fields to be created.
+	NumCPAFields int64 `default:"2" validate:"range:[0,20]"`
```
agarciamontoro, line 26: Is the upper limit a hard requirement?

loadtest/control/gencontroller/controller.go
```diff
@@ -96,13 +96,37 @@ func (c *GenController) Run() {
...
+				run:        c.createCPAField,
+				frequency:  int(c.config.NumCPAFields),
+				idleTimeMs: 1000,
+			},
+		}
+		c.runActions(actions, func() bool {
+			return st.get(StateTargetCPAFields) >= c.config.NumCPAFields
+		})
+	}
+
```
agarciamontoro, line 118: I should have been more clear in my comment the other day. This should happen *before* the regular actions but *after* the user has logged in (we can do it right after `initActions`), otherwise they cannot perform any action.

loadtest/control/gencontroller/controller.go
```diff
@@ -96,13 +96,37 @@ func (c *GenController) Run() {
...
 	initActions := []control.UserAction{
 		control.SignUp,
 		c.login,
 		control.GetPreferences,
 		c.createTeam,
 	}
 
+	if cpaEnabled {
+		initActions = append(initActions, c.createCPAValues)
```
agarciamontoro, line 127: As this need to happen after `createCPAFields` and `createCPAFields` should happen after the `initActions`, I'm afraid we'll have to run two independent blocks. `createCPAValues` can be called once, though, no need to wrap it in `c.runActions`, I think.

loadtest/control/simulcontroller/actions.go
```diff
@@ -754,6 +755,28 @@ func (c *SimulController) updateSidebarCategory(u user.User) control.UserActionR
...
+	values := make(map[string]json.RawMessage)
+	values[field.ID] = value
+
+	err = u.PatchCPAValues(values)
+	if err != nil {
+		return control.UserActionResponse{Err: control.NewUserError(err)}
+	}
+
+	return control.UserActionResponse{Info: fmt.Sprintf("created randomized CPA values for user %s", u.Store().Id())}
```
agarciamontoro, line 777: Typo (and a nit to remove "randomized"):
```suggestion
	return control.UserActionResponse{Info: fmt.Sprintf("updated CPA field for user %s", u.Store().Id())}
```

loadtest/control/simulcontroller/controller.go
```diff
@@ -148,6 +148,12 @@ func getActionList(c *SimulController) []userAction {
 			frequency:        0.0040,
 			minServerVersion: control.MinSupportedVersion, // 5.26.0
 		},
+		{
+			name:             "UpdateCustomAttribute",
+			run:              c.updateCustomAttribute,
+			frequency:        0.0040,
+			minServerVersion: control.MinSupportedVersion, // 10.9.0
+		},
```
agarciamontoro, line 156: We need to specify the version here:
```suggestion
		{
			name:             "UpdateCustomAttribute",
			run:              c.updateCustomAttribute,
			frequency:        0.0040,
			minServerVersion: semver.MustParse("10.9.0"),
		},
```

loadtest/utils.go
```diff
@@ -77,6 +77,28 @@ func PromoteToAdmin(admin, userForPromotion *userentity.UserEntity) error {
...
+
+	for _, field := range fields {
+		_, err := admin.CreateCPAField(field)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
```
agarciamontoro, line 101: I think this is no longer needed, right?

loadtest/user/userentity/custom_profile_attributes.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/mattermost/mattermost/server/public/model"
+)
+
+// CreateCPAField creates a new custom profile attribute field.
+func (ue *UserEntity) CreateCPAField(field *model.PropertyField) (*model.PropertyField, error) {
+	new_field, _, err := ue.client.CreateCPAField(context.Background(), field)
+	if err != nil {
+		return nil, err
+	}
+	return new_field, nil
```
agarciamontoro, line 16: Nit:
```suggestion
	newField, _, err := ue.client.CreateCPAField(context.Background(), field)
	if err != nil {
		return nil, err
	}
	return newField, nil
```

loadtest/user/userentity/custom_profile_attributes.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// GetCPAFields retrieves all the cpa field values available.
```
agarciamontoro, line 19: Typo? Still not sure about terminology :joy: 
```suggestion
// GetCPAFields retrieves all the cpa fields available.
```

loadtest/user/userentity/custom_profile_attributes.go
```diff
@@ -0,0 +1,49 @@
...
+	return values, nil
+}
+
+// PatchCPAValues patches (or creates) a given user's custom profile attributes.
+func (ue *UserEntity) PatchCPAValues(values map[string]json.RawMessage) error {
+	_, _, err := ue.client.PatchCPAValues(context.Background(), values)
+	if err != nil {
+		return err
+	}
+	return nil
```
agarciamontoro, line 48: Nit:
```suggestion
	return err
```

agarciamontoro (changes requested):
Looking almost perfect! I left mostly nits and typos, which are non-blocking and you can ignore, but the comment about the order of the actions should be addressed.

The simulcontroller code is perfect!

Thank you for your infinite patience, @sbishel :) I will try to run this for you this week.

loadtest/control/gencontroller/actions.go
sbishel: I had written basically the same code you have above, based on your previous comments. Then as I continued investigating, it doesn't appear that user `u` is ever a sysadmin and a sysadmin user is passed separately. Which is why I used `c.sysadmin.CreateCPAField`.

Looking at this code
https://github.com/mattermost/mattermost-load-test-ng/blob/master/api/agent.go#L446-L455

Let me know, I'm probably misunderstanding something. Hard when you don't get to run it... :)

loadtest/control/gencontroller/config.go
sbishel: Yes, we currently only allow 20 custom profile properties.

sbishel:
@agarciamontoro - Biggest remaining issue is if sysadmin users a passed it. Let me know if I'm misreading the code, hard when you don't get to run it... :)

loadtest/control/gencontroller/actions.go
agarciamontoro: I believe this relies on the `UsersConfiguration.PercentOfUsersAreAdmin` setting. See https://github.com/mattermost/mattermost-load-test-ng/blob/1e61b93969d2a2c7705981c3d62b6d0c30899ecf/api/agent.go#L373-L376 and https://github.com/mattermost/mattermost-load-test-ng/blob/1e61b93969d2a2c7705981c3d62b6d0c30899ecf/api/agent.go#L392-L395 But I'm not 100% sure, so let me double-check this first.

agarciamontoro:
@sbishel, replied in the pending comment. Let's discuss in Mattermost.

loadtest/control/gencontroller/actions.go
sbishel: Ok...you sold me. Adding the sysadmin check. I missed that.

agnivade:
Hey @sbishel and @agarciamontoro. How is it going with this? LMK if I can help with anything.

agarciamontoro:
Hey @agnivade, @sbishel has successfully ran the code after many blockers in the way. Of course, we don't have CPA data in our dumps, so the first thing we'd need to do would be to extend the dumps with some CPA fields in the system (https://mattermost.atlassian.net/browse/MM-64420). I'll take a look at this PR again as soon as I can, and I just asked for your review to make sure there's another pair of eyes on this :)

sbishel:
Thanks @agarciamontoro for helping me get this finished up.

loadtest/control/simulcontroller/actions.go
```diff
@@ -2016,6 +2039,22 @@ func (c *SimulController) openUserProfile(u user.User) control.UserActionRespons
...
 	}
+
+	cpaEnabled, resp := control.CustomProfileAttributesEnabled(u)
+	if resp.Err != nil {
+		return resp
+	}
+	if cpaEnabled {
+		// attempt to retrieve from store
+		attributes := u.Store().GetCPAValues(post.UserId)
+		if len(attributes) == 0 {
```
agnivade, line 2050: Is this logic similar to the webapp where we try to fetch from redux store, and if not there, we make an API call?

loadtest/store/memstore/random.go
```diff
@@ -399,6 +400,18 @@ func (s *MemStore) RandomCategory(teamID string) (model.SidebarCategoryWithChann
 	return category, nil
 }
 
+func (s *MemStore) RandomProperty() *model.PropertyField {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
```
agnivade, line 405: This is a bug, because `s.GetCPAFields()` also gets a read lock. Nested mutexes (even if read-only) will create issues.

agnivade (changes requested):
Looks pretty good. There's just a small bug. Thanks @sbishel !

agarciamontoro (approved):
I was finally able to validate this works!

I used the gencontroller to create 15 CPA fields (I had to increase the `PercentOfUsersAreAdmin` value in config.json to 0.5 so that it could finish):

image

And then I increased the frequency of the `UpdateCustomAttribute` action in the simulcontroller to 1 to test the new action and it worked as expected:

image

image

agarciamontoro:
I've created https://mattermost.atlassian.net/browse/MM-64736 to review the frequency of the new action in a few weeks. We can merge this now!

agarciamontoro:
Thank you for your patience and your work, @sbishel!!!!!

PR #908, opened 2025-04-23
MM-63724: Remove excessive calls to GetChannelsForTeam

agarciamontoro:
#### Summary
This PR does a few things now, after Agniva's comments:
- It adds a new function to get *all* the channel memberships of *all* teams using the `/users/$user_id/channel_members` endpoint, named `GetAllChannelMembersForUser`. This is now called from `control.Reload`, right after `GetChannelsForUser`, mimicking [the behaviour in the webapp](https://github.com/mattermost/mattermost/blob/6ae0efd285bfd925e28ce86933c2a4b2170cc093/webapp/channels/src/components/team_controller/team_controller.tsx#L60-L66).
- It removes the calls to `GetChannelsForTeamForUser` both from `control.Login` and from `simulcontroller.loadTeam`. We can do this because we were already calling `GetChannelsForUser` in `control.Reload`: this call retrieves *all* channels from *all* teams.
- It removes the call to `GetChannelMembersForUser` from `simulcontroller.loadTeam` because we already have all channel memberships from the new `GetAllChannelMembersForUser` in `control.Reload`.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63724

### DIFF
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -696,6 +696,38 @@ func (ue *UserEntity) GetChannelMembers(channelId string, page, perPage int) err
 	return ue.store.SetChannelMembers(channelMembers)
 }
 
+// GetAllChannelMembersForUser gets all channel memberships for the
+// specified user regardless of the team the channels are part of
+func (ue *UserEntity) GetAllChannelMembersForUser(userId string) error {
+	perPage := 200
+	page := 0
+	for {
+		// Get perPage members from /users/$user_id/channel_members
+		membersWithTeamData, _, err := ue.client.GetChannelMembersWithTeamData(context.Background(), userId, page, perPage)
+		if err != nil {
+			return err
+		}
+
+		// Retrieve the embedded ChannelMember struct, we don't store the team data
+		plainMembers := make(model.ChannelMembers, len(membersWithTeamData))
+		for i, memberWithTeamData := range membersWithTeamData {
+			plainMembers[i] = memberWithTeamData.ChannelMember
+		}
+
+		// Set the slice of members we got
+		if err := ue.store.SetChannelMembers(plainMembers); err != nil {
+			return err
+		}
+
+		// Stop when we get to the last page
+		if len(membersWithTeamData) < 200 {
+			return nil
+		}
+
+		page++
+	}
+}
+
 // GetChannelMembersForUser gets the channel members for the specified user in
 // the specified team.
 func (ue *UserEntity) GetChannelMembersForUser(userId, teamId string) error {
--- a/loadtest/user/userentity/actions.go
+++ b/loadtest/user/userentity/actions.go
@@ -16,6 +16,7 @@ import (
 	"regexp"
 	"strconv"
 
+	"github.com/blang/semver"
 	"github.com/graph-gophers/graphql-go"
 	"github.com/mattermost/mattermost-load-test-ng/loadtest/user"
 	"github.com/mattermost/mattermost/server/public/model"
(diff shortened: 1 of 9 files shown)


agarciamontoro:
I think that happens when we call `GetPublicChannelsForTeam` in the `joinChannel` action, but this is actually independent of this change.

In this PR we're removing an additional call to UserEntity.GetChannelsForTeam, which is an alias of UserEntity.GetChannelsForTeamForUser using the current user ID instead of explicitly passing one. Both end up calling `Client4.GetChannelsForTeamForUser`

Not sure if this clarifies things.

agarciamontoro:
Oh, wait, the one that loads all channels from all teams for a user is `/users/me/channels`, I had overlooked that one :man_facepalming: But it seems we call it in `Reload`: https://github.com/mattermost/mattermost-load-test-ng/blob/92253bfd2a4157fe86ca5595972f4b5ed06610c1/loadtest/control/actions.go#L833 
The second one, the one that loads all channel members from all teams for a user, is `/users/$user_id/channel_members`. There is no call to that one in the load-test tool as far as I can see, only to `/teams/$teamId/channels/members` (`UserEntity.GetChannelMembersForUser`), which seems to be the corresponding one *for a single* team. That is also called from `Reload`, but [just once for the current team](https://github.com/mattermost/mattermost-load-test-ng/blob/92253bfd2a4157fe86ca5595972f4b5ed06610c1/loadtest/control/actions.go#L827). It's called in other places as well, but again for a single team.

In short: I think the changes here are still good, but we still need to add that call to `/users/$user_id/channel_members`.

Thoughts, @agnivade?

agnivade:
Yes, we'd need to add `/users/$user_id/channel_members`. But we also need to remove 

```
} else {
		if _, err := u.GetChannelsForTeamForUser(team.Id, u.Store().Id(), true); err != nil {
		}

		if err := u.GetChannelMembersForUser(u.Store().Id(), team.Id); err != nil {
		}
	}
``` 

from switchTeam.

agarciamontoro:
Ah, that would be what you removed in your PR, right? Ok, I think that would be enough*.

* famous last words

loadtest/user/userentity/actions.go
```diff
@@ -696,6 +696,38 @@ func (ue *UserEntity) GetChannelMembers(channelId string, page, perPage int) err
 	return ue.store.SetChannelMembers(channelMembers)
 }
 
+// GetAllChannelMembersForUser gets all channel memberships for the
+// specified user regardless of the team the channels are part of
+func (ue *UserEntity) GetAllChannelMembersForUser(userId string) error {
+	perPage := 200
+	page := 0
```
agnivade, line 703: Apologies for some more work. But I actually improved the API by streaming all data if you set page=-1: https://github.com/mattermost/mattermost/pull/30771/files#diff-f49fddb8b3d75c63925d36a97a8d5389337826b11ef544950f4898202d142489R6677.

So we have 2 options:
1. Merge this PR now and revert the changes that you made.
2. Or wait till my PR is merged, and update the API to use -1 as page number.

I am fine with either. Up to you.

agarciamontoro: Oh, nice! No need to be sorry, that's actually a great improvement, thanks! I'll go with 2., will wait until your PR is merged and I'll make the change in the meantime.

agarciamontoro: Actually, we'll need both implementations to support older server versions.

agarciamontoro: Check https://github.com/mattermost/mattermost-load-test-ng/pull/908/commits/ffc0f7fa1c6a303161bdc7869bdfa7fdfafaf9a7 As the client itself has not changed, there is no need to wait for your PR to be merged, assuming it will be released in 10.9.0.

agarciamontoro:
Note that this is a very large deployment (same specs as your 200k), the one I was not able to get above 60k. I expect something similar to happen here.

agnivade (approved):
Nice work! Excited to see the results.

agnivade:
Yep, definitely an improvement. Which is great!

loadtest/user/userentity/actions.go
```diff
@@ -697,6 +698,60 @@ func (ue *UserEntity) GetChannelMembers(channelId string, page, perPage int) err
...
 }
 
+// GetAllChannelMembersForUser gets all channel memberships for the
+// specified user regardless of the team the channels are part of
+func (ue *UserEntity) GetAllChannelMembersForUser(userId string) error {
+	serverVersion := ue.Store().ServerVersion()
+	// Versions 10.9.0 and later support streaming all the members by setting
+	// page to -1, so there is no need to paginate
+	if serverVersion.GTE(semver.MustParse("10.9.0")) {
+		membersWithTeamData, _, err := ue.client.GetChannelMembersWithTeamData(context.Background(), userId, -1, 0)
```
streamer45, line 708: I suppose that from a client-side perspective, we aren't truly streaming anything since this will return the whole list?

Perhaps there could be a way to leverage the new iterators concept in Go 1.23+ and actually stream them through.

loadtest/user/userentity/actions.go
```diff
@@ -697,6 +698,60 @@ func (ue *UserEntity) GetChannelMembers(channelId string, page, perPage int) err
...
+			plainMembers[i] = memberWithTeamData.ChannelMember
+		}
+
+		// Set the slice of members we got
+		if err := ue.store.SetChannelMembers(plainMembers); err != nil {
+			return err
+		}
+
+		// Stop when we get to the last page
+		if len(membersWithTeamData) < 200 {
```
streamer45, line 746: We should use `perPage` here. Even better, leverage something like https://github.com/mattermost/mattermost/pull/29165#discussion_r1861323917

streamer45 (approved):
Nice work!

loadtest/user/userentity/actions.go
agnivade: The client needs the entire data all at once. So there isn't really an opportunity to stream inside the client.

loadtest/user/userentity/actions.go
agarciamontoro: Whoops, that's true, fixed! Will defer that nice `pager` for another PR.

agnivade:
Ah @agarciamontoro, this PR still wouldn't work without my branch because the client changes required here: https://github.com/mattermost/mattermost/pull/30771/files#diff-85d1aeff45b42e9c7d3b165d2bffbc341bc060acd73ea90f88e97278bcb77539 are not merged yet. Not a big deal, I will test it off a separate branch first to verify, and then merge the PR.

After that, we need to bump the model version in go.mod file.

PR #909, opened 2025-04-30
MM-63827: Bump Grafana version and demote anon role

agarciamontoro:
#### Summary
This PR partially tackles https://mattermost.atlassian.net/browse/MM-63827 by:
1. Bumping Grafana version
2. Demoting the role of anonymous access to just Viewer

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63827

PR #911, opened 2025-05-07
MM-63827: Generate random password for Grafana admin user

agarciamontoro:
#### Summary
Second part of https://mattermost.atlassian.net/browse/MM-63827, in addition to #909. This PR generates a random password for Grafana's `admin` user.

The less pretty part is that we need to store the password locally to persist it between calls to `ltctl` commands, so that we can show it on `deployment info`. As this flow is only needed when deploying to Terraform, I've created a new `GeneratedValues` struct that stores values generated during deployment that are not user-specified (only this Grafana admin pass for now). I added a field of that type to the `Terraform` struct, and then I persist that struct to disk in a new file inside `TerraformStateDir`, which is already a directory that `ltctl` controls and that stores Terraform-related state.

I first though about implementing a full reader/writer for this struct, so that the caller simply needs to do something like Get/Set, but I ended up coding only the minimal interface I need ( (in `generated_values.go`) to get from and write to disk, so for example, the caller is required to call `PersistGeneratedValues` if needed. If we see that we need something more advanced in the future, we can generalize it from here, but for now, I'd say that this is enough.

The tests are mostly written by aider.

Output when `deployment create` finishes (sanitized so that it is not shown in CI and similar)
```
Grafana URL: http://35.153.175.63:3000
Grafana credentials:
  - User: admin
  - Pass: ********
```
Output of `deployment info` (explicit so that the user can get the credentials they need)
```
Grafana URL: http://35.153.175.63:3000
Grafana credentials:
  - User: admin
  - Pass: GZLqB1rPWvlyIgq20DVZDvJiM0nrH8zH
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63827

### DIFF
--- a/deployment/terraform/metrics.go
+++ b/deployment/terraform/metrics.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math/rand"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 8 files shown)


deployment/terraform/metrics.go
```diff
@@ -312,6 +312,24 @@ func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent) error {
...
 
+	// Change Grafana admin password
+	// No need to be cryptographically secure here, we just need a new password
+	passLen := 32
+	chars := []rune("abcdefghijklmnopqrstuvwxyz" + "ABCDEFGHIJKLMNOPQRSTUVWXYZ" + "0123456789")
+	s := make([]rune, passLen)
+	for j := 0; j < passLen; j++ {
+		s[j] = chars[rand.Intn(len(chars))]
+	}
+	password := string(s)
```
agnivade, line 323: This could be refactored to a util

agnivade (commented):
Just one small comment. Looks great otherwise

deployment/terraform/metrics.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-load-test-ng/pull/911/commits/2456d87b04074229e42b5be7c4d3f548852c0640

agnivade (approved):

cpoile (approved):
Thank you for the tests!

PR #912, opened 2025-05-07
MM-64198: Fix unreachable targets from the metrics instance

agarciamontoro:
#### Summary
To fix the issue at hand, I first thought that our use of `ingress` and `egress` blocks inside the `aws_security_group` resources was the culprit. This assumption was based on this note from [the Terraform docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/security_group):
image

So I did that: I changed all those blocks by stand-alone `aws_vpc_security_group_{ingress,egress}_rule` resources. That's what the first commit does. But I tested it and... everything failed in the same way.

Digging a bit more, I finally understood what the actual problem was: the rules that we had were not being applied at all! This happens because the metrics instance was reaching the targets (app, agent and proxy instances) via their public IPs. This started happening when we merged #881, since we incorrectly set the IPs in the metrics' `hosts` file to the ones returned by `GetConnectionIP`, which by default returns the public ones (unless the `ConnectionType` setting is set to `private`). Thus, the *actual* fix for the issue is what's changed in the second commit.

image

I do want to merge both commits, though, since the first one still improves the code by following the best practices recommended by the Terraform documentation.

Also, once this PR is merged, I'll split the `cluster.tf` file in separate files (probably per type of node: app, agent, proxy...), since it's becoming large and hard to follow.

There's also one thing left to fix, which is that Keycloak doesn't seem to export metrics on `:8080/metrics`. I'll investigate this separately.
image

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64198


agarciamontoro:
Not required to review, @fmartingr, but take a look if you want :)

deployment/terraform/metrics.go
```diff
@@ -124,19 +124,19 @@ func (t *Terraform) setupMetrics(extAgent *ssh.ExtAgent) error {
 		mmTargets = append(mmTargets, fmt.Sprintf("%s:8067", host))
 		nodeTargets = append(nodeTargets, fmt.Sprintf("%s:9100", host))
 		netpeekTargets = append(netpeekTargets, fmt.Sprintf("%s:9045", host))
-		hosts += fmt.Sprintf("%s %s\n", val.GetConnectionIP(), host)
```
agnivade, line 127: I had a feeling that things might go wrong when we added the GetConnectionIP method. Very nice work on this!

agnivade (approved):

cpoile (approved):
🎉

PR #914, opened 2025-05-12
MM-62365: Fix errors in submitPerformanceReport

agnivade:
There were various bugs in the implmentation logic.
1. Start and End are ms, not s.
2. The Version field is compulsory to be set in the
report struct.
3. The timestamp field is unused entirely in the server.
So we remove it here and will remove from server
as well.

https://mattermost.atlassian.net/browse/MM-62365


isacikgoz (approved):

agarciamontoro (approved):
Thank you, @agnivade!

PR #916, opened 2025-05-16
Fix validation of sample JSON files

streamer45:
#### Summary

Looks like `deployer.sample.json` wasn't valid JSON (due to an extra comma). I added a script to run as part of `check-style` to avoid merging invalid JSON files in the future.


agnivade (approved):

agarciamontoro (approved):
If JSON was a sensible format accepting trailing commas...

Thank you for the script, much needed!

PR #917, opened 2025-05-19
Add lambda function and invoker

agnivade:
I am adding the POC of the lambda function I have been working on.
It is at a reasonable state right now, but I am not making it part
of ltctl just about yet.


agnivade:
Adding @saturninoabril as well because this consists of UI testing.

agarciamontoro (approved):
Great work! Looking really good, thanks for adding it early to the repo :) As this is still in development, I have not gone into much detail in the review. Even the comments I did still write, those are all nits, so feel free to ignore them, at this stage it's better to iterate fast.

I do have a general question, though: Have you compared this against the feature to inject actions to the controllers? I think that some of the goals you had in mind can be achieved using that as well.

lambda/README.md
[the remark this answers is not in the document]
agnivade: Sure, I will add that in the README of the invoker.

lambda/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Required Libraries and Dependencies
+
+- `puppeteer-core`: Provides the API to control Chromium.
+- The `@sparticuz/chromium` library is provided as a layer.
+- Node.js 14 or higher recommended
+
+## Installation
+
+First, use this layer in your function : `arn:aws:lambda:us-east-1:764866452798:layer:chrome-aws-lambda:50`
```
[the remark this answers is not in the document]
agnivade: Yes, if you open the lambda function UI then it becomes clear. "Layers" in a lambda function is similar to docker layers, where you just store some data and copy paste it every time the function starts. 

Try visiting the `agnivaTest` lambda function in AWS and open the layers tab.

cmd/invoker/main.go
```diff
@@ -0,0 +1,228 @@
...
+		fmt.Printf("Error parsing Lambda response: %v\n", err)
+		return
+	}
+
+	// Parse the response body
+	var responseBody ResponseBody
+	if err := json.Unmarshal([]byte(response.Body), &responseBody); err != nil {
+		fmt.Printf("Error parsing response body (%q): %v\n", response.Body, err)
+		return
+	}
```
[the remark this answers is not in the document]
agnivade: No it's just a string.

lambda/index.js
```diff
@@ -0,0 +1,148 @@
+const puppeteer = require('puppeteer-core');
```
streamer45, line 1: It would be nice to explore whether using `playwright` is an option these days. That would unify with our e2e pipelines and allow us to do a lot of interesting things.

cmd/invoker/main.go
```diff
@@ -0,0 +1,228 @@
...
+	joinTeam := flag.String("join_team", "", "Team name to join")
+	functionName := flag.String("function", "mattermost-login", "Name of the Lambda function")
+	region := flag.String("region", "us-east-1", "AWS region")
+	outputFile := flag.String("output", "", "File to save screenshot (if debug is enabled)")
+	count := flag.Int("count", 1, "Number of concurrent Lambda invocations")
+
+	flag.Parse()
+
+	// Validate required parameters
+	if *username == "" || *password == "" || *url == "" {
```
streamer45, line 61: Any reason for using pointers?

cmd/invoker/main.go
```diff
@@ -0,0 +1,228 @@
...
+		fmt.Println("Error: username, password, and url are required")
+		flag.Usage()
+		return
```
streamer45, line 64: We should ideally use `os.Exit` with a non-zero code.

cmd/invoker/main.go
```diff
@@ -0,0 +1,228 @@
...
+	_, err := fmt.Sscanf(parts[len(parts)-1], "%d", &startNum)
+	if err != nil {
+		return "", 0, fmt.Errorf("invalid number suffix in username: %v", err)
+	}
+
+	return prefix, startNum, nil
+}
+
+// invokeSync performs a synchronous Lambda invocation
+func invokeSync(client *lambda.Client, functionName string, event LambdaEvent, outputFile string) {
```
streamer45, line 113: I feel this should return a proper `error`.

lambda/README.md
```diff
@@ -0,0 +1,31 @@
...
+- Websocket reconnect.
+- Large number of users joining a team in a short time.
+- Large number of users joining a channel in a short time.
+- Possible others.
+
+## Required Libraries and Dependencies
+
+- `puppeteer-core`: Provides the API to control Chromium.
+- The `@sparticuz/chromium` library is provided as a layer.
+- Node.js 14 or higher recommended
```
streamer45, line 13: I'd probably include an `.nvmrc` file to align with our main repos (currently at `20.11`). To note, anything below v20 is pas end-of-life I believe.

cmd/invoker/main.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Launch goroutines for each invocation
+	for i := 0; i < count; i++ {
+		wg.Add(1)
+		go func(index int, event LambdaEvent) {
+			defer wg.Done()
+
+			// Add random sleep proportional to count
+			// The higher the count, the more spread out the starts will be
+			maxSleepMs := count * 200 // 200ms per count as a scaling factor
```
streamer45, line 193: Do you see this as something configurable down the line?

lambda/README.md
```diff
@@ -0,0 +1,31 @@
...
+npm install
+Zip the entire project: zip -r deployment.zip index.js package.json node_modules
+Upload the code: aws lambda update-function-code --function-name <functionName> --zip-file fileb://deployment.zip
+Invoke the function: aws lambda invoke --function-name <functionName> --cli-binary-format raw-in-base64-out --payload '{"username": "<username>", "password": "<password>", "url": <url>}' response.json
+```
+
+## Testing locally:
+
+```
+node local-test.js --url "<>" -u <> -p <> --delay 5000
```
streamer45, line 30: Nice, super useful!

lambda/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Installation
+
+First, use this layer in your function : `arn:aws:lambda:us-east-1:764866452798:layer:chrome-aws-lambda:50`
+
+Then,
+```bash
+npm install
+Zip the entire project: zip -r deployment.zip index.js package.json node_modules
+Upload the code: aws lambda update-function-code --function-name <functionName> --zip-file fileb://deployment.zip
```
streamer45, line 23: Is `fileb://` a thing or a typo?

streamer45 (approved):
Nice! Nothing blocking but a few comments/questions.

lambda/README.md
agnivade: It is indeed a thing. An AWS thing.

cmd/invoker/main.go
agnivade: Yeah definitely. I left it like this for now, but down the line we'll need to do something better.

lambda/index.js
agnivade: Yep definitely something to do for future.

cmd/invoker/main.go
agnivade: That's just what the flag package returns.

cmd/invoker/main.go
agnivade: Thanks, I looked into it. By default, it uses `ExitOnError` (`var CommandLine = NewFlagSet(os.Args[0], ExitOnError)`). And ExitOnError is defined as `Call os.Exit(2) or for -h/-help Exit(0).` So it seems like exit(2) is the idiomatic usage for any flag error.

cmd/invoker/main.go
streamer45: I always use the `Var` variant. Example -> https://github.com/mattermost/mattermost-plugin-calls/blob/0aefee7b5827224b5c4a1ab8f517a368d97fe1c2/lt/cmd/lt/main.go#L70-L71

agnivade: Ah yes, that also works. Will leave it at this.

PR #918, opened 2025-05-19
[MM-61922] Improve validation and documentation of `RestTimeSec` and `UpdateIntervalMs`

streamer45:
#### Summary

I reviewed the ticket and related logic, and as a result, improved the documentation and configuration.

The initial login (user increment) rate is effectively fixed as of [mattermost-load-test-ng#862](https://github.com/mattermost/mattermost-load-test-ng/pull/862). I tested several configurations, and as long as sufficient resources are allocated, the rate remains constant and aligns with expectations.

`RestTimeSec` only takes effect after a degradation event occurs (i.e., once the query hit threshold is reached). At that point, the coordinator enters a different phase/state, and `RestTimeSec` determines the new rate of user increment/decrement (with some jitter introduced by the `UpdateIntervalMs`-based ticker).

I still believe `UpdateIntervalMs` could be deprecated, as making it configurable doesn't seem to add much value. A default of 1000 makes logical sense since it aligns with computing login rates per second. Let me know if you agree or if you'd prefer to keep it configurable to avoid introducing a breaking change. (If needed, we could ignore the value from the config and log a warning instead.)

image

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-61922

### DIFF
--- a/config/coordinator.sample.toml
+++ b/config/coordinator.sample.toml
@@ -5,85 +5,85 @@ RestTimeSec = 2
 [ClusterConfig]
 MaxActiveUsers = 2000
 
-    [[ClusterConfig.Agents]]
-    Id = 'lt0'
-    ApiURL = 'http://localhost:4000'
+[[ClusterConfig.Agents]]
+Id = 'lt0'
+ApiURL = 'http://localhost:4000'
 
 [MonitorConfig]
 PrometheusURL = 'http://localhost:9090'
-UpdateIntervalMs = 2000
+UpdateIntervalMs = 1000
 
-    [[MonitorConfig.Queries]]
-    Description = 'Percentage of HTTP 5xx server errors'
-    Legend = 'Percent'
-    Query = '(sum(rate(mattermost_api_time_count{status_code=~"5.."}[1m]))/sum(rate(mattermost_api_time_count[1m])))*100'
-    Threshold = 0.025
-    MinIntervalSec = 60
-    Alert = true
+[[MonitorConfig.Queries]]
+Description = 'Percentage of HTTP 5xx server errors'
+Legend = 'Percent'
+Query = '(sum(rate(mattermost_api_time_count{status_code=~"5.."}[1m]))/sum(rate(mattermost_api_time_count[1m])))*100'
+Threshold = 0.025
+MinIntervalSec = 60
+Alert = true
 
-    [[MonitorConfig.Queries]]
-    Description = 'Average client request duration'
-    Legend = 'Avg duration (s)'
-    Query = 'sum(rate(loadtest_http_request_time_sum[1m]))/sum(rate(loadtest_http_request_time_count[1m]))'
-    Threshold = 0.1
-    MinIntervalSec = 60
-    Alert = true
+[[MonitorConfig.Queries]]
+Description = 'Average client request duration'
+Legend = 'Avg duration (s)'
+Query = 'sum(rate(loadtest_http_request_time_sum[1m]))/sum(rate(loadtest_http_request_time_count[1m]))'
+Threshold = 0.1
+MinIntervalSec = 60
+Alert = true
 
-    [[MonitorConfig.Queries]]
-    Description = '99th percentile of client request duration'
-    Legend = 'P99 duration (s)'
-    Query = 'histogram_quantile(0.99, sum(rate(loadtest_http_request_time_bucket[1m])) by (le))'
-    Threshold = 2
-    MinIntervalSec = 60
-    Alert = true
+[[MonitorConfig.Queries]]
+Description = '99th percentile of client request duration'
+Legend = 'P99 duration (s)'
+Query = 'histogram_quantile(0.99, sum(rate(loadtest_http_request_time_bucket[1m])) by (le))'
+Threshold = 2
+MinIntervalSec = 60
+Alert = true
 
-    [[MonitorConfig.Queries]]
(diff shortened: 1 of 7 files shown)


coordinator/performance/monitor.go
```diff
@@ -101,7 +101,6 @@ func (m *Monitor) runQueries() Status {
 				mlog.String("query_threshold", fmt.Sprintf("%2.8f", query.Threshold)),
 			)
 			status = Status{Alert: true}
-			break
```
streamer45, line 104: This change is unrelated, but while reviewing this logic, I thought it would be better to alert about (and log) all the queries that hit the threshold rather than the first one only.

coordinator/config.go
```diff
@@ -31,6 +34,23 @@ type Config struct {
...
+func (c Config) IsValid() error {
+	if c.NumUsersInc <= 0 {
+		return errors.New("NumUsersInc should be greater than 0")
+	}
+	if c.NumUsersDec <= 0 {
+		return errors.New("NumUsersDec should be greater than 0")
+	}
+	if c.RestTimeSec <= 0 {
+		return errors.New("RestTimeSec should be greater than 0")
+	}
```
agnivade, line 46: Can we also include the values of NumUsersInc/NumUsersDec/RestTimeSec respectively in the error messages?

coordinator/config.go
```diff
@@ -31,6 +34,23 @@ type Config struct {
...
+	if c.RestTimeSec*1000 < c.MonitorConfig.UpdateIntervalMs {
+		return fmt.Errorf("RestTimerSec should greater than MonitorConfig.UpdateIntervalMs/1000 (%d)", c.MonitorConfig.UpdateIntervalMs/1000)
```
agnivade, line 48: Typo at RestTimeSec. Let's also include the value of RestTimeSec

agnivade (commented):
Ignoring the value, and logging a warning sounds fine to me.

coordinator/config.go
agnivade: Yep sounds good.

agarciamontoro (approved):
All the changes look good to me! I would even add the image you added in the PR description to the docs, because it's quite helpful.

PR is approved, so what follows is just a discussion that seems timely now :)

---

One thing we have already discussed but that I wanted to bring up again is: all this happens in average, but the users in each one of those login batches (e.g., the 10 users if `NumUsersInc` is set to 10) actually log in more or less concurrently. This, I believe, can somehow hinder measurements, because adding 10 users simultaneously every second may overload the server, while adding 1 user every 100ms may not. In both cases, we login 10 users per second in average.

The problem is that with the current logic, we cannot say "let's add 1 user every 100ms", because we reject any `UpdateIntervalMs` that is lower than 1000ms. I understand this is due to the potentially large delay on getting data back from Prometheus, but maybe we could decouple those two things: we check queries every second, but we still allow for smaller time windows between each block of user increments.

In any case, this is a larger effort and a breaking change, but it's something I really missed when doing my investigation for the ramp-up rates.

streamer45:
Thanks @agarciamontoro. Indeed, that's a valid point that we'll need to follow up on. 

The main decision and complexity to address is what to do when an alert comes in before we've finished adding users. I imagine these additions would be queued and processed asynchronously, so we'd need to cancel that operation and start decreasing instead.

Would the decrease also happen one user at a time at regular intervals? How would that affect the coordinator’s loop? And would that be sufficient to bring the threshold back down quickly?

I think we need to answer these questions before committing to an implementation.

agarciamontoro:
Yes, we do need to discuss some important questions before changing anything here, let's do that outside of this PR :)

streamer45:
@agarciamontoro To confirm, are you okay with the proposal to ignore `UpdateIntervalMs` going forward (default to 1000) and log a warning if it's found in config? If so, I'll make the changes.

config/coordinator.sample.toml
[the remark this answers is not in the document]
streamer45: Sorry :/

examples/config/ci/coordinator.json
```diff
@@ -88,7 +87,7 @@
   },
   "NumUsersInc": 8,
   "NumUsersDec": 8,
-  "RestTimeSec": 1,
```
streamer45, line 91: @agarciamontoro Not sure why this differed from our historical default of 2 seconds.

streamer45:
Sanity test looks good

image

examples/config/ci/coordinator.json
agarciamontoro: Hmmm, I don't remember the reason either, to be honest. Maybe it was to try and make the coordinator converge sooner? I'm not really sure. About removing `UpdateIntervalMs` change: CI is still running an old version of the tool, but the default is 2000 in any case, so we don't expect any change there, right?

agarciamontoro: Actually, chaning RestTimeSec from 2 to 1 will make things different, so I would move that change to another PR (after having checked the tests run normally)

streamer45: @agarciamontoro Sounds good, I can revert this bit. For my own understanding, is this file used as a source for our CI runs?

agarciamontoro: Yes it is. We should probably move it out of `examples/`, if that's confusing.

agnivade (approved):

PR #919, opened 2025-05-20
feat: support multiple operating systems in the provisioner

fmartingr:
#### Description
TODO

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-61275


fmartingr:
@agarciamontoro setting you as first reviewer, please add a second one that seems appropriate to this changes. Thanks.

agnivade (approved):
Nice work!

deployment/config.go
```diff
@@ -31,6 +31,10 @@ type Config struct {
 	AWSAvailabilityZone string `default:"us-east-1c"`
 	// AWSAMI is the AMI to use for all EC2 instances.
 	AWSAMI string `default:"ami-0fa37863afb290840"`
+	// OperatingSystem
+	OperatingSystemKind string `default:"debian" validate:"oneof:{debian,rhel}"`
+	// AWSAMIUser is the user to use when connecting to the AMI.
+	AWSAMIUser string `default:"ubuntu"`
```
agarciamontoro, line 37: Can we avoid adding this one and rely exclusively on OperatingSystemKind? So we get `ubuntu` from `debian`  and whatever RHEL users from `rhel`.

deployment/terraform/agent.go
```diff
@@ -225,7 +242,7 @@ func (t *Terraform) initLoadtest(extAgent *ssh.ExtAgent, initData bool) error {
 	if err != nil {
 		return err
 	}
-	dstPath := "/home/ubuntu/mattermost-load-test-ng/config/config.json"
+	dstPath := fmt.Sprintf("/home/%s/mattermost-load-test-ng/config/config.json", t.Config().AWSAMIUser)
```
agarciamontoro, line 245: Same nit for these dstPath. Maybe it makes sense to get a common function `fromHomeDir` or something similar.

deployment/terraform/assets/datasource.yaml
```diff
@@ -14,6 +14,7 @@ datasources:
   editable: true
   jsonData:
     timeInterval: "5s"
+    httpMethod: "GET"
```
agarciamontoro, line 17: :eyes:

deployment/terraform/assets/provisioners/debian/job.sh
```diff
@@ -17,8 +17,18 @@ do
...
+      # Install OpenTelemetry collector, using ubuntu user to avoid permission issues
+      wget https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.120.0/otelcol-contrib_0.120.0_linux_amd64.deb &&
+      sudo dpkg -i otelcol-contrib_0.120.0_linux_amd64.deb &&
+      sudo sed -i 's/User=.*/User=ubuntu/g' /lib/systemd/system/otelcol-contrib.service &&
+      sudo sed -i 's/Group=.*/Group=ubuntu/g' /lib/systemd/system/otelcol-contrib.service &&
+      sudo systemctl daemon-reload &&
+      sudo systemctl restart otelcol-contrib &&
       exit 0
-   n=$((n+1)) 
+   n=$((n+1))
```
agarciamontoro, line 31: Nice, thanks!

deployment/config.go
fmartingr: It's not that simple, since Ubuntu uses `ubuntu` while Debian uses `admin`. CentOS uses `ec2-user`. I don't know what RHEL (or others) use.

deployment/terraform/assets/datasource.yaml
fmartingr: Unsure about this one, but it wasn't working for me until I set the method to get (either manually in the web UI or by updating this file)

deployment/terraform/agent.go
fmartingr: https://github.com/mattermost/mattermost-load-test-ng/pull/919/commits/b794119eabfda6923ae5b592c40eeb61be380b10

agarciamontoro (approved):
Thanks! My suggestion was actually to get rid of the "/home/<username>" prefix that we repeat everywhere, so we could call this new function like 
```go
terraform.FromHomeDir("mattermost-load-test-ng/config/simulcontroller.json")
```
so that the function would build the home directory from its own `t.Config().AWSAMIUser`, which we are now passing to every call of `ExpandWithUser`.

I don't want to block this further, though, it's not superimportant, so I'll approve and leave it up to you :)

deployment/config.go
agarciamontoro: Ok, fair enough.

agarciamontoro: Thanks for the context :)

PR #920, opened 2025-05-26
[AI assisted for tests] MM-64392: Fix grafana pass generation in comparisons

agarciamontoro:
#### Summary
In #911, we added logic to persist the dynamically generated Grafana password. We do this by storing it locally in the `TerraformStateDir/generated_values.json` file. This works great for single deployments, but when using the `comparison` command, `TerraformStateDir` is shared among all deployments in the comparison. As the filename is hardcoded, only dependant on `TerraformStateDir`, the deployments in a comparison will inevitably override each other's data, so that the `generated_values.json` file ends up containing only the data of the last deployment that edited it. This, in turn, triggers an error when generating the results of the comparison, since dashboards cannot be uploaded without the admin password.

In order to fix this, we prefix the filename with the deployment ID, which is empty in single deployments and something like `deployment%d` in comparison deployments. This will create one file per deployment, making them fully independent.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64392

### DIFF
--- a/deployment/terraform/create.go
+++ b/deployment/terraform/create.go
@@ -77,7 +77,7 @@ func New(id string, cfg deployment.Config) (*Terraform, error) {
 		return nil, fmt.Errorf("unable to create Terraform state directory %q: %w", cfg.TerraformStateDir, err)
 	}
 
-	genValues, err := readGenValues(cfg)
+	genValues, err := readGenValues(id, cfg)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


agarciamontoro:
Tested it with a simple comparison and the dashboards were successfully uploaded. Merging and cherry-picking to `release-1.27` now.

PR #925, opened 2025-06-04
MM-64214: Fix server version string parsing

agarciamontoro:
#### Summary
The string [returned by the server in the `X-Version-ID` header](https://github.com/mattermost/mattermost/blob/1bca62dc838811dd1d20b9f25f9d57f8d1c74376/server/channels/web/handlers.go#L222) is a semver-compatible version followed by other strings separated by more dots. We need to extract the version itself for the parsing to succeed.

This bug was introduced in #908.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64214

### DIFF
--- a/api/agent.go
+++ b/api/agent.go
@@ -360,6 +360,15 @@ func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{},
 			mlog.Error("Failed to get server version", mlog.Err(err))
 		}
 	}
+
+	// The server version string looks something like
+	// 10.8.0.-14635315842.c5915cb6b6a40c3468ba242e5742cf35.true
+	// but we are only interested in the first part, 10.8.0, which is semver-compatible
+	serverVersionParts := strings.Split(serverVersionStr, ".")
+	if len(serverVersionParts) < 3 {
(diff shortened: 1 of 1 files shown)


agnivade:
How come this didn't get caught in the original PR? Seems like a clear bug to me. Is it possible to add some unit tests for this?

agarciamontoro:
> How come this didn't get caught in the original PR? Seems like a clear bug to me.

Because we don't have enough test coverage here.

> Is it possible to add some unit tests for this?

Not really, this would need an integration test with a running MM server.

I discussed these testing issues with @streamer45 yesterday and we found a compromise (I wanted a whole testing epic hehe): in the upcoming fill in the gaps epic, I'll create a ticket to add CI that runs a short, but full (deployed to AWS) test, so that we can find these issues easier.

agarciamontoro:
Yes, it's related. But the idea here would be to expand that to run it for every PR.

agarciamontoro:
The problem is that the header would send that like `10.8.0.-dev`, with a point before the prerelease part. That breaks the parsing of any semver library :(

PR #926, opened 2025-06-11
[MM-61192] Webapp load test

M-ZubairAhmed:
#### Summary
This PR enables load testing of Mattermost using actual web browsers for realistic client-side simulation, managed via a new browser agent and Node.js server, with full provisioning and orchestration support which enables simulation of real user interactions via actual browsers (using Playwright), in addition to the existing API/userentity-based approach.

#### Ticket Link
- [MM-64515](https://mattermost.atlassian.net/browse/MM-64515) Build a Playwright wrapper server to manage virtual user sessions via API
- [MM-64589](https://mattermost.atlassian.net/browse/MM-64589) Add introductory end-to-end tests to the Playwright wrapper server
- [MM-64559](https://mattermost.atlassian.net/browse/MM-64559) Create a custom Playwright reporter with useful information for load testing
- [MM-64562](https://mattermost.atlassian.net/browse/MM-64562) Create browser controller in loadtest and do a local run test
- [MM-64560](https://mattermost.atlassian.net/browse/MM-64560) Determine the number of workers required for Playwright tests
- [MM-64606](https://mattermost.atlassian.net/browse/MM-64606) Expose a config.json file for configurability of the Playwright wrapper server
- [MM-64676](https://mattermost.atlassian.net/browse/MM-64676) Introduce terraform changes for starting the browser api service by @agnivade
- [MM-64677](https://mattermost.atlassian.net/browse/MM-64677) Ensure browser server takes in values from load test config files
- [MM-64785](https://mattermost.atlassian.net/browse/MM-64785) Simplify the simulation test to be run from both the loadtesting api and manually via Playwright runner
- [MM-64765](https://mattermost.atlassian.net/browse/MM-64765) Enable both console and file logging output simultaneously in LTBrowser API
- [MM-64770](https://mattermost.atlassian.net/browse/MM-64770) Implement the /loadagent API for the load-test agent to accommodate create, run, and stop handlers for the browser controller.
- [MM-64820](https://mattermost.atlassian.net/browse/MM-64820) Add caller information to logs using pino-caller
- [MM-64829](https://mattermost.atlassian.net/browse/MM-64829) Generate OpenAPI v3 spec file for LTBrowser API
- [MM-64828](https://mattermost.atlassian.net/browse/MM-64828) Delete api of LTBrowser should not use request body as per HTTP specification
- [MM-64909](https://mattermost.atlassian.net/browse/MM-64909) nvm and consequently
node.js, isn’t being installed through the provisioning scripts on the instances
- [MM-65016](https://mattermost.atlassian.net/browse/MM-65016) Add automatic screenshot when simulation test errors out
- [MM-64954](https://mattermost.atlassian.net/browse/MM-64954) Create smoke simulation runner which enables dry runs of tests in AWS instance
- [MM-64931](https://mattermost.atlassian.net/browse/MM-64931) LTBrowser API browser sessions launch failures in Ubuntu/AWS Environments
- [MM-64929](https://mattermost.atlassian.net/browse/MM-64929) LTBrowser API log messages doesnt contain error responses

[MM-64515]: https://mattermost.atlassian.net/browse/MM-64515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MM-64589]: https://mattermost.atlassian.net/browse/MM-64589?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MM-64559]: https://mattermost.atlassian.net/browse/MM-64559?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MM-64562]: https://mattermost.atlassian.net/browse/MM-64562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MM-64560]: https://mattermost.atlassian.net/browse/MM-64560?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MM-64606]: https://mattermost.atlassian.net/browse/MM-64606?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[MM-64676]: https://mattermost.atlassian.net/browse/MM-64676?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


agnivade (commented):
Just some initial comments.

browser/.env.sample
```diff
@@ -0,0 +1,10 @@
+# Playwright's
+# see https://playwright.dev/docs/browsers#hermetic-install
+PLAYWRIGHT_BROWSERS_PATH=0
+
+# Server's
+DEBUG_LOGS='true'
+PORT='YOUR PORT'
+HOST='YOUR HOST'
```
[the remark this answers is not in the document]
M-ZubairAhmed: done, made it accessible via this route

browser/src/simulations/scenario1.ts
```diff
@@ -0,0 +1,28 @@
...
+    await page.goto(serverURL);
+    await page.waitForNavigation();
+    await handlePreferenceCheckbox(page);
+    await performLogin({page, userId, password});
+
+    while (true) {
+      await postInChannel({page});
+      await scrollInChannel(page, 'sidebarItem_off-topic', 40, 400, 500);
+      await scrollInChannel(page, 'sidebarItem_town-square', 40, 400, 500);
+    }
```
agarciamontoro, line 24: When does this finish?

browser/src/utils/config.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Generates a random port in the specified range
+ * Useful when you want to avoid sequential ports for parallel tests
+ */
+export function getRandomPortForTests(): number {
+  const min = 10000;
+  const max = 65000;
+  return Math.floor(Math.random() * (max - min + 1)) + min;
+}
```
agarciamontoro, line 58: There's no "specified range" here, just a hardcoded one. Which is good, no problem there, but it doesn't match the comment.

browser/src/utils/log.ts
```diff
@@ -0,0 +1,65 @@
...
+    },
+
+    warn: (message: string) => {
+      return logger.warn(message);
+    },
+
+    info: (message: string) => {
+      return logger.info(message);
+    },
+  };
```
agarciamontoro, line 64: Nit:
```suggestion
  return {
    error: logger.error,
    warn: logger.warn,
    info: logger.info,
  };
```

coordinator/cluster/cluster.go
```diff
@@ -127,6 +162,13 @@ func (c *LoadAgentCluster) Shutdown() {
 			}
 		}(ag)
 	}
+
+	for _, browserAgent := range c.browserAgents {
+		if _, err := browserAgent.Destroy(); err != nil {
+			c.log.Error("cluster: failed to stop browser agent", mlog.Err(err))
+		}
+	}
```
agarciamontoro, line 170: Something like this (not tested):
```suggestion
	for _, browserAgent := range c.browserAgents {
		go func(ag *client.Agent) {
			dever wg.Done()
			if _, err := browserAgent.Destroy(); err != nil {
				c.log.Error("cluster: failed to stop browser agent", mlog.Err(err))
			}
		}(browserAgent)
	}
```

coordinator/cluster/cluster.go
```diff
@@ -218,6 +260,34 @@ func (c *LoadAgentCluster) Status() (Status, error) {
...
+	for _, browserAgent := range c.browserAgents {
+		st, err := browserAgent.Status()
+		// Agent probably crashed. We create it again.
+		if errors.Is(err, client.ErrAgentNotFound) {
+			if err := createAgent(browserAgent, c.ltConfig); err != nil {
+				c.log.Error("browser agent create failed", mlog.Err(err))
+			}
+		} else if err != nil {
+			c.log.Error("cluster: failed to get status for browser agent:", mlog.Err(err))
+		}
```
agarciamontoro, line 273: In case of an error, we should either `continue` or directly return an error.

coordinator/cluster/config.go
```diff
@@ -27,11 +27,23 @@ type LoadAgentClusterConfig struct {
 	// MaxActiveUsers defines the upper limit of concurrently active users to run across
 	// the whole cluster.
 	MaxActiveUsers int `default:"1000" validate:"range:(0,]"`
+	// BrowserAgents is a list of the browser agents API endpoints to be used during
+	// the load-test. It's length defines the number of browser agents instances
```
agarciamontoro, line 31: Typo
```suggestion
	// the load-test. Its length defines the number of browser agents instances
```

coordinator/cluster/config.go
```diff
@@ -27,11 +27,23 @@ type LoadAgentClusterConfig struct {
...
+	// used during a load-test.
+	BrowserAgents []LoadAgentConfig `default_size:"5"`
```
agarciamontoro, line 33: We can also `validate` this one with a `range(0,]`

docs/browser.md
```diff
@@ -0,0 +1,400 @@
...
+     - Scrolling through message history
+
+### Configuration
+
+The project uses TypeScript and is configured with:
+- `tsconfig.json` for TypeScript compilation
+- `playwright.config.ts` for Playwright test settings
+- `prettier.config.cjs` for code formatting
+- Environment variables for runtime configuration
+
```
agarciamontoro, line 42: This is very thorough, but I'm a bit concerned it will be difficult to keep it updated, specially in these early stages of the browser agents. Thoughts?

docs/browser.md
```diff
@@ -0,0 +1,400 @@
...
+make help                 # Show all available commands
+make install             # Install dependencies and Playwright browser
+make install-dependencies # Install only Node.js dependencies
+make install-playwright  # Install Playwright Chromium browser
+make build              # Build the TypeScript server
+make start              # Start server in production mode
+make start-dev          # Start server in development mode
+make start-watch        # Start server in watch mode (auto-restart)
+make clean              # Clean build artifacts and dependencies
+```
```
agarciamontoro, line 214: I suggest we document the commands in the Makefile itself and here only say that `make help` will list them all.

docs/browser.md
```diff
@@ -0,0 +1,400 @@
...
+   - Rate limiting
+   - Proper error handling
+
+## Contributing
+
+1. Fork the repository
+2. Create a feature branch
+3. Commit your changes
+4. Push to the branch
+5. Create a Pull Request
```
agarciamontoro, line 280: Do we need this section?

docs/browser.md
```diff
@@ -0,0 +1,400 @@
...
+
+## API Documentation
+
+The Browser Load Testing Agent exposes a RESTful API for managing browser sessions and monitoring health status.
+
+### Endpoints
```
agarciamontoro, line 286: As before, I think we should avoid this unless we can automatically generate it somehow. It may be ok for now, but it can become outdated very quickly.

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,300 @@
...
+// TODO: Get this from the coordinator file
+const BROWSER_AGENT_API_URL = "BROWSER_AGENT_API_URL"
+
+// BrowserController is a controller that manages browser sessions
+// by communicating with the browserLt API server.
+type BrowserController struct {
+	id            int
+	user          user.User
+	status        chan<- control.UserStatus
+	rate          float64
```
agarciamontoro, line 29: Why do we need the rate here?

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+	if status == nil {
+		return nil, errors.New("status channel cannot be nil")
+	}
+
+	// Get browser API URL from environment variable (same as browser server uses)
+	browserAPIURL := os.Getenv("BROWSER_AGENT_API_URL")
+	if browserAPIURL == "" {
+		return nil, errors.New("BROWSER_AGENT_API_URL environment variable is required")
+	}
```
agarciamontoro, line 73: I understand we will be getting this from the config, right?

browser/src/lib/browser_manager.ts
```diff
@@ -0,0 +1,303 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {Browser, BrowserContext, chromium, Page} from 'playwright';
+
+import {log} from '../app.js';
+import {testManager} from '../lib/test_manager.js';
+import {isBrowserHeadless} from '../utils/config.js';
+
+const CLEANUP_TIMEOUT = 4 * 1000; // 2 seconds
```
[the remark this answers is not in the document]
M-ZubairAhmed: done

browser/src/lib/browser_manager.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    return {
+      isRemoved: true,
+      message: `Browser instance scheduled for removal for user ${userId}`,
+    };
+  }
+
+  /**
+   * Start tests asynchronously for a browser session
+   * It infroms if the test failed or was stopped.
```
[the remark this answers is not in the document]
M-ZubairAhmed: done

browser/src/simulations/login.ts
```diff
@@ -0,0 +1,55 @@
...
+  log.info('[simulation][run][handlePreferenceCheckbox]');
+
+  try {
+    // Try to find the checkbox with a short timeout
+    await page.waitForSelector('label.get-app__preference input.get-app__checkbox', {timeout: 2000});
+
+    // If found, click it
+    await page.click('label.get-app__preference input.get-app__checkbox');
+
+    await page.waitForSelector('a.btn.btn-tertiary.btn-lg');
```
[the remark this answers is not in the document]
M-ZubairAhmed: this will be done in https://mattermost.atlassian.net/browse/MM-64564

browser/src/simulations/post_in_channel.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  try {
+    await page.waitForSelector('#post_textbox');
+    await page.type('#post_textbox', `Hello, world! ${new Date().toISOString()}`, {delay: 100});
+    await page.keyboard.press('Enter');
+
+    log.info('[simulation][pass][postInChannel]');
+  } catch (error) {
+    throw {error, testId: 'postInChannel'};
+  }
```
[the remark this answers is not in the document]
M-ZubairAhmed: as may be i should document, yes this all gets logged in test manager

browser/src/simulations/scenario1.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+// See LICENSE.txt for license information.
+
+import {test} from '@playwright/test';
+
+import {handlePreferenceCheckbox, performLogin} from './login.js';
+import {postInChannel} from './post_in_channel.js';
+import {scrollInChannel} from './scrolling_in_channel.js';
+
+test('Scenario 1', async ({page}) => {
+  await page.goto('https://community.mattermost.com/core/channels/zubairloadtest1');
```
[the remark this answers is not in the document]
M-ZubairAhmed: now i changed

browser/src/simulations/scenario1.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+  await page.waitForNavigation();
+  await handlePreferenceCheckbox(page);
+  await performLogin({page, userId: 'zubair.loadtest3@maildrop.cc', password: 'Password-1!'});
```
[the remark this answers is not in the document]
M-ZubairAhmed: remove 😛

browser/src/simulations/scenario1.ts
M-ZubairAhmed: it doesn't until manually stopped

browser/src/utils/config.ts
M-ZubairAhmed: changed

browser/src/utils/log.ts
M-ZubairAhmed: done

docs/browser.md
M-ZubairAhmed: removed

docs/browser.md
M-ZubairAhmed: yes got it

loadtest/control/browsercontroller/controller.go
M-ZubairAhmed: implementation is now changed, remove this getenv

M-ZubairAhmed:
> @M-ZubairAhmed - is this ready to be reviewed?

Its ready now. Thank you

agnivade (commented):
Gave it a first pass

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,338 @@
...
+		},
+		isRunning: false,
+	}, nil
+}
+
+// Run starts the browser controller by creating a browser session.
+// This corresponds to adding a user in the browser load test.
+func (c *BrowserController) Run() {
+	c.mu.Lock()
+	if c.isRunning {
```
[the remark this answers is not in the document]
M-ZubairAhmed: i am using that to apply mutex lock on the function.

browser/src/utils/config.ts
```diff
@@ -0,0 +1,61 @@
...
+import fs from 'fs';
+import {fileURLToPath} from 'url';
+
+/**
+ * Read and parse the config.json file located in config/config.json
+ */
+export function loadConfigJson() {
+  try {
+    const dirname = path.dirname(fileURLToPath(import.meta.url));
+    const configPath = path.resolve(dirname, '../../../config/config.json');
```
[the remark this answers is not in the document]
M-ZubairAhmed: Good point! I’ve created a ticket https://mattermost.atlassian.net/browse/MM-64819. We’ll handle the JSON format first and revisit TOML support in an upcoming update. Thanks for flagging that!

browser/src/simulations/scenario_1.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {handlePreferenceCheckbox, performLogin} from './login.js';
+import {postInChannel} from './post_in_channel.js';
+import {scrollInChannel} from './scrolling_in_channel.js';
+import type {BrowserInstance} from '../lib/browser_manager.js';
+
+export async function scenario1({page, userId, password}: BrowserInstance, serverURL: string, runInLoop = true) {
```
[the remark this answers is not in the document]
M-ZubairAhmed: I’ve renamed it to noop to align with the existing noop controller. It reflects the simplicity of the operation and should be easy to recognize in context.

browser/src/lib/browser_manager.ts
```diff
@@ -0,0 +1,302 @@
...
+      this.activeBrowserSessions.set(userId, creationFailedInstance);
+
+      log.error(`[browser_manager] Failed to create context for "${userId}"`);
+      return {
+        isCreated: false,
+        message: `Failed to create context for user "${userId}"`,
+      };
+    }
+
+    // Try to create the page third
```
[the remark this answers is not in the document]
M-ZubairAhmed: Improved the structure.

browser/src/lib/browser_manager.ts
```diff
@@ -0,0 +1,302 @@
...
+      this.activeBrowserSessions.set(userId, creationFailedInstance);
+
+      log.error(`[browser_manager] Failed to create browser instance for "${userId}"`);
+      return {
+        isCreated: false,
+        message: `Failed to create browser instance for user "${userId}"`,
+      };
+    }
+
+    // Try to create the context second
```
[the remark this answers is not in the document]
M-ZubairAhmed: changed the structure

api/server.go
```diff
@@ -35,11 +36,14 @@ func (a *api) getResource(id string) (interface{}, bool) {
...
 }
 
+// It is used to store the load-test agent in the resources map.
+// which can be retrieved later using getResource
 func (a *api) setResource(id string, res interface{}) bool {
 	a.mut.Lock()
 	defer a.mut.Unlock()
 	if _, ok := a.resources[id]; !ok {
 		a.resources[id] = res
+		mlog.Info("API Resource set for id: " + id)
```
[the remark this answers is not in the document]
M-ZubairAhmed: I thought it would be good to log that as well, but yeah it can be removed.

api/client/agent/agent.go
```diff
@@ -171,6 +171,8 @@ func (a *Agent) Status() (loadtest.Status, error) {
 func (a *Agent) Run() (loadtest.Status, error) {
 	var status loadtest.Status
 	resp, err := a.apiPost(a.apiURL+a.id+"/run", nil)
+
+	//TODO: create a new endpot '/run-browser' for browser agent
```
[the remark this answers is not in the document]
M-ZubairAhmed: No removed this and other TODOs which i added

api/agent.go
```diff
@@ -591,3 +607,32 @@ func createCustomEmoji(config *loadtest.Config) error {
...
+// Checks if the current instance is going to run browser agents.
+func isBrowserAgentInstance() bool {
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		mlog.Warn("Failed to get home directory for searching agent_type.txt file", mlog.Err(err))
+		return false
+	}
+
+	// Check for the agent type file created by configureAndRunAgents in deployment/terraform/agent.go
+	agentTypeFile := homeDir + "/" + deployment.AgentTypeFileName
```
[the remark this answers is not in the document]
M-ZubairAhmed: Done, used `filepath.Join` to create the file path.

browser/src/lib/browser_manager.ts
```diff
@@ -0,0 +1,302 @@
...
+      instance = {...instance, browser};
+      this.activeBrowserSessions.set(userId, instance);
+    } catch (error) {
+      const creationFailedInstance: BrowserInstance = {
+        ...instance,
+        state: SessionState.CREATION_FAILED,
+      };
+      this.activeBrowserSessions.set(userId, creationFailedInstance);
+
+      log.error(`[browser_manager] Failed to create browser instance for "${userId}"`);
```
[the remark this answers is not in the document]
M-ZubairAhmed: The line numbers not being included by default. Looking at our actual log output, Pino only includes level, timestamp, PID, hostname and message.

While we could add filename/line number logging using pino-caller for development, it has significant performance overhead and is explicitly [not recommended for production use.](https://github.com/pinojs/pino-caller?tab=readme-ov-file#install) . The [simulation] prefix is particularly valuable as it creates a clear audit trail of test execution others prefixes i have removed.

I created a ticket to add pino-caller conditionally for development environments only https://mattermost.atlassian.net/browse/MM-64820

loadtest/control/browsercontroller/controller.go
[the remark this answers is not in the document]
M-ZubairAhmed: The `isRunning` flag is needed because `BrowserController` is different from other controllers. Other controllers use pure channel-based synchronization and their Run() method blocks until stopped.
`BrowserController` makes HTTP API calls to an external(which is internal!) browser service, which can fail or hang. Without the `isRunning` flag, we could have:
- Multiple concurrent Run() calls if the first one fails quickly
- Multiple concurrent Stop() calls trying to remove the same browser session
- Race conditions between Run() and Stop() methods
The `isRunning` flag with mutex protection prevents these race conditions that are unique to the HTTP-based nature of this controller.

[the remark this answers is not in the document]
agnivade: All of the above could be done by simply wrapping the Run and Stop methods with a mutex. Do we need a flag for that?

api/agent.go
```diff
@@ -116,7 +117,10 @@ func (a *api) createLoadAgentHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	isBAInstance := isBrowserAgentInstance()
+	isBAInstance, err := isBrowserAgentInstance()
+	if err != nil {
+		mlog.Warn("failed to detect agent_type", mlog.Err(err))
```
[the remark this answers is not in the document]
M-ZubairAhmed: Added the above message in the log

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,297 @@
...
+	}
+	if status == nil {
+		return nil, errors.New("status channel cannot be nil")
+	}
+
+	return &BrowserController{
+		id:              id,
+		user:            user,
+		status:          status,
+		rate:            1.0,
```
agarciamontoro, line 73: Commenting here as well in case my answer in the old thread is not visible: we don't really need to store the rate in this controller, we can simply implement SetRate as `SetRate(_ float64) error { return nil }`

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,297 @@
...
+			c.status <- c.newInfoStatus(resp.Info)
+		}
+
+		// If stop signal is received then stop the controller
+		select {
+		case <-c.stopChan:
+			return
+		default:
+		}
+	}
```
agarciamontoro, line 128: Shouldn't we do this via the browser itself?

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,297 @@
...
+	if userStore.Password() == "" {
+		return fmt.Errorf("password is empty")
+	}
+
+	requestBody := AddBrowserRequest{
+		UserID:   userStore.Username(),
+		Password: userStore.Password(),
+	}
+
+	_, err := c.makeRequestToLTBrowserApi("POST", "/browsers", requestBody)
```
agarciamontoro, line 235: Instead of hardcoding `POST` or `DELETE` in these requests, we should use `http.MethodPost`, `http.MethodDelete` and so on.

browser/src/simulations/noop_scenario.ts
```diff
@@ -0,0 +1,32 @@
...
+
+import {handlePreferenceCheckbox, performLogin} from './login.js';
+import {postInChannel} from './post_in_channel.js';
+import {scrollInChannel} from './scrolling_in_channel.js';
+import type {BrowserInstance} from '../lib/browser_manager.js';
+
+// NoopScenario is a very basic scenario.
+// It just performs very basic actions in a loop. Such as:
+// Login, Post in channel, Scroll in channel and then repeat the process.
+export async function noopScenario({page, userId, password}: BrowserInstance, serverURL: string, runInLoop = true) {
```
agarciamontoro, line 12: Nit: three operations don't look like a "noop" to me 😂 What about `postAndScrollScenario` or something similar?

browser/src/simulations/noop_scenario.ts
```diff
@@ -0,0 +1,32 @@
...
+    await handlePreferenceCheckbox(page);
+    await performLogin({page, userId, password});
+
+    // Runs the simulation atleast once and then runs it in continuous loop if simulationMode is true
+    do {
+      const scrollCount = runInLoop ? 40 : 3;
+      await postInChannel({page});
+      await scrollInChannel(page, 'sidebarItem_off-topic', scrollCount, 400, 500);
+      await scrollInChannel(page, 'sidebarItem_town-square', scrollCount, 400, 500);
+    } while (runInLoop);
```
agarciamontoro, line 28: Typo, and should `simulationMode` be `runInLoop`?
```suggestion
    // Runs the simulation at least once and then runs it in continuous loop if runInLoop is true
    do {
      const scrollCount = runInLoop ? 40 : 3;
      await postInChannel({page});
      await scrollInChannel(page, 'sidebarItem_off-topic', scrollCount, 400, 500);
      await scrollInChannel(page, 'sidebarItem_town-square', scrollCount, 400, 500);
    } while (runInLoop);
```

loadtest/control/browsercontroller/controller_test.go
```diff
@@ -0,0 +1,435 @@
...
+		require.Contains(t, statusF.Err.Error(), "browser controller was not initialized")
+	})
+
+	t.Run("controller lifecycle with browser API not running", func(t *testing.T) {
+		controllerWithUser, statusChan := newController(t)
+
+		go controllerWithUser.Run()
+
+		// Test that controller is running after Run is called
+		status := <-statusChan
```
agarciamontoro, line 109: Should we add a `select` with a timeout in case the code fails, so that the test can finish in a reasonable amount of time? This comment applies to other channels being read in this file.

loadtest/control/browsercontroller/controller_test.go
```diff
@@ -0,0 +1,435 @@
...
+					require.NoError(t, err)
+					require.Equal(t, "testuser", requestBody.UserID)
+					require.Equal(t, "testpass", requestBody.Password)
+				} else if method == "DELETE" {
+					var requestBody RemoveBrowserRequest
+					err := json.NewDecoder(r.Body).Decode(&requestBody)
+					require.NoError(t, err)
+					require.Equal(t, "testuser", requestBody.UserID)
+				}
+			}
```
agarciamontoro, line 233: Nit: Can we use a `switch` here covering only the "POST" and "DELETE" cases?

agarciamontoro (changes requested):
Amazing work, @M-ZubairAhmed! I've left a lot of comments, but most of them are nits.

agnivade:
Thanks for the work on this @M-ZubairAhmed. I gave it a test myself locally and found some more issues. Most of them are minor , and just needs some polish.

1. New fields under `BrowserLogSettings` in config.json should be validated during browser service start. Currently, we are loading the json object in `loadConfigJson()` and then accessing each field in the code. This is creating undefined references during runtime and leads to weird errors which are hard to diagnose. We should validate all of these fields during the loading of the file itself.
2. The errors thrown inside the try-catch blocks in `createBrowserSession` are not logged at all. There is a log statement, but we also need to log the `error` variable of the exception. Without this, debugging becomes very difficult.
3. In test_manager.ts, L66, there is no such field: error.error.message. Let's correct it to `error.error`.
4. Preferably just throw the error in noop_scenario.ts rather than `errror?.error` to keep things consistent with other scenarios.
5. Lastly, there's a deadlock in the BrowserController Run and Stop methods. We are using a mutex to protect the Run and Stop methods, but the Run method doesn't exit until the `c.stopChan` is closed. And `c.stopChan` won't be closed until the Stop method gets to run. And that won't happen because it is blocked on the mutex locked by the Run method. Hence, deadlock. Let's fix this as well.

Overall, great work. Fully realize you are stepping out of your comfort zone here with Go code. Please let me know if there's something I can help you with.

api/server.go
```diff
@@ -35,6 +36,8 @@ func (a *api) getResource(id string) (interface{}, bool) {
 	return val, ok
 }
 
+// It is used to store the load-test agent in the resources map.
+// which can be retrieved later using getResource
 func (a *api) setResource(id string, res interface{}) bool {
```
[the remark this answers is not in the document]
M-ZubairAhmed: Thanks did'nt knew this

coordinator/status.go
```diff
@@ -82,10 +82,12 @@ func (s State) String() string {
...
-	NumErrors      int64     // Total number of errors received from the load-test agents cluster.
-	SupportedUsers int       // Number of supported users.
+	State              State     // State of Coordinator.
+	StartTime          time.Time // Time when Coordinator has started.
+	StopTime           time.Time // Time when Coordinator has stopped.
+	ActiveUsers        int       // Total number of currently active users across the load-test agents cluster.
+	NumErrors          int64     // Total number of errors received from the load-test agents cluster.
+	SupportedUsers     int       // Number of supported users.
+	ActiveBrowserUsers int       // Total browser users
+	NumBrowserErrors   int64     // Total browser errors
```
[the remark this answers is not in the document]
M-ZubairAhmed: Is this a convention as well?

deployment/opensearch/roundtripper_test.go
```diff
@@ -130,7 +130,7 @@ func checkSignature(t *testing.T, req *http.Request) {
 
 	// Check all items were correctly computed
 	require.Equal(t, algorithm, "AWS4-HMAC-SHA256")
-	today := time.Now().Format("20060102")
+	today := time.Now().UTC().Format("20060102")
```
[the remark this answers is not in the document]
M-ZubairAhmed: This change is necessary to fix a flaky test that was failing intermittently due to system's timezone.

deployment/terraform/coordinator.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/mattermost/mattermost/server/public/shared/mlog"
 )
 
+const (
+	LT_API_PORT = "4000"
```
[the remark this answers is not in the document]
M-ZubairAhmed: Thanks for linking the Go style guide! You're absolutely right about using camelCase.

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,297 @@
...
+	"net/http"
+	"sync"
+	"time"
+
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/control"
+	"github.com/mattermost/mattermost-load-test-ng/loadtest/user"
+)
+
+// This is the URL of the LTBrowser API server ran from /browser
+const LT_BROWSER_API_URL = "http://localhost:5000"
```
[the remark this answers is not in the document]
M-ZubairAhmed: Done as well

coordinator/cluster/cluster.go
M-ZubairAhmed: You're absolutely right! There was a bug in the wait group logic. I was adding len(c.agents) + len(c.browserAgents) to the wait group, but the browser agents loop was missing the goroutine and wg.Done() call.

coordinator/cluster/cluster.go
M-ZubairAhmed: I improved the error handling by adding `continue` statements to both error cases and added clear comments explaining the reasoning.

coordinator/cluster/config.go
M-ZubairAhmed: This typo is fixed.

loadtest/control/browsercontroller/controller.go
M-ZubairAhmed: For now keeping it here is fine for simplicity. Once the browser simulation tests get more robust, we can move this logic there.

browser/src/simulations/noop_scenario.ts
M-ZubairAhmed: Fixed the typo

loadtest/control/browsercontroller/controller.go
M-ZubairAhmed: That's a good suggestion, its is now replaced with those.

browser/src/simulations/noop_scenario.ts
M-ZubairAhmed: Yes makes sense

loadtest/control/browsercontroller/controller_test.go
M-ZubairAhmed: > in case the code fails,

Can you clarify which code would fail?

> the test can finish in a reasonable amount

Simulation test are supposed to be unbounded and run till infinity

browser/src/lib/browser_manager.ts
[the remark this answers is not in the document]
M-ZubairAhmed: The logs now are 

```
INFO [2025-07-14 09:55:24.307 +05:30] (28430) <BrowserTestSessionManager.shutdown (file://browser/src/lib/browser_manager.ts:1:5546)>: no active browser sessions to clean up in browser_manager

INFO [2025-07-14 09:55:24.310 +05:30] (28430) <stopServer (file://browser/src/server.ts:1:578)>: LTBrowser server stopped
```

agnivade:
Everything looks good. The only remaining thing would be to run a full blown test using Terraform. Have you done that yet?

streamer45 (commented):
Massive work on this 👏🙌 Gave it a super quick pass, but it looks great overall. Looking forward to seeing it in action soon!

My only serious note is that, conceptually, treating browser-driven users the same as any other user/controller would have been (in my opinion) cleaner and would have leaked far fewer implementation details, likely requiring close to zero changes at higher levels (e.g., the coordinator).

browser/src/server.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {app, log} from './app.js';
+
+// This is a constant for the API URL of the LTBrowser server.
+// Also defined in loadtest/control/browsercontroller/controller.go as LTBrowserApi
+const LT_BROWSER_API_URL = 'http://localhost:5000';
```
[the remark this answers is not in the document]
M-ZubairAhmed: Yes, even the LT API URL is currently hardcoded in the app, but we can take care of that in upcoming work.

agnivade:
@M-ZubairAhmed - Were you able to test successfully with a Terraform deployment?

deployment/terraform/assets/provisioners/debian/agent.sh
```diff
@@ -16,12 +16,13 @@ until [ "$n" -ge 3 ]; do
...
 		sudo apt-get install -y numactl linux-tools-aws &&
-		# Install nvm and latest LTS Node.js
+		# Install nvm - Node.js version manager
 		curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | bash &&
 		export NVM_DIR="$HOME/.nvm" &&
 		[ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh" &&
-		nvm install --lts &&
-		nvm use --lts &&
+		# Install the Node.js version defined in .nvmrc and use it
+		nvm install &&
```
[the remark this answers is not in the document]
M-ZubairAhmed: Ideally, we’d want to stick with whatever version we’ve defined in NVM — in this case, that’s the LTS version. That said, Node can have its own quirks, so it’s good to be extra cautious and manually check if everything works with that version.

docs/faq.md
```diff
@@ -179,3 +179,7 @@ You will have to run this for every loadtest agent you have. These will be appen
 These are intended for users that need to override the connection URL to their Mattermost server in their load tests environments, in most cases because there's a custom reverse proxy in front of the load-test deployment.
 
 This **should not be used in most cases** as the loadtest agent will automatically detect the server URL and scheme from the configuration and deployed services.
+
+### When starting deployment create, it fails due to `No default subnet for availability zone: 'AWS_SUBNETS_HERE'.`
+
+This likely happens because AWS doesn’t have an available subnet in the zone we explicitly set in the Terraform config. You can fix this by leaving `AWSAvailabilityZone` empty in the `config/deployer.json`.
```
agarciamontoro, line 185: Thanks! We could add something like:
> Note that leaving this empty may increment the traffic-related AWS costs for very large tests, since resources are no longer forced to be deployed in the same Availability Zone.

loadtest/control/browsercontroller/controller.go
```diff
@@ -0,0 +1,294 @@
...
+			return
+		default:
+		}
+	}
+
+	if err := c.addBrowser(); err != nil {
+		c.status <- c.newErrorStatus(control.NewUserError(err))
+		return
+	}
+	c.status <- c.newInfoStatus("browser added successfully")
```
agarciamontoro, line 123: I know we discussed this in our call, but don't remember our decision: are we keeping the `control` actions here and remove them in a future PR, or are we removing them now? We should be aware that this essentially creates two sessions: one controlled by the `control` functions, another controlled by the browser.

agarciamontoro (changes requested):
Again, amazing work!!!!!! There are a few things we still need to fix, but minor things at this point. I think you're still struggling with the Terraform deployments, let me know if I can help with anything!

docs/faq.md
M-ZubairAhmed: added the note

browser/src/simulations/post_and_scroll_scenario.ts
```diff
@@ -0,0 +1,29 @@
...
+  await handlePreferenceCheckbox(page);
+  await performLogin({page, userId, password});
+
+  // Runs the simulation at least once and then runs it in a continuous loop if simulationMode is true
+  do {
+    const scrollCount = runInLoop ? 40 : 3;
+    await postInChannel({page});
+    await scrollInChannel(page, 'sidebarItem_off-topic', scrollCount, 400, 500);
+    await scrollInChannel(page, 'sidebarItem_town-square', scrollCount, 400, 500);
+  } while (runInLoop);
```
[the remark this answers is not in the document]
M-ZubairAhmed: changed, it was an old variable name

defaults/validate.go
```diff
@@ -253,6 +253,14 @@ func validateFromRange(value reflect.Value, mins, maxs, minInterval, maxInterval
...
 		}
 		val = value.Float()
+	case reflect.Slice:
+		if mins != "" {
+			min, err = strconv.ParseFloat(mins, 64)
+		}
+		if maxs != "" {
+			max, err = strconv.ParseFloat(maxs, 64)
+		}
+		val = float64(value.Len())
```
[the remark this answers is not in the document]
M-ZubairAhmed: reverted this validation check

loadtest/control/browsercontroller/controller.go
M-ZubairAhmed: I removed the login and join team method and moved them to simulation tests so we no longer have the duplicate sessions anymore

agnivade (approved):
The only pending request from my side was to get a Terraform deployment up and running. Now that it's done, this is good to go!

api/agent.go
```diff
@@ -333,7 +347,7 @@ func getServerVersion(serverURL string) (string, error) {
 
 // NewControllerWrapper returns a constructor function used to create
 // a new UserController.
-func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{}, userOffset int, namePrefix string, metrics *performance.Metrics) (loadtest.NewController, error) {
+func NewControllerWrapper(config *loadtest.Config, controllerConfig interface{}, userOffset int, namePrefix string, metrics *performance.Metrics, isBrowserAgentInstance bool) (loadtest.NewController, error) {
 	maxHTTPconns := loadtest.MaxHTTPConns(config.UsersConfiguration.MaxActiveUsers)
```
agarciamontoro, line 351: We should change this to `MaxActiveBrowserUsers` if it's a browser agent instance:
```suggestion
	maxHTTPconns := loadtest.MaxHTTPConns(config.UsersConfiguration.MaxActiveUsers)
	if isBrowserAgentInstance {
		maxHTTPconns = loadtest.MaxHTTPConns(config.UsersConfiguration.MaxBrowserActiveUsers)
	}
```

loadtest/loadtest.go
```diff
@@ -305,6 +324,7 @@ func New(config *Config, nc NewController, log *mlog.Logger) (*LoadTester, error
 		status:            Status{},
 		activeControllers: make([]control.UserController, 0),
 		idleControllers:   make([]control.UserController, 0),
+		isBrowserAgent:    isBrowserAgent,
 		log:               log,
```
agarciamontoro, line 328: Adding the comment here because I cannot add it to the specific line (322), but `statusChan` should be something like
```go
statusChan: make(chan control.UserStatus, config.UsersConfiguration.MaxActiveUsers + config.UsersConfiguration.MaxActiveBrowserUsers),
```

agarciamontoro (changes requested):
I've been testing this and I found some weird behaviour, as we discussed in the DM. Not sure if this is me or something that we need to fix. The summary of what I tried:
- I configured 2 normal agents and 2 browser agents, and I set the following config in `coordinator.json`:
  - `"MaxActiveUsers": 1`
  - `"MaxActiveBrowserUsers": 4`
- Then, in `config.json`:
- `"MaxActiveUsers" : 1`
- `"MaxActiveBrowserUsers": 2`

With this, I expected to have 1 server user and 4 browser users. 5 users in total. However, I saw this in Grafana, just one user connected:
only one user connected

Later, inspecting the logs in the browser instance itself, I saw this error over and over:
```
failed to add browser with the browser controller failed as response is in-between (200,300]: response status code: 400, response body: {"success":false,"error":{"code":"CREATE_SESSION_FAILED","message":"Failed to create browser instance for user agmbrowser-browser-agent-0-10, error: Error: browserType.launch: Executable doesn't exist at /home/ubuntu/mattermost-load-test-ng/browser/node_modules/playwright/node_modules/playwright-core/.local-browsers/chromium_headless_shell-1179/chrome-linux/headless_shell\n╔═════════════════════════════════════════════════════════════════════════╗\n║ Looks like Playwright Test or Playwright was just installed or updated. ║\n║ Please run the following command to download new browsers:              ║\n║                                                                         ║\n║     npx playwright install                                              ║\n║                                                                         ║\n║ <3 Playwright Team                                                      ║\n╚═════════════════════════════════════════════════════════════════════════╝"}} <nil>
```

docs/config/coordinator.md
```diff
@@ -26,6 +26,28 @@ The URL to the load-test API server that will run the agent.
...
+
+*string*
+
+The URL to the load-test API server that will run the agent. Since this is an agent as well, it will be the same as the load-test API server. This should not be confused with the LTBrowser API server which is a separate server that is controlled by the load-test API server.
+
+### MaxActiveBrowserUsers
+
+*int*
+
+The maximum number of concurrently active browser users to be run across the entire load-agent cluster. Note that there is also a MaxActiveBrowserUsers setting in config/config.json which limits the number of browser users per individual agent, while this setting limits the total number of browser users across all agents in the cluster. The MaxActiveBrowserUsers value in the cluster configuration should be less than the product of MaxActiveBrowserUsers per browser agent multiplied by the number of browser agents.
```
agarciamontoro, line 49: We should add a note here stating that this cannot be larger than MaxActiveUsers, since MaxActiveUsers is the cap of total users connected.

agarciamontoro (approved):
After that nasty `--only-shell`/`--no-shell` got out of the way (thank you for looking into that!) I was able to finally test this from scratch, deploying and starting a test with browser agents.

This was an *amazing* piece of work, @M-ZubairAhmed, thank you so much for working on this and for your infinite patience with my review comments and testing. We're ready to merge*!!!


\* I did add a couple of comments on docs and sample config files :grimacing:

PR #927, opened 2025-06-18
MM-64590: Use the maximum permitted size for RX ring buffer

agarciamontoro:
#### Summary
Instead of using a default of 1024 (and instead of using a default of 8192 as proposed in the ticket), we now always use the maximum permitted size for the RX ring buffer.

I have not found a way to get the maximum and actual sizes of the buffer other than parsing the text output of `ethtool`. `man ethtool` says there's a `--json` flag, but only available to some subcommands, and I haven't been able to run it for `ethtool -g`.

I separated the actual command call from the parsing so that we could at least test the parsing. I've also conducted manual tests and this works as expected.

As a bonus, we now log (INFO level) the old, new and max sizes of the buffer.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-64590

### DIFF
--- a/deployment/terraform/create_utils.go
+++ b/deployment/terraform/create_utils.go
@@ -76,7 +76,7 @@ type ethtoolOutput struct {
 
 // getEthtoolRxSizes runs the ethtool command and parses its output.
 func getEthtoolRxSizes(sshc *ssh.Client) (ethtoolOutput, error) {
-	getRXCmd := "sudo ethtool -g ens3"
+	getRXCmd := "sudo ethtool -g $(ip route show to default | awk '{print $5}')"
 	out, err := sshc.RunCommand(getRXCmd)
 	if err != nil {
 		return ethtoolOutput{}, fmt.Errorf("error running ssh command; output: %q; cmd: %q, err: %w", string(out), getRXCmd, err)


streamer45:
Not to derail this, but looking at [ethtool source](https://git.kernel.org/pub/scm/network/ethtool/ethtool.git/tree/ethtool.c#n2134), the information is a simple syscall away. AI was able to get this for me fairly easily:

```go
package main

import (
	"fmt"
	"os"
	"syscall"
	"unsafe"
)

const (
	SIOCETHTOOL        = 0x8946
	ETHTOOL_GRINGPARAM = 0x00000010
	IFNAMSIZ           = 16
)

type EthtoolRingparam struct {
	Cmd               uint32
	RxMaxPending      uint32
	RxMiniMaxPending  uint32
	RxJumboMaxPending uint32
	TxMaxPending      uint32
	RxPending         uint32
	RxMiniPending     uint32
	RxJumboPending    uint32
	TxPending         uint32
}

type ifreq struct {
	Name [IFNAMSIZ]byte
	Data uintptr
}

func getRingParams(iface string) (*EthtoolRingparam, error) {
	fd, err := syscall.Socket(syscall.AF_INET, syscall.SOCK_DGRAM, 0)
	if err != nil {
		return nil, fmt.Errorf("socket error: %v", err)
	}
	defer syscall.Close(fd)

	ring := &EthtoolRingparam{
		Cmd: ETHTOOL_GRINGPARAM,
	}

	var ifr ifreq
	copy(ifr.Name[:], iface)
	ifr.Data = uintptr(unsafe.Pointer(ring))

	_, _, errno := syscall.Syscall(syscall.SYS_IOCTL,
		uintptr(fd),
		uintptr(SIOCETHTOOL),
		uintptr(unsafe.Pointer(&ifr)),
	)

	if errno != 0 {
		return nil, fmt.Errorf("ioctl error: %v", errno)
	}

	return ring, nil
}

func main() {
	if len(os.Args) != 2 {
		fmt.Fprintf(os.Stderr, "Usage: %s <interface>\n", os.Args[0])
		os.Exit(1)
	}

	iface := os.Args[1]
	ring, err := getRingParams(iface)
	if err != nil {
		fmt.Fprintf(os.Stderr, "Failed to get ring params: %v\n", err)
		os.Exit(1)
	}

	fmt.Printf("Interface: %s\n", iface)
	fmt.Printf("RX Pending: %d / Max: %d\n", ring.RxPending, ring.RxMaxPending)
	fmt.Printf("TX Pending: %d / Max: %d\n", ring.TxPending, ring.TxMaxPending)
}
```

Again, this would have to run remotely and unless we compile something we'd require a Go runtime installed, so I'm not suggesting a change of approach, but I wanted to provide a potentially more robust option up for discussion.

deployment/terraform/create.go
```diff
@@ -877,11 +876,6 @@ func (t *Terraform) setupProxyServer(extAgent *ssh.ExtAgent, instance Instance)
 			cacheObjects = "50m"
 			cacheSize = "16g" // Ideally we'd like half of the total server mem. But the mem consumption rarely exceeds 10G
 			// from my tests. So there's no point stretching it further.
-
-			// MM-58179
-			// We are increasing the receive ring buffer size on the network card. This proved to significantly lower packet loss
-			// (and retransmissions) on particularly bursty connections (e.g. websockets).
-			rxQueueSize = 8192
```
streamer45, line 884: My only (minor) concern (which I shared already) is that so far we've tested with a fixed value, which feels a bit more stable and reproducible. We don’t really have evidence that using the maximum is always the best choice, since it's ultimately a tradeoff between latency and throughput. 

That said, I don’t think we plan to vertically scale the proxy nodes any further, especially given that we’re taking the DNS-based load balancing route beyond 100k users. Also, surfacing this through logs is a great improvement which helps mitigate that concern to some extent.

deployment/terraform/create_utils.go
```diff
@@ -35,3 +36,97 @@ func (t *Terraform) setupPrometheusNodeExporter(sshClient *ssh.Client) error {
...
+	}
+
+	oldRxSize := rxSizes.actualRX
+
+	// Modify RX queue size to the maximum permitted
+	incRxSizeCmd := fmt.Sprintf("sudo ethtool -G $(ip route show to default | awk '{print $5}') rx %d", rxSizes.maxRX)
+	cmd := fmt.Sprintf("%s && sudo sysctl -p && sudo systemctl restart nginx", incRxSizeCmd)
+	if out, err := sshc.RunCommand(cmd); err != nil {
+		return fmt.Errorf("error running ssh command; output: %q; cmd: %q; err: %w", string(out), cmd, err)
+	}
```
streamer45, line 55: Since we implemented a `get` util, perhaps I would have expected a `set` one as well.

streamer45 (approved):
Approach makes sense. Left non-blocking thoughts.

deployment/terraform/create_utils.go
agarciamontoro: I'd thought about that, but it would add a new layer of indirection for just 6 lines, without the benefit of testing (this would need an E2E test) nor de-duplication (this is the only place where we set this value), which were the main reasons for implementing the `get` util.

deployment/terraform/create_utils.go
```diff
@@ -35,3 +36,97 @@ func (t *Terraform) setupPrometheusNodeExporter(sshClient *ssh.Client) error {
...
+	rxSizes, err := getEthtoolRxSizes(sshc)
+	if err != nil {
+		return fmt.Errorf("error getting the RX sizes from the ethtool: %w", err)
+	}
+
+	oldRxSize := rxSizes.actualRX
+
+	// Modify RX queue size to the maximum permitted
+	incRxSizeCmd := fmt.Sprintf("sudo ethtool -G $(ip route show to default | awk '{print $5}') rx %d", rxSizes.maxRX)
+	cmd := fmt.Sprintf("%s && sudo sysctl -p && sudo systemctl restart nginx", incRxSizeCmd)
```
agnivade, line 52: Since we are always setting the max queue size, I wonder if we can come up with a single unix one-liner which parses the maxRX from the `-g` command, and feeds that to the `-G` command? Or if that's cumbersome, this is fine as well. 0/5

agnivade (approved):

PR #932, opened 2025-06-27
fix: prometheus exporter setup behind rhel for agents

fmartingr:
#### Summary

Probably caused by #919 

The prometheus setup was locked behind a RHEL check, which I think shouldn't be the case.

### DIFF
--- a/deployment/terraform/agent.go
+++ b/deployment/terraform/agent.go
@@ -191,10 +191,6 @@ func (t *Terraform) configureAndRunAgents(extAgent *ssh.ExtAgent) error {
 				return
 			}
 
-			if err := t.setupPrometheusNodeExporter(sshc); err != nil {
-				mlog.Error("error setting up prometheus node exporter", mlog.Err(err), mlog.Int("agent", agentNumber))
-			}
-
 			cmd = "sudo systemctl restart otelcol-contrib && sudo systemctl restart prometheus-node-exporter"
 			if out, err := sshc.RunCommand(cmd); err != nil {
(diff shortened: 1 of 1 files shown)


fmartingr:
> Actually, this was me: [#919 (comment)](https://github.com/mattermost/mattermost-load-test-ng/pull/919#discussion_r2113277931). On a deeper look, looks like I've made a mistake. Do you remember any context around that?

I recall you asking about that for SELinux, not for the prom exporter 🤔  Would it make sense to have it excluded in non-rhel clusters anyway?

agarciamontoro:
I wonder why this was not there before #919, and why we added it with a RHEL check then. Also, I'm a bit concerned this adds too many metrics in large tests, like ceiling tests, where we can have hundreds of agents.

fmartingr:
> I wonder why this was not there before #919, and why we added it with a RHEL check then. Also, I'm a bit concerned this adds too many metrics in large tests, like ceiling tests, where we can have hundreds of agents.

You are right. Let me know if I should just remove it altogether or put it behind a configuration flag.

fmartingr:
> I'm ok getting it in as is, but we should try to estimate how much data per agent per hour/day it consumes.

If you can tell me how do you want to measure this I can take care. Leaving the PR blocked in the meanwhile.

agnivade:
We do need the prom exporter for all agents right? Otherwise how can we get the node level metrics from agents?

agnivade:
Hmm .. I am seeing this `sudo apt-get install -y prometheus-node-exporter &&` in agent.sh. Should we remove this then?

agarciamontoro:
@agnivade, are you planning to add this only in the browser agents? In that case, we can approve and merge this PR.

PR #940, opened 2025-07-09
MM-63312: Add logic to refresh token in the background

agarciamontoro:
#### Summary

This PR depends on #939.

The original fix proposed by Antonis in https://github.com/mattermost/mattermost-load-test-ng/pull/891 (the first two commits in this PR) modified the authentication logic to assume a role, provided by the user, that is expected to have all the permissions needed by the tool. However, that solution ended up hitting the same issue: whenever `GetAWSConfig` or `GetAWSCreds` were called, the credentials used to assume the user-provided role were the ones in the default credentials chain; i.e., the ones in the environment. As the credentials in CI are based on role chaining, they expire after just one hour. Thus, we were still trying to assume a role from expired credentials, which failed.

This PR fixes this by constantly refreshing the credentials, using the currently retrieved credentials (not the ones in the environment) as a base to re-assume the role (this is why we need #939, because we need to keep those credentials in the state all the time we can). We need to do this *before* those credentials are expired (that still happens in one hour because we cannot overcome the role chaining constraint), so we repeat the refresh logic every 25 minutes (we could do it every 59 minutes and 59 seconds, but out of an abundance of caution, I wanted to fit at least two refresh steps in the expiry time window).

I've tested this extensively and it works. I'll repeat a new test once #939 is merged and this one is approved.

Btw, I'm only 1/5 on the condition used to start the refresh logic (that AWSRoleARN is non-empty): we could add a new config setting, as well as another one for the duration and refresh logic, but I don't think it's completely necessary. My mind can be changed, though, I'm open to comments :)

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-63312


deployment/terraform/utils.go
```diff
@@ -352,20 +355,114 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
-// GetAWSConfig returns the AWS config, using the profile configured in the
-// deployer if present, and defaulting to the default credential chain otherwise
-func (t *Terraform) GetAWSConfig() (aws.Config, error) {
+// Durations for the expiry of the AWS Role credentials and for its refresh
+// interval, which, out of an abundance of caution, is a bit shorter than half
+// the expiry duration of the role.
+const (
+	// One hour is the maximum allowed by role chaining
+	AWSRoleTokenDuration        = 1 * time.Hour
+	AWSRoleTokenRefreshInterval = 25 * time.Minute
```
agnivade, line 364: Can we do `AWSRoleTokenRefreshInterval = AWSRoleTokenDuration/2 ` for simplicity? Also, are we really gaining anything by reducing the interval by a further 5 minutes? Or is it just a peace of mind thing? If we really want to reduce further than half, I'd just prefer to run it 4 times per hour. Every 25 minutes looks a bit odd to me.

deployment/terraform/utils.go
```diff
@@ -352,20 +355,114 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+}
+
+// GetAWSConfig returns the AWS config, using the profile configured in the
+// deployer if present, and defaulting to the default credential chain otherwise.
+// If a role ARN is provided, it will assume that role.
+func (t *Terraform) GetAWSConfig() (aws.Config, error) {
+	t.awsCfgMut.Lock()
+	defer t.awsCfgMut.Unlock()
+	return t.awsCfg.Copy(), nil
 }
```
agnivade, line 466: This does not seem to be called anymore. Intentional?

deployment/terraform/utils.go
```diff
@@ -352,20 +355,114 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+		// Refresh the credentials once so that the time of the interval below
+		// starts now
+		if err := t.refreshAWSCredentialsFromRole(); err != nil {
+			return fmt.Errorf("unable to refresh token: %w", err)
+		}
+
+		// Start a goroutine refreshing the credentials every AWSRoleTokenRefreshInterval.
+		// If it fails, exit the goroutine, since it needs the previous credentials to be
+		// valid for the refresh logic to work.
+		go func() {
```
agnivade, line 405: Do we need to bother about shutting down the goroutine during exit? Or just letting it die is good enough?

deployment/terraform/utils.go
```diff
@@ -352,20 +355,114 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+// Durations for the expiry of the AWS Role credentials and for its refresh
+// interval, which, out of an abundance of caution, is a bit shorter than half
+// the expiry duration of the role.
+const (
+	// One hour is the maximum allowed by role chaining
+	AWSRoleTokenDuration        = 1 * time.Hour
+	AWSRoleTokenRefreshInterval = 25 * time.Minute
+)
+
+func (t *Terraform) InitCreds() error {
```
agnivade, line 367: Can we give some godocs here?

deployment/terraform/utils.go
```diff
@@ -352,20 +355,114 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+	if t.config.AWSRoleARN != "" {
+		// See https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials/stscreds#hdr-Assume_Role
+		// Create the credentials from AssumeRoleProvider to assume the role
+		// identified by AWSRoleARN, and use them in the loaded config
+		stsSvc := sts.NewFromConfig(cfg)
+		creds := stscreds.NewAssumeRoleProvider(stsSvc, t.config.AWSRoleARN)
+		cfg.Credentials = aws.NewCredentialsCache(creds)
+	}
+
+	t.awsCfgMut.Lock()
```
agnivade, line 391: Curious to understand why is this mutex needed? Are we calling `t.Init` more than once from somewhere?

deployment/terraform/utils.go
```diff
@@ -352,20 +355,114 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
-			context.Background(),
-			regionOpt,
-		)
+	return nil
+}
+
+// refreshAWSCredentialsFromRole refreshes the current credentials by assuming
+// the role again, with an expiry duration of one hour, which is the maximum
+// allowed when using role chaining
+func (t *Terraform) refreshAWSCredentialsFromRole() error {
```
agnivade, line 422: Is it possible at all to test this? Seems like a vital logic.

deployment/terraform/utils.go
agarciamontoro: I thought about this, but there is really no explicit exit. When we create the `Terraform` struct, we can use it for `Create`, or for `Destroy`, or for inspecting the output, but there is no explicit signal of "we're done with this", so I just decided to let it die.

deployment/terraform/utils.go
agnivade: Ah sorry! I thought this method was implemented new for this PR :facepalm:

agarciamontoro: The diff UI is terrible, I don't blame you :joy:

streamer45:
@agarciamontoro Is it possible to rebase/force push to aid review? I am seeing all the previous code we already merged now that the base branch has changed.

agarciamontoro:
@streamer45: Whoops, sorry for that, fixed now :) I should have used @cpoile's tool!

deployment/terraform/create.go
```diff
@@ -82,11 +85,17 @@ func New(id string, cfg deployment.Config) (*Terraform, error) {
...
 
-	return &Terraform{
+	t := &Terraform{
 		id:        id,
 		config:    &cfg,
 		genValues: genValues,
-	}, nil
+	}
+
+	if err := t.InitCreds(); err != nil {
```
streamer45, line 94: Does this need to run outside of CI? If so, I'd be nice to have a `os.Getenv("CI") == true` or something.

deployment/terraform/utils.go
```diff
@@ -352,20 +355,116 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
 
-// GetAWSConfig returns the AWS config, using the profile configured in the
-// deployer if present, and defaulting to the default credential chain otherwise
-func (t *Terraform) GetAWSConfig() (aws.Config, error) {
+// Durations for the expiry of the AWS Role credentials and for its refresh
+// interval, which, out of an abundance of caution, is a quarter of the expiry
+// duration of the role.
+const (
+	// One hour is the maximum allowed by role chaining
+	AWSRoleTokenDuration        = 1 * time.Hour
```
streamer45, line 363: nit: 1 multiplier is redundant.

deployment/terraform/utils.go
```diff
@@ -352,20 +355,116 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+		if err := t.refreshAWSCredentialsFromRole(); err != nil {
+			return fmt.Errorf("unable to refresh token: %w", err)
+		}
+
+		// Start a goroutine refreshing the credentials every AWSRoleTokenRefreshInterval.
+		// If it fails, exit the goroutine, since it needs the previous credentials to be
+		// valid for the refresh logic to work.
+		go func() {
+			ticker := time.Tick(AWSRoleTokenRefreshInterval)
+			for range ticker {
```
streamer45, line 409: I guess now that we no longer need to explicitly stop, we could just combine it into a single line.

deployment/terraform/utils.go
```diff
@@ -352,20 +355,116 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+		return fmt.Errorf("failed to retrieve current credentials: %w", err)
+	}
+
+	// Create new STS service and assume role again
+	stsSvc := sts.NewFromConfig(t.awsCfg)
+	creds := stscreds.NewAssumeRoleProvider(stsSvc, t.config.AWSRoleARN, func(opt *stscreds.AssumeRoleOptions) { opt.Duration = AWSRoleTokenDuration })
+	newCredCache := aws.NewCredentialsCache(creds)
+
+	// Test new credentials
+	newCreds, err := newCredCache.Retrieve(context.Background())
```
streamer45, line 444: Could this (and related calls) block for a long time? (asking since we are not passing a timeout).

deployment/terraform/utils.go
```diff
@@ -352,20 +355,116 @@ func getServerURL(output *Output, deploymentConfig *deployment.Config) string {
...
+	return nil
+}
+
+// GetAWSConfig returns the AWS config, using the profile configured in the
+// deployer if present, and defaulting to the default credential chain otherwise.
+// If a role ARN is provided, it will assume that role.
+func (t *Terraform) GetAWSConfig() (aws.Config, error) {
+	t.awsCfgMut.Lock()
+	defer t.awsCfgMut.Unlock()
+	return t.awsCfg.Copy(), nil
```
streamer45, line 467: Wondering if using an `atomic.Pointer` might have been simpler, but I guess it doesn't matter that much.

streamer45 (approved):
Nothing blocking :+1:

deployment/terraform/create.go
agarciamontoro: Good point! We do need to call `InitCreds` because that's how we authenticate now, but we can avoid the refresh goroutine if we're not in CI. Done in https://github.com/mattermost/mattermost-load-test-ng/pull/940/commits/5297daa8784db29d113791fb7d2d33fa769c6f3b.

deployment/terraform/utils.go
agarciamontoro: Yeah, but `0 + time.Hour` was weirder.

deployment/terraform/utils.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-load-test-ng/pull/940/commits/6cf5211ffd14ddf2114fbe4d3f0549eb77e7672f

deployment/terraform/utils.go
agarciamontoro: Ah, great catch! Added a 30-second timeout for each call in the goroutine: https://github.com/mattermost/mattermost-load-test-ng/pull/940/commits/eb007124c0e9fdc66d047f54c4b6b82eba675b79

deployment/terraform/utils.go
agarciamontoro: Yeah, we could, but agree that it probably doesn't matter much.

deployment/terraform/utils.go
agnivade: `0 + time.Hour` is much clearer! We should use that everywhere!

agarciamontoro:
My latest test with #941 cherry-picked in a test branch based off of this one was successful :tada: Now that that PR is merged, I'll do one final test and I'll merge afterwards.

PR #941, opened 2025-07-10
fix: make RestoreAssets copy directories recursively

fmartingr:
## Summary

- Fixed `RestoreAssets` function to properly copy directories recursively
- The original implementation only copied single files, but directories like "provisioners" require recursive copying to match go-bindata behavior
- Added comprehensive test coverage for all asset functions

### RestoreAssets Function (`deployment/terraform/assets.go`)
- Uses `fs.Stat` to check if asset is a directory
- Uses `fs.WalkDir` for efficient recursive directory traversal
- Maintains proper file permissions (0644 for files, 0755 for directories)
- Handles both files and directories correctly

### Test Coverage (`deployment/terraform/assets_test.go`)
- **Asset functions**: 100% coverage for `Asset`, `MustAsset`, `AssetString`, `MustAssetString`
- **RestoreAssets**: 79.3% coverage with tests for:
  - Single file restoration
  - Recursive directory restoration
  - Multiple file restoration
  - Error handling for non-existent assets
  - File and directory permissions
  - Cross-function consistency

## Test Results

All tests pass:
- 14 new test functions covering all asset functions
- Validates recursive copying of `provisioners` directory with nested `debian` and `rhel` subdirectories
- Ensures file permissions and directory structure are preserved

## Background

This fixes the issue where `RestoreAssets("provisioners")` would fail because the function didn't support recursive directory copying after the migration from go-bindata to Go's standard `embed` package.

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/deployment/terraform/assets.go
+++ b/deployment/terraform/assets.go
@@ -5,6 +5,7 @@ package terraform
 
 import (
 	"embed"
+	"io/fs"
 	"os"
 	"path/filepath"
 )
@@ -39,6 +40,55 @@ func MustAssetString(name string) string {
 
 // RestoreAssets writes an embedded asset to the given directory
 func RestoreAssets(dir, name string) error {
+	assetPath := "assets/" + name
+
+	// Check if it's a directory
+	fileInfo, err := fs.Stat(assetsFS, assetPath)
+	if err != nil {
+		return err
+	}
+
+	if fileInfo.IsDir() {
+		// It's a directory, copy recursively using WalkDir
+		return fs.WalkDir(assetsFS, assetPath, func(path string, d fs.DirEntry, err error) error {
+			if err != nil {
+				return err
+			}
+
+			// Calculate the relative path from the asset root
+			relPath, err := filepath.Rel(assetPath, path)
+			if err != nil {
+				return err
+			}
+
+			// Skip the root directory itself
+			if relPath == "." {
+				return nil
+			}
+
+			targetPath := filepath.Join(dir, name, relPath)
+
+			if d.IsDir() {
+				// Create directory
+				return os.MkdirAll(targetPath, 0755)
+			} else {
+				// Copy file
+				data, err := assetsFS.ReadFile(path)
+				if err != nil {
+					return err
+				}
+
+				// Create parent directory if it doesn't exist
+				if err := os.MkdirAll(filepath.Dir(targetPath), 0755); err != nil {
+					return err
+				}
+
+				return os.WriteFile(targetPath, data, 0644)
+			}
+		})
+	}
+
+	// It's a file, copy it to the destination
 	data, err := Asset(name)
 	if err != nil {
 		return err
--- a/deployment/terraform/assets_test.go
+++ b/deployment/terraform/assets_test.go
@@ -1,0 +1,376 @@
+// Copyright (c) 2019-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package terraform
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// Test Asset function
+func TestAsset(t *testing.T) {
+	// Test reading a valid file
+	data, err := Asset("outputs.tf")
+	require.NoError(t, err)
+	require.NotEmpty(t, data)
+	
+	// Convert to string and verify it contains expected content
+	content := string(data)
+	require.Contains(t, content, "output")
+}
+
+func TestAsset_NonExistentFile(t *testing.T) {
+	// Test reading a non-existent file
+	_, err := Asset("non_existent_file.txt")
+	require.Error(t, err)
+}
+
+func TestAsset_ValidTerraformFile(t *testing.T) {
+	// Test reading specific terraform files
+	terraformFiles := []string{
+		"outputs.tf",
+		"variables.tf",
+		"cluster.tf",
+		"elasticsearch.tf",
+	}
+	
+	for _, fileName := range terraformFiles {
+		data, err := Asset(fileName)
+		require.NoError(t, err, "Failed to read %s", fileName)
+		require.NotEmpty(t, data, "File %s should not be empty", fileName)
+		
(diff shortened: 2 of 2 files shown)


deployment/terraform/assets.go
```diff
@@ -39,6 +40,56 @@ func MustAssetString(name string) string {
 
 // RestoreAssets writes an embedded asset to the given directory
 func RestoreAssets(dir, name string) error {
+	assetPath := "assets/" + name
```
agarciamontoro, line 43: Why not use `filepath.Join` here?
```suggestion
	assetPath := filepath.Join("assets", name)
```

deployment/terraform/assets.go
```diff
@@ -39,6 +40,56 @@ func MustAssetString(name string) string {
...
+
+	// Check if it's a directory
+	fileInfo, err := fs.Stat(assetsFS, assetPath)
+	if err != nil {
+		return err
+	}
+
+	if fileInfo.IsDir() {
```
agarciamontoro, line 51: Nit: We're not checking if it's a directory with the `Stat` call, but with the `IsDir` call.

deployment/terraform/assets_test.go
```diff
@@ -0,0 +1,376 @@
...
+func TestAsset(t *testing.T) {
+	// Test reading a valid file
+	data, err := Asset("outputs.tf")
+	require.NoError(t, err)
+	require.NotEmpty(t, data)
+
+	// Convert to string and verify it contains expected content
+	content := string(data)
+	require.Contains(t, content, "output")
+}
```
agarciamontoro, line 25: The tests in `TestAsset_ValidTerraformFile` make this one redundant.

deployment/terraform/assets_test.go
```diff
@@ -0,0 +1,376 @@
...
+		require.NoError(t, err, "Failed to read %s", fileName)
+		require.NotEmpty(t, data, "File %s should not be empty", fileName)
+
+		// Verify it's a terraform file by checking for terraform syntax
+		content := string(data)
+		require.True(t, strings.Contains(content, "resource") ||
+			strings.Contains(content, "variable") ||
+			strings.Contains(content, "output") ||
+			strings.Contains(content, "data"),
+			"File %s should contain terraform syntax", fileName)
```
agarciamontoro, line 53: I find it a bit weird that we have this or clause in the `require`. Should we convert this test into a table test, so each of the files has its own expected output? If we do that, we can also merge this test and `TestAsset_NonExistentFile` adding an expected error.

deployment/terraform/assets_test.go
```diff
@@ -0,0 +1,376 @@
...
+		require.NotEmpty(t, content, "File %s should not be empty", fileName)
+
+		// Verify it's a YAML file by checking for YAML syntax
+		require.True(t, strings.Contains(content, ":") ||
+			strings.Contains(content, "---") ||
+			strings.Contains(content, "name:") ||
+			strings.Contains(content, "type:"),
+			"File %s should contain YAML syntax", fileName)
+	}
+}
```
agarciamontoro, line 127: Why do we have a specific test for YAML files? Also, I have the same comment here about the or clause in the `require`, as well as the opportunity to convert this into a proper table test and merge it with the other `TestAssetString_*` tests.

deployment/terraform/assets_test.go
```diff
@@ -0,0 +1,376 @@
...
+	testCases := []struct {
+		fileName        string
+		expectedContent string
+	}{
+		{"outputs.tf", "output"},
+		{"variables.tf", "variable"},
+		{"cluster.tf", "resource"},
+		{"datasource.yaml", ":"},
+		{"dashboard.yaml", ":"},
+	}
```
agarciamontoro, line 197: We have test cases here now, nice!

deployment/terraform/assets.go
```diff
@@ -39,6 +40,56 @@ func MustAssetString(name string) string {
...
+			targetPath := filepath.Join(dir, name, relPath)
+
+			// Check if it's a directory and only proceed for files, since files are the only ones we want to copy
+			// and we make sure the target directory exists before copying. This prevents empty folders.
+			if !d.IsDir() {
+				// Copy file
+				data, err := assetsFS.ReadFile(path)
+				if err != nil {
+					return err
+				}
```
agarciamontoro, line 78: Nit: we're reading the file here, not copying.

deployment/terraform/assets.go
```diff
@@ -39,6 +40,56 @@ func MustAssetString(name string) string {
...
+
+				// Create parent directory if it doesn't exist
+				if err := os.MkdirAll(filepath.Dir(targetPath), 0755); err != nil {
+					return err
+				}
+
+				return os.WriteFile(targetPath, data, 0644)
```
agarciamontoro, line 85: This code block and the one for copying it if the required asset is a file is virtually the same. Could we refactor into a util function to use it in both places?

agarciamontoro (changes requested):
Nice! I just tested it and it fixed the issue, thanks!

I have a few non-blocking suggestions in the main code, and some others in the test code: we are mixing table tests and normal tests, and I would like to be consistent with the rest of the codebase and have a single test for each function, and then inner `t.Run` subtests for different cases. Thoughts?

deployment/terraform/assets.go
```diff
@@ -37,19 +38,67 @@ func MustAssetString(name string) string {
...
-	if err := os.MkdirAll(filepath.Dir(outputPath), 0755); err != nil {
+			// Skip the root directory itself
+			if relPath == "." {
+				return nil
+			}
+
+			targetPath := filepath.Join(dir, name, relPath)
+
+			// Check if it's a directory and only proceed for files, since files are the only ones we want to copy
+			// and we make sure the target directory exists before copying. This prevents empty folders.
```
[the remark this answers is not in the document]
agarciamontoro: Actually, only a Stat call: https://cs.opensource.google/go/go/+/refs/tags/go1.24.5:src/os/path.go;l=20-24 Still some overhead, but probably negligible.

PR #944, opened 2025-09-02
[MM 65580] Add simulation test timeouts as configurable property in config.json

M-ZubairAhmed:
#### Summary
* Added a new `SimulationTimeoutMs` parameter to the browser configuration schema, with validation to ensure it is a non-negative number. 
* Set the default timeout for all Playwright page actions in `BrowserTestSessionManager` to the configured `SimulationTimeoutMs` value.
* In `goToChannel`, added a wait for the `#virtualizedPostListContent` element to ensure the channel is fully loaded before proceeding.
* In `postInChannel`, added a check to wait until the message textbox is cleared after posting, ensuring the post action is complete.

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-65580

### DIFF
--- a/browser/src/simulations/go_to_channel.ts
+++ b/browser/src/simulations/go_to_channel.ts
@@ -13,6 +13,9 @@ export async function goToChannel(page: Page, channelId: string): Promise<void> 
     await channel.waitFor({state: 'visible'});
     await channel.click();
 
+    // # Wait until the loading screen is gone and the channel is loaded
+    await page.locator('#virtualizedPostListContent').waitFor({state: 'visible'});
+
     log.info('pass--goToChannel');
   } catch (error) {
     throw {error, testId: 'goToChannel'};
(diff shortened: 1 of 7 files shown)


agarciamontoro (changes requested):
Nice! I think we need to review how the default/disabling it works, but apart from that it's perfect. Also, a general nit: if the values are large, should we use seconds instead of milliseconds as the unit?

browser/src/simulations/post_in_channel.ts
```diff
@@ -19,6 +19,9 @@ export async function postInChannel({page}: {page: Page}): Promise<void> {
     await postButton.waitFor({state: 'visible'});
     await postButton.click();
 
+    // # Wait until textbox is cleared
```
fmartingr, line 22:
```suggestion
    // Wait until textbox is cleared
```

browser/src/simulations/go_to_channel.ts
```diff
@@ -13,6 +13,9 @@ export async function goToChannel(page: Page, channelId: string): Promise<void>
     await channel.waitFor({state: 'visible'});
     await channel.click();
 
+    // # Wait until the loading screen is gone and the channel is loaded
```
fmartingr, line 16:
```suggestion
    // Wait until the loading screen is gone and the channel is loaded
```

PR #945, opened 2025-09-18
MM-65142: Add AWSAMIUser to output struct

agarciamontoro:
#### Summary
The `comparison collect` command does not work because Output.AWSAMIUser is empty. We need to populate it when loading the output.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65142


fmartingr (approved):
Upsies 😓

cpoile (approved):

PR #948, opened 2025-10-01
MM-65977: Add new MattermostPlugins setting

agarciamontoro:
#### Summary
This PR introduces a new `MattermostPlugins` setting, which lets users specify the exact set of plugins to install in the deployment, with the possibility to specify either officially released plugins through their release URL (or any URL, for that matter), or locally built ones through a local filesystem path.

Note that this was kept as simple as possible: just a map of plugin IDs to their tarballs. On our meeting, @M-ZubairAhmed and I discussed about adding all sorts of bells and whistles here: a common way of setting plugin-specific configuration, some validation to only allow specific plugins (since we will need data to our datasets for each that we add), some way of specifying just the plugin and the version, to retrieve the officially released ones... I left all that aside for now, and implemented the most simple solution, we can iterate in the future if needed.

There's a couple of additional commits here:
1. I had to move the `ProvisionURL` function under the `Terraform` struct, since it had a bug with locally built files and to fix it I needed to have access to the home directory. While I was there, I also changed the signature so that it returned the path of the provisioned file.
2. A random improvement to an error log line that would have helped me in a debugging session.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-65977

### DIFF
--- a/deployment/terraform/utils.go
+++ b/deployment/terraform/utils.go
@@ -536,3 +536,32 @@ func generatePseudoRandomPassword(length int) string {
 	}
 	return string(s)
 }
+
+// ProvisionURL takes a URL pointing to a file to be provisioned.
+// It works on both local files prefixed with file:// or remote files.
+// In case of local files, they are uploaded to the server.
+func (t *Terraform) ProvisionURL(client *ssh.Client, url, filename string) (string, error) {
+	dstPath := fmt.Sprintf("/home/%s/%s", t.config.AWSAMIUser, filename)
+	if strings.HasPrefix(url, "file://") {
+		// upload file from local filesystem
+		path := strings.TrimPrefix(url, filePrefix)
+		info, err := os.Stat(path)
+		if err != nil {
+			return "", err
+		}
+		if !info.Mode().IsRegular() {
+			return "", fmt.Errorf("build file %s has to be a regular file", path)
+		}
+		if out, err := client.UploadFile(path, dstPath, false); err != nil {
+			return "", fmt.Errorf("error uploading build: %w %s", err, out)
+		}
+	} else {
+		// download build file from URL
+		cmd := fmt.Sprintf("wget -O %s %s", dstPath, url)
+		if out, err := client.RunCommand(cmd); err != nil {
+			return "", fmt.Errorf("failed to run cmd %q: %w %s", cmd, err, out)
+		}
+	}
+
+	return dstPath, nil
+}
--- a/comparison/deployment.go
+++ b/comparison/deployment.go
@@ -9,7 +9,6 @@ import (
 	"path/filepath"
 	"sync"
 
-	"github.com/mattermost/mattermost-load-test-ng/deployment"
 	"github.com/mattermost/mattermost-load-test-ng/deployment/terraform"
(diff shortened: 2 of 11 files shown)


deployment/terraform/utils.go
```diff
@@ -536,3 +536,32 @@ func generatePseudoRandomPassword(length int) string {
...
+			return "", fmt.Errorf("build file %s has to be a regular file", path)
+		}
+		if out, err := client.UploadFile(path, dstPath, false); err != nil {
+			return "", fmt.Errorf("error uploading build: %w %s", err, out)
+		}
+	} else {
+		// download build file from URL
+		cmd := fmt.Sprintf("wget -O %s %s", dstPath, url)
+		if out, err := client.RunCommand(cmd); err != nil {
+			return "", fmt.Errorf("failed to run cmd %q: %w %s", cmd, err, out)
```
cpoile, line 562: The only thing I wonder is, should we be verifying the download?

cpoile (approved):
Cool, nice work! Just one non-blocking question (ping me with the answer so I don't miss it :) )

deployment/terraform/utils.go
agarciamontoro: What do you have in mind? At first I thought about maybe allowing only URLs like `https://plugins.releases.mattermost.com/release/mattermost-plugin-playbooks-v2.4.1-linux-amd64.tar.gz`, from which we can infer the signature file, which is uploaded alongside those, but that seemed overkill for this feature. Something like the user providing a checksum of the file doesn't seem too useful to me either.

Open to ideas for more relaxed ways of verifying downloads, though :)

cpoile: I think it's all or nothing really. The basics would be to only accept urls like that (or from github), and only accept them when the sig is right next to it (which you can get by replacing the ext). Then add the `mattermostPluginPublicKey` constant from the server code, and run the verify process as done here https://github.com/mattermost/mattermost/blob/master/server/channels/app/plugin_signature.go#L78

Otherwise, just accept that the user using the load test knows what they're doing, and the worst that could happen is they screw up their load test machines?

agarciamontoro: > I think it's all or nothing really. The basics would be to only accept urls like that (or from github), and only accept them when the sig is right next to it (which you can get by replacing the ext). Then add the mattermostPluginPublicKey constant from the server code, and run the verify process as done here [mattermost/mattermost@master/server/channels/app/plugin_signature.go#L78](https://github.com/mattermost/mattermost/blob/master/server/channels/app/plugin_signature.go#L78)

Got it! I see what you say, and I would agree on this for a more production-ready tool, but our main customer right now is us, developers, and the infrastructure is always temporary, so:

> Otherwise, just accept that the user using the load test knows what they're doing, and the worst that could happen is they screw up their load test machines?

Yeah, this was my main reasoning. There are a million ways in which the user can screw up their load test machines using this tool. Like, a lot. This is a DDoS engine, after all :joy: Also, we do need to let the user upload local builds so that adding coverage for or benchmarking new features can be done easily by developers. And that would be impossible to verify.

So unless you feel very strongly on this, I would assume users know what they're doing (and handle any screaming afterwards, if needed :) )

PR #997, opened 2026-09-17
MM-70767: Fix noop controller never setting current team

agarciamontoro:
#### Summary
- `joinTeam` in the noop controller adds team membership server side but never sets the current team in the local user store.
- The next action, `joinChannel`, requires a current team and fails every cycle for every noop user with "current team should be set".
- Set the current team both when joining a new team and when already a member of all available teams.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-70767


carlisgg (approved):
LGTM

